PR #31, opened 2013-05-10
Fix rogue talent Improved Sprint

cala:
Fix rogue sprint no longer working when talent Improved Sprint was active

This commit [482ca12](http://github.com/mangos-zero/server-old/commit/482ca12) is one of those pending backports from old mangos-zero concatenated in: https://github.com/cmangos/issues/wiki/classic_backporting-todo-zero-commits

I tested it and this is my report:
> Improved sprint is indeed broken as it prevent sprint from working. This commit fixes it.


Kelzior:
not working for me

Argonn:
Isn't it possible to match the 2 types of Aura? (109 & 31) or just based on the Id's ?

PR #54, opened 2014-02-17
Fix NPCs having the wrong orientation while fleeing.

Patman64:
SelectHostileTarget was not taking into account loss-of-control and was
mistakenly making the creature face the player even while it was running
away.

Fixes cmangos/issues#375.


cyberium:
This sound not be a definitive solution.

PR #55, opened 2014-02-21
Fix audio repeating due to SMSG_INIT_WORLD_STATES.

Patman64:
Also simplify Player::SendInitWorldStates significantly.

I remove the sending of the default world states if the script isn't available because the script should be available under all circumstances, and if it isn't we shouldn't pretend that it is.

Fixes cmangos/issues#266.


Patman64:
Was fixed instead in https://github.com/cmangos/mangos-classic/commit/11e6ad7b15c5d12bd4674e4574cf64f7da442459.

PR #56, opened 2014-02-21
Lift the check for facing your target out of Spell::CheckRange.

Patman64:
Whether the caster is facing the target or not has nothing to do with
whether the target is in range. CheckRange returned early for spells with
SPELL_RANGE_IDX_COMBAT, skipping the facing-target check.

Also remove code that allows spells triggered by auras to be cast
regardless of range, reverting part of [z0555].

This fixes cmangos/issues#384.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4327,12 +4327,15 @@ SpellCastResult Spell::CheckCast(bool strict)
 
     if (!m_IsTriggeredSpell)
     {
-        if (!m_triggeredByAuraSpell)
-        {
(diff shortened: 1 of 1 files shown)


cala:
I will test this soon(tm). Can you provide a few examples (spells ?) in order to properly test this, please ?

VladimirMangos:
I worry that check added for case when client dbc data have wrong ranges for aura triggered spells. I already not remember cases. But this need re-check before remove if...

cyberium:
@killerwife please close this when you'll feel it.

killerwife:
Closed by https://git.io/vQfwi

PR #59, opened 2014-03-02
Add config option to mangosd for tweaking reactor timing.

PatSmuk:
This can be used to significantly reduce idle CPU usage.

More info here: http://cmangos.net/thread-576.html

### DIFF
--- a/src/game/WorldSocketMgr.cpp
+++ b/src/game/WorldSocketMgr.cpp
@@ -154,12 +154,13 @@ class ReactorRunnable : protected ACE_Task_Base
             MANGOS_ASSERT(m_Reactor);
 
             SocketSet::iterator i, t;
+            const int timeout = sConfig.GetIntDefault("Network.Timeout", 100000);
 
             while (!m_Reactor->reactor_event_loop_done())
             {
                 // dont be too smart to move this outside the loop
                 // the run_reactor_event_loop will modify interval
-                ACE_Time_Value interval(0, 10000);
+                ACE_Time_Value interval(0, timeout);
 
                 if (m_Reactor->run_reactor_event_loop(interval) == -1)
                     break;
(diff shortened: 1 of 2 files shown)


sofar:
Confirmed this bug on Linux. 15% CPU utilization on a relatively new Core i5 processor without any clients logged in.

This bug needs re-opening. @cyberium - please reopen this, your screenshot shows you run windows and 3 people have now confirmed this bug on Linux.

sofar:
Looking at the code section here, I don't understand why polling 100 times per second to do the following things is needed:

- accept new connections from clients
- close old connections

None of this is needed to increase *gameplay* latency. The timeout could easily be 1 second, and nobody would ever notice a thing.

Once the sockets to the clients are established, this poll loop isn't the loop that handles the client packet stream.

I'd merge this patch as is, it's a straightforward and simple patch.

sofar:
The following drops my CPU from ~20% to 8%

```
diff --git a/src/game/WorldSocketMgr.cpp b/src/game/WorldSocketMgr.cpp
index b65708d..d409c65 100644
--- a/src/game/WorldSocketMgr.cpp
+++ b/src/game/WorldSocketMgr.cpp
@@ -159,7 +159,9 @@ class ReactorRunnable : protected ACE_Task_Base
             {
                 // dont be too smart to move this outside the loop
                 // the run_reactor_event_loop will modify interval
-                ACE_Time_Value interval(0, 10000);
+
+                // poll every 10ms when there are player connections, idle otherwise
+                ACE_Time_Value interval(0, m_Connections > 0 ? 10000 : 1000000);
 
                 if (m_Reactor->run_reactor_event_loop(interval) == -1)
                     break;
```

Lillecarl:
Thanks for confirming! :-)

cyberium:
Thanks alot for your feedback.
Fortunatly in some weeks, we will remove ACE lib completly so no reason to reopen this.

Lillecarl:
Great news! Good to see that old fart going away! :-)

PR #73, opened 2014-05-07
In AV, defenders shouldn't despawn at point tag

Rockette:
In AV, defenders shouldn't despawn at point tag.

No sources find, but memory and BattleGroundAV.cpp:443 confirm this.


cyberium:
Ok i did some check and this solution is bad. The main reason is actualy if you dont despawn them they will constantly respawn. But also for @Schmoozerd explained reason

PR #78, opened 2014-05-25
Fix crash in MoveMap extractor.

Patman64:
Also include new build of MoveMapGen.exe (built with VS Express 2013, Release Win32, about 1/3 of the size).


cala:
Never had a crash (does not mean there no need to fix), but it is hard for me to provide a feedback when everything is working.

Patman64:
@cala Use [this script](https://gist.github.com/Patman64/eb03adfb8f6c8bbc5930) to extract mmaps and it will crash within seconds.

cyberium:
i cannot accept that fix, removing a delete to fix a crash will only cause memory issues

PR #79, opened 2014-05-25
Add more advanced version of mmap_extractor.py.

Patman64:
- Extract per-tile rather per-map to distribute work between threads much
  more evenly.
- Scans for finished tiles to avoid doing duplicate work (e.g. if the
  extraction is interrupted).
- Number of threads can be adjusted with command-line argument (default is
  number of cores - 1).
- Use separate thread for printing to avoid garbage output.
- Times each operation and displays remaining tile/map count.

This is applicable to all versions of cmangos (the map list just needs to be changed).

### DIFF
--- a/contrib/mmap/mmap_extract.py
+++ b/contrib/mmap/mmap_extract.py
@@ -1,38 +1,93 @@
 #!/usr/bin/python
 
 """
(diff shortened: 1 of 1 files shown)


cyberium:
Thank you but we wil not add python.
Maybe this should be native in our c++ code

PR #83, opened 2014-08-16
Fixed the segmentation fault during shutdown

Mezoriti:
It fixes the segmentation fault that occurs during shutdown of mangosd.

### DIFF
--- a/src/game/ScriptMgr.cpp
+++ b/src/game/ScriptMgr.cpp
@@ -84,7 +84,7 @@ ScriptMgr::ScriptMgr() :
 
 ScriptMgr::~ScriptMgr()
 {
-    UnloadScriptLibrary();
+    //UnloadScriptLibrary();
 }
 
 // /////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


cala:
I still have a segmentation fault whenever I shutdown the server, usually with `.server shutdown 1`

Phatcat:
@cala so did this PR fix the error for you, if you are still able to reproduce this?

Jose123456:
I'm running into this problem now that I figured out [how to properly load the ScriptDev2 library](http://cmangos.net/thread-7481-post-46739.html#pid46739).

Unfortunately this pull no longer applies because the code it modifies has changed a lot. Since the ScriptMgr is initialized in World::SetInitialWorldSettings() now, I believe the proper place to unload the scripting library is in World::CleanupsBeforeStop().

I've made those changes here: https://github.com/Jose123456/mangos-classic/commit/96d98b654713bf0341ad1094cb00fa8fc473951e

PR #88, opened 2014-08-18
Fix Paladin Bonus Healing not scaling with gear with Flash of light and Holy Light

meladath:
Since Holy Light and Flash of Light are handled in their own, they need to have bonus healing added just like other healing does in SPELL_EFFECT_HEAL.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -3644,13 +3644,20 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
         }
         case SPELLFAMILY_PALADIN:
         {
+            Unit* caster = GetAffectiveCaster();
+            
             // Holy Light
             if (m_spellInfo->SpellIconID == 70)
             {
+
                 if (!unitTarget || !unitTarget->isAlive())
                     return;
                 int32 heal = damage;
                 int32 spellid = m_spellInfo->Id;            // send main spell id as basepoints for not used effect
+
+                heal = caster->SpellHealingBonusDone(unitTarget, m_spellInfo, heal, HEAL); //Potential bug fix, added bonus healing from
+                heal = unitTarget->SpellHealingBonusTaken(caster, m_spellInfo, heal, HEAL); //caster and source as in spelleffect HEAL 10
+
                 m_caster->CastCustomSpell(unitTarget, 19968, &heal, &spellid, NULL, true);
             }
             // Flash of Light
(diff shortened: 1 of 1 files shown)


meladath:
Haven't had the spare time to even glance at mangos for months sadly. Very surprised this commit hasn't been looked at. Any news?

Schmoozerd:
Infact it is very good to poke such patches now and then. Usually if one has not the "yay, i do it right now" feeling things get a bit overlooked, sadly.

Anyhow, i cannot tell much about this patch. Only comment i have is:
why does/ is this so/ should it be so the spell that is cast with CastCustomSpell not use the Bonus-Calculation?

meladath:
I am not entirely surely. I tested the healing without and without additional healing gear on. Without adding in this patch, the bonus healing did nothing, with it, it increased normally (well, seemingly normally, I can't confirm it is 100% blizzlike but it looks far more like it than before).

I didn't look into the actual spell casting functions themselves, simply the parts leading up to them, it seems that when other spells are cast they get the bonus added shortly before it calls the cast function, but since FoL and HL are custom spells, they are handled in this switch which then does not add on those bonuses.

I am unaware if the CastCustomSpell should add the bonus itself internally or not as opposed to applying it before the call like other spells, but if thats the case it is not working as intending and may be breaking all instances of CastCustomSpell too?

Schmoozerd:
Hmm, i think i know the difference:

https://github.com/meladath/mangos-classic/blob/8afb1afdce79b4883f1dd06db5841aad53f38076/src/game/SpellEffects.cpp#L1664

_Maybe_ this should rather be m_triggeredSpellInfo instead of m_spellInfo (if the spell is triggered)

poke @VladimirMangos about this, do you have any thoughts?

VladimirMangos:
In genral for bonuses expected spellInfo ofc. BUT as i remember, discussed spell have different (maybe, not check dbc) data in different game versions and have same problem as discibed in wotlk case for healing totem:
*// Healing Stream Totem
*// spell have SPELL_DAMAGE_CLASS_NONE and not get bonuses from owner, use main spell for bonuses
where we use for special case triggered spell infor as you point above.
I not sure for generic case apply (need check all healing triggred spells) but this specific case fit same problem, possible we can use generic code (m_triggred ? m_ttrrigere : m_spellInfo) for bonus calculation but Unit::SpellHealingBonus* have spell specific code caluclation cases and need be sure that replace will not break this special cases apply....

meladath:
Any progress on this? Uni finished soon so I might be able to start looking into cMANGOS again at some time soon, but this is a pretty big issue currently, and even if this fix isnt accepted I think the issue should be looking into.

cala:
Also poke myself to look at this and @Tobischinski since we discussed this yesterday.

Kelzior:
Works good, thanks

Phatcat:
poke @cala you and tobs ever figured out something?

Warlockbugs:
Thank you, but this PR is incorrect. Bonus coefficients are properly processed for these triggered heals, however, auto-calculated coefficient was incorrect for classic. Triggered spell is instant and uses default coefficient for instant spells, rather than expected 2.5s one. This was fixed in https://github.com/cmangos/mangos-classic/commit/72b6848dd61265f33d80192f173680df7c8d731a

PR #90, opened 2014-08-29
Fix a client freeze caused by malformed paths.

Patman64:
Some mmaps-generated paths contain points which are so close together that
the offset from the destination (when truncated) is zero. This was causing
the client to seize up.

Fixes cmangos/issues#235.

### DIFF
--- a/src/game/movement/packet_builder.cpp
+++ b/src/game/movement/packet_builder.cpp
@@ -73,19 +73,30 @@ namespace Movement
 
     void WriteLinearPath(const Spline<int32>& spline, ByteBuffer& data)
     {
-        uint32 last_idx = spline.getPointCount() - 3;
+        uint32 pointCount = spline.getPointCount() - 3;
+        uint32 last_idx = pointCount;
         const Vector3* real_path = &spline.getPoint(1);
         Vector3 destination = real_path[last_idx];
 
+        size_t lastIndexPos = data.wpos();
         data << last_idx;
         data << destination;
         if (last_idx > 1)
         {
             Vector3 offset;
             // first and last points already appended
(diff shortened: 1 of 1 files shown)


cala:
Greetings gentlemen,

I understand that there is not an agreement about this PR: should it be done this way since the issue came from Blizz format? Should it be done in a more proper way?

However, this PR fixes an issue that is **really** annonying since it can lead to heavy freeze or instant crash of the client anytime a creature with movement get a malformed path. And it happens often, randomly and unexpectedly. To be honest, I never build a playable server without this patch. As a note, I did not encounter this issue on TBC or WotLK.

So, between the "fix this properly" side and the "fix the issue" side, here is my proposal:
- merge the pull request
- add a "FIX ME" notice and a bit of explanation into the code to keep track that this needs to be reworked
- open an issue on repository explaining that this part of the code needs to be reworked and why
- link the new issue to the merged commit and this discussion in order to keep a trace of everything

Unless someone oppose to this proposal, I'll do it soon(tm).
Thanks :smile:

Patman64:
The patch is fine. Paths that are valid in mangos can become invalid when converted to WoW's format, so I added code to ensure validity *when the conversion happens*.

You can modify all the path generating code in mangos to never generate a path that can't be sent to the client, but it seems like an error-prone and round-about way to accomplish the same thing.

cala:
In https://github.com/cmangos/mangos-classic/commit/04d5e3624e6649c297660b49ff61641f9d9e245d thank you.
As said, I close this pull request but let the related issue open: https://github.com/cmangos/issues/issues/235 so the discussion can continue in case we want to do it in a better way. From my point of view, this patch works fine: I've used it daily since months without any problem. No performances issues, no crash or freeze.

PR #92, opened 2014-10-05
Fix stealthing animation for group members.

hallabro:
Previously stealthed party members would go invisible for a brief moment
which causes some animation issues. The stealthed players should be
visible during transition for other group and raid members.

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -6548,6 +6548,11 @@ bool Unit::isVisibleForOrDetect(Unit const* u, WorldObject const* viewPoint, boo
     if (m_Visibility == VISIBILITY_OFF)
         return false;
 
+    // grouped players should always see stealthed party members
+    if (GetTypeId() == TYPEID_PLAYER && u->GetTypeId() == TYPEID_PLAYER)
+        if (((Player*)this)->IsGroupVisibleFor(((Player*)u)) && u->IsFriendlyTo(this))
+            return true;
+
     // raw invisibility
     bool invisible = (m_invisibilityMask != 0 || u->m_invisibilityMask != 0);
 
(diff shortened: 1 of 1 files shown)


cala:
@hallabro I (finally) tested your PR this week-end and on both Classic and TBC cores and it works very well. The stealth animation is indeed far better. Thanks! :+1: 

We wanted to see is WotLK core also requires this fix, but the stealth animation already looks fine on that core.
So, in order for us to understand precisely how your fix work and why WotLK seems to not need it, could you please explain a bit further how it works? Thus, we can know if we only push it to Classic/TBC or also on WotLK (even thought it has no *visible* effect). Thank you!

hallabro:
As I remember it `isVisibleForOrDetect` is used by the core at some point to determine if a unit should be marked as visible and sent to the client.

Before this patch I think there are a few cycles where `m_Visibility == VISIBILITY_GROUP_STEALTH)` where is `false` - because `m_Visibility` will only have changed once the stealth has taken place (and the unit has been removed from the client, as the visibility check failed).

The stuttering animation as seen before comes from the unit being deleted from the client, and re-added only a few cycles afterwards.

So basically this patch checks if the target is a party member before allowing the unit to get deleted from the client in the first place, if that makes sense.

cala:
That is very clear. Thank you for this quick answer!

As I understand it, we need to check that WotLK core has the same visibility check cycles than TBC/Classic cores and if client unit removals are also similar. Based on that, we can decide if the cores are behaving differently or if the differences relies in the clients.

Thanks!

cala:
Bump for myself to look at WotLK code.

cala:
I'd like to have the opinion of @cyberium or @Schmoozerd or another senior dev if they can spare the time.

This patch fixes this issue: on Classic and TBC, when a player goes to stealth, he/she becomes invisible for party members for a few seconds before reappearing in stealth state. The video posted above by hallabro shows very well this issue and the fixed state. The issue does not exist on WotLK.

The patch fixes the issue for Classic and TBC. It has zero effect on WotLK (there was no issue in WotLK and the patch broke nothing when I tested it).

I look at the code and saw zero difference between Classic, TBC and WotLK core for this part of the code. My guess is that the WotLK client handles player's stealth/invisibility differently in WotLK compared to TBC/Classic but I saw no explanations to this in Blizz patch notes and I'm not a client expert either.

I think it is safe to push this patch to Classic/TBC core. I also think it would be proper to add it also to WotLK to keep the same code everywhere. I also think that the way stealthed player visibility is handled by the patch is cleaner than what it previously was (less risk of false positive, in my humble opinion).
But, WotLK player stealth in group is ok without the patch, so I would like some other dev feedback on pushing or not this patch to WotLK also. Thanks.

VladimirMangos:
I agree that code more clean and safe for apply to all core versions.
As i remember reason for place check to current point has been in attempt avoid check in most cases for performence (it's very often called function). But if this triggering problems then better use suggested code version.

cala:
@VladimirMangos Thank you very much for you answer!

cala:
In https://github.com/cmangos/mangos-wotlk/commit/26a76bdfe5d42a4c19fc703613963a62f447c6e9 pending backport to TBC/Classic. Thanks everyone.

evil-at-wow:
As you can see, the backport to TBC was done. And I agree that it looked much better when I tested it. Nice job!

cala:
Thanks @evil-at-wow for the backport, I'll do the batch for Classic soon(tm).

PR #97, opened 2014-10-12
Update AccountMgr.cpp

Flying--Dutchman:
Username wasn't changed to uppercases, therefore the SHA1 hash is always wrong when users have lowercases in their username. 
Result: Users cannot log in with telnet.


cala:
Thank you.
In https://github.com/cmangos/mangos-wotlk/commit/a21b5ce53a041e6f0e88bb37167f4abe95a7c2aa

PR #98, opened 2014-10-12
New group leader assignment. No more raid/dungeon disbands req.

DSkywalk:
I create this patch some years ago and is partly based on Koani patch.

The patch allows change the leader if it is disconnected. But lets the leader keep his place if he returns in time (this timer could be configured by the admin).

Thanks and keep the good work :+1:


Phatcat:
@DSkywalk You still have any interest in this?
If so, then if you rebase this and resolve the merge conflicts, as well as make the non-blizzlike things in this off by default and add an option to switch it on (in the configs, like other such changes), then I see no problem with this and won't mind testing it and also eventually merging it (as long as it doesn't make it harder for us to achieve true emulation, since that is our goal, then configurability into just about anything is always welcome if you ask me)

killerwife:
@Phatcat @Warlockbugs implemented smth similar recently. Maybe ask him if its needed at all?

Phatcat:
I thought I had just recently read about the passing lead on disconnect not long ago.
this would in any way need to be rebased, but this PR still did contain a request for more configurability with this, and that change I still think he should have the chance to restructure so it could be implemented.

Warlockbugs:
I've implemented a new group leader choosing mechanism when group leader leaves the group in a similar PR for TBC with proper officer priorities. All that's left to do there is to add a timer with a check for leader being offline for 5 minutes.

Phatcat:
I guess it's somewhat safe to assume that we are going to go with your solution then, @Warlockbugs 
 I would be filled with joy if you would implement that configurability (off by default, of course) and give a small nod to @DSkywalk for his idea in the commit message or somethin' the like, maybe..
I am gonna close this, then. Sorry @DSkywalk , this is by no means anything personal or in any means a reflection upon your code.

DSkywalk:
np, thanks to review my code two years later ;-P

PR #101, opened 2014-10-31
Fix player skills capping at 300/305 with racial bonuses

scotty0100:
GetBaseSkillValue should not return with added bonus instead it should remove any existent bonus.

(Else it's not base value)


cala:
Poke myself to look at this, before or after my holidays.

cyberium:
This is not so simple. It require more research/work.

PR #103, opened 2014-11-30
Use of trap gameobjects on friendly units (like Basic Campfire)

z0ldrax:


cala:
I think this should be pushed to master first and then backported to TBC/Classic.

@VladimirMangos if you are OK with the last changes made by @z0ldrax to the pull request, I'll push it to WOTLK master.

VladimirMangos:
Why bool var defined before if(IsPositive())?

Only see one possible improvement for more clear code: move spell classification code to new inline function into SpellMgr.h. From other side this can be single case when it used so maybe no reason do.

PR #104, opened 2014-12-20
Fix MeleeDamageBonusTaken coeff calculation

Laizerox:
This change fixes cases when pVictim->MeleeDamageBonusTaken is called to
get Victims (In this case creatures) AURA_MOD_DAMAGE_TAKEN spell coeff
calculation.

Previously or as it is now when we enter MeleeDamageBonusTaken we will
never calculate coeff based on caster and this will return in bad
behaviour when fighting against Creatures in PVE Environment.

This applies to all cores.


cala:
I'm closing the PR: in master https://github.com/cmangos/mangos-wotlk/commit/fff42711c155f13604f0538b726fe89ce6a72467 and now pending backport to Classic core. Thanks to all of you. :)

PR #105, opened 2015-03-01
Store cast item GUID in Spell to prevent crash when item is deleted

mityada:
If cast item is somehow deleted during a spell cast, a crash may occur when the spell is fired. The easiest way to reproduce it is to use the hearthstone and then throw it away until the cast finishes.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3723,6 +3723,7 @@ void Spell::TakeReagents()
                 }
 
                 m_CastItem = NULL;
+                m_CastItemGUID.Clear();
             }
         }
 
@@ -6261,6 +6262,7 @@ void Spell::ClearCastItem()
         m_targets.setItemTarget(NULL);
 
     m_CastItem = NULL;
+    m_CastItemGUID.Clear();
 }
 
 bool Spell::HasGlobalCooldown()
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -2477,6 +2477,9 @@ void Spell::prepare(SpellCastTargets const* targets, Aura* triggeredByAura)
 {
     m_targets = *targets;
 
+    if (m_CastItem)
+        m_CastItemGUID = m_CastItem->GetObjectGuid();
+
     m_spellState = SPELL_STATE_PREPARING;
 
     m_castPositionX = m_caster->GetPositionX();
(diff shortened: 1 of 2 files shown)


Patman64:
Wow, good catch.

VladimirMangos:
Maybe good have some functions for hide direct access to item pointer and guid (get, clear, set) for avoid future typos and bugs in related code...

Patman64:
Also: similar problem exists for GOs (e.g. start mining a node, someone else loots it and it disappears, you continue mining even though it's gone).

Phatcat:
Is there a reason we did not merge this?

cala:
The reason is probably a lack of manpower to review it, like many other interesting pull requests. I often remind myself to look at those when I can spare the time but my todo-list usually fills faster than I can empty it... :(

Phatcat:
Alright so I tested this and it works. I also looked at the code and it is very minimal and direct. I think we should consider merging this as I see no regression by doing so. All it would take is changing the NULL to a nullptr.

Phatcat:
And also move the newly added contents of UpdatePointers to its own function as Vladimir suggested, maybe a void Spell::UpdateOriginalCastItemPointer() ?

mityada:
@Phatcat, it seems to me that Vladimir suggested a more global solution, such as a wrapper class for GUID with a method to retrieve valid pointer. There is quite a lot of places (particularly in scripts) where pointer is stored between the world ticks, and it's never safe. It will also help to simplify code where GUID is already used.

VladimirMangos:
Just GUID use can be not optimal case for performens, but most safe maybe. But store GUID + chached pointer with update in appropriate places can cleanup code and prevent unsafe use direct pointer var access. In Spell lot cases when it can be used: m_unitTarget + m_unitTargetGUID and similar. Maybe template<class T> class { Guid, T* ptr } wiht appropriate member-functions for hide like pairs...

Phatcat:
Well, you two seem to know what you're talking about a lot more. I don't see why it would hold back this pr in particular, though, as what you are talking about do sound a lot more general?

Phatcat:
It may not be safe to store pointer between world ticks, but correct me if I'm wrong in saying it should be a hell of a lot more safer than crashing, no?

mityada:
@Phatcat, the thing is, it is not safe just because it will crash (to be more precise, will cause undefined behaviour) when you dereference pointer after underlying object is deleted.

Lillecarl:
https://github.com/cmangos/mangos-classic/commit/cb5aefa223f38d622d33564663c7f82bfc418a36
https://github.com/cmangos/mangos-tbc/commit/4f28d3372c09054a3fbac43a2b6ac24177ef84c6
https://github.com/cmangos/mangos-wotlk/commit/e8638d5cefc95c7b96c735a6a3678ca15103319a

Merged.

PR #109, opened 2015-08-17
Disarm trap spell fix

leewalkergm:
The Disarm trap spell had an unlimited range, so I forced it to it's proper range.


leewalkergm:
@Xfurry can this be checked over?

Phatcat:
This seems to me like a hackfix, adding a specific case to a general function. This is not the way to handle it.
Also don't use tabs, use 4 white spaces instead.

cyberium:
Yes we cant add such kind of code.

PR #110, opened 2015-09-04
Use real unit movemenflags when building movement

Laizerox:


ryancheung:
This would cause client crash if you teleport to felwood!

Laizerox:
Yes I'm aware that it currently crashes client. It's because of splines (most likely) and I'm looking into it with @Tobschinski

Muehe:
May I humbly ask why you closed this PR? Was related to cmangos/issues#729, right?

Laizerox:
Fixed in cmangos/mangos-classic@1def0a8 and cmangos/mangos-classic@0d95ddc

PR #111, opened 2015-09-09
[z2682] Resolve error for new installation

Muehe:
Occurred after two new columns were added in https://github.com/cmangos/mangos-classic/commit/cd32eab75c5beaaa176276f38f1474670d8788bc for fresh installations without preexisting mangos DB. See http://cmangos.net/thread-7203.html


cala:
In https://github.com/cmangos/mangos-classic/commit/cab89a452d02aa4b325736ed147e013c390ecc67
Thank you.

PR #112, opened 2015-09-28
Fix for branching within exclusiveGroup questline

killerwife:
Fixed a special case when quest A has prequest B, that has
exclusiveGroup<0, and exclusiveGroup relates to quest C, hence quest A
should continue just after parent activation/completion not whole
exclusiveGroup completion.


cala:
In https://github.com/cmangos/mangos-classic/commit/2414114bcc585dd4a691b37a02528cb32c2ad232
Thank you. :+1:

PR #115, opened 2016-03-01
Feline swiftness manual fix (Fixes #434)

leewalkergm:
This is a manual fix for Feline swiftness when not in form.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4674,7 +4674,7 @@ SpellCastResult Spell::CheckCast(bool strict)
                             !((Player*)m_caster)->CanUseBattleGroundObject())
                         return SPELL_FAILED_TRY_AGAIN;
 
-					if (((Player*)m_caster)->GetDistance(go->GetPositionX(), go->GetPositionY(), go->GetPositionZ()) > 5)
+					if (m_spellInfo->Id == 1842 && ((Player*)m_caster)->GetDistance(go) > 5)
 						return SPELL_FAILED_OUT_OF_RANGE;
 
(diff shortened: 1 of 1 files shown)


leewalkergm:
This depends on PR#109

Phatcat:
@leewalkergm No, what? those prs are completely unrelated. This pr does not depend on your trap pr. I see that you have included that pr in this pr as well, though, which is a whole other mistake and indicates that you branched it off your disarm branch instead of master for this feline swiftness pr. That's not how you should do it. Also what you are doing here with the druid is out of line with what is being done in the special cases of the generic and other class cases. This is probably not how we should be handling this, is it really?

src/game/Spell.cpp
```diff
@@ -2685,6 +2685,19 @@ void Spell::cast(bool skipCheck)
                 AddPrecastSpell(25771);                     // Forbearance
             break;
         }
+		case SPELLFAMILY_DRUID:
+		{
+			switch (m_spellInfo->Id)
+			{
+			case 24866:
+				if (!m_caster->IsInFeralForm())
```
SatyPardus: This is still wrong. You have to check if the current form is FORM_CAT and not if it is any form.

kartoffelsup:
Spells: **17002** (Feline Swiftness 1) and **24866** (Feline Swiftness 2) have the **Stances** flag set to 1, and **Cat Form is also Stance 1**. Now, I don't know where this check should be placed but I believe this would be a better approach. (If I'm not mistaken, the **Stances** Spell flag is a requirement?)

cyberium:
implemented in another way thank you

PR #116, opened 2016-04-15
Pet fixup round 1 (Pet.cpp, SpellEffects.cpp)

Phatcat:
Should apply the correct health and mana to guardians according to the db table entries instead of basing them off a static formula.

The reason for this fix is in part to correct the Silithid Swarm summons, which a couple of mobs in Tanaris casts. The Silithid Swarms should only have 1 hp, as they are when being summoned on spawn. but when being summoned in fight, they are being set as guardians, and by that have a different amount of hp, one that does not seem fair nor blizzlike at all.

To reproduce the issue I am talking about teleport to Tanaris, go to the Noxious Lair area, pick a fight with a Swarmer and get ready.

This fix appropriately sets any guardians health and mana to whatever it is listed to be in the database.

*Update:
Commit 2: the second commit removes the custom formulas for health (if pet level info is not in db) and damage, and instead bases health, mana and damage off of the db entries.
The only reason for changing SpellEffects is because guardians are by default ALWAYS summoned as the level of the caster, which is not how it's supposed to function when a creature summons a guardian.

*Update 2:
Squashed into a single commit.

*Update 3:
Added a few changes in a new commit.

*Update 4:
Corrected all summons, meaning that all pets, guardians, whatever, have correct levels. this goes for hunter pets as well.

Shutting down PR and my entire fork and reforking again. Got branch issues on my end not allowing me to work on other prs either.


killerwife:
Seems legit, tho tabs need to be removed.

Phatcat:
And so they have been. thanks for the head's up ;)

cala:
Hi,

Thank you for this pull request. The issue you are addressing there are valid. However, @cyberium already start to work on a big rework of the pet system. His work was stalled for the ACE removal process but I have good hope he will be able to resume it now.

To avoid your PR to collide with his rework, I suggest you take contact with him. :smile:

cyberium:
I will resume it for sure, but when...
What i can say at first view is @Phatcat is in the right direction.

Phatcat:
Yeah... I think I'm working on pets.

killerwife:
Possibly try to work out something. @cyberium he has been hell bent on solving this from day one. Since you are busy IRL maybe try to steer him in the right direction and I'll help him as well and then you can take a look later. He seems to write even safer code than I do.

PR #121, opened 2016-04-27
Add outstring_log function with empty args

0x2aff:
Why there is no outstring_log() ?
Everyone loves less compiler errors :)


Phatcat:
What? to get less compiler errors you fix the errors, you don't stop showing them...

cyberium:
iam affread @Phatcat is right this resolve nothing

Laizerox:
But compiler error is that we are passing empty string to a sprintf to create empty line. Which ia just a warning.

Phatcat:
Alright, that makes a lot more sense. Do we really want a new function just to get rid of a few warnings, though?

PR #122, opened 2016-04-27
Fix mmap-extractor

0x2aff:
Didn't compile on linux. Don't know if compilation worked on windows.


cyberium:
should be fixed Thanks

PR #123, opened 2016-05-02
Breath timer

Phatcat:
Base duration of breath should be 60 seconds in Vanilla and TBC, not 3
minutes as it was changed to in patch 3.0.2 for Wrath (and how it
currently is on cmangos classic)

http://wowwiki.wikia.com/wiki/Breath?oldid=335923

http://wowwiki.wikia.com/wiki/Breath

It even states in the uncached, current version of the wiki, my last
link, which patch it was changed in, and what the duration was prior to
that; 60 seconds.

The forsaken racial trait should be changed to reflect this, meaning
that undead should have 3 minutes base, regardless of what they had
before, as my first link clearly states. Someone point me in the
direction of this and I'll change it.


cala:
Good catch.

cala:
Regarding the Forsaken racial traits, the spell DBC should already handle that: I doubt this would be hardcoded (I may be wrong).

PR #125, opened 2016-05-02
Fix "Stoneform" - dwarven racial ability.

TarasHots:
This spell did not remove disease and bleed effects.

Fixes https://github.com/cmangos/issues/issues/896.


Phatcat:
Also tabs would need to be replaced with 4 whitespace characters.

TarasHots:
Thanks for quick feedback! Working...

TarasHots:
After investigation about SPELL_ATTR_EX_DISPEL_AURAS_ON_IMMUNITY flag, I've found out that it used to dispell auras by some specific type, passed in misc value. Stoneform spell have misc value set to 4, that corresponds to DISPEL_POISON. Indeed, everything works beautiful with poisons... Immunes are added, and they are dispelled if any. I have several questions about further work on this:

- Do I understand mentioned above correctly?
- As I see, ApplySpellDispelImmunity method still should be called for poison and disease...
- Not sure what to do with bleed mechanics. Their logic should be placed out of dispell-related things. Something like HandleModMechanicImmunityMask woud be great, but those methods are dependant from misc value, hance cannot be usable here

Please, correct me where I am wrong.

VladimirMangos:
Maybe some more spells with appropriate auras exist linked related to spell with expexted misk values?

TarasHots:
@VladimirMangos nope. Check [this screen](https://i.gyazo.com/011a77bea5bdde1ae902521cb73053db.png)

TarasHots:
Please, check my latest version. I've added method similar to ApplySpellDispelImmunity but for mechanics. It adds needed mechanic, and works with SPELL_ATTR_EX_DISPEL_AURAS_ON_IMMUNITY flag if present...

VladimirMangos:
Look like mechanic case is right direction!

TarasHots:
Guys, share your thoughts about this please. Do you agree that these changes are reasonable? Am I missed smth?

I've additionally checked other similar methods, and it seems the only way to do it, I guess. I've also checked another Stoneform's attribute - SPELL_ATTR_EX_UNAFFECTED_BY_SCHOOL_IMMUNE - that seemed not necessary in this case...

VladimirMangos:
SPELL_ATTR_EX_DISPEL_AURAS_ON_IMMUNITY apply to mechanic is proper way look like as I say early, but i meaning case process mechanic immunity in same way
Stoneform have another spell 20612 with SPELL_AURA_MECHANIC_IMMUNITY (77) misc = 15 ( MECHANIC_BLEED ). So for mechanic part rthgsi spell msust be applied proeprtly with main spell. Also spell have dispell missing part

TarasHots:
@VladimirMangos I see your point and completely agree about that way of implementation. But, how 20612 spell is linked to 20594? There are no parent/child connection and 20612 is not triggered by 20594. Additionaly, method responsible for SPELL_AURA_MECHANIC_IMMUNITY is not triggered after Stoneform' usage...

VladimirMangos:
second spell expected to be applied/remove with main spell, so you can look at SpellAuraHolder::HandleSpellSpecificBoosts

TarasHots:
@VladimirMangos really thanks for pointing the correct way. HandleSpellSpecificBoosts made everything smoothly and cool. I haven't encountered any stacking issues and it works as intended

VladimirMangos:
Commites into https://github.com/cmangos/mangos-classic/commit/0181ff1332a78e707d65d5992f1d79ff7af5bd48
Thank you)

PR #127, opened 2016-05-11
Stop npcs from auto attacking while casting

Phatcat:
Fixes the non AI parts of [pr 916](https://github.com/cmangos/issues/issues/916)

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/127)

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -4879,20 +4879,31 @@ bool Unit::Attack(Unit* victim, bool meleeAttack)
     if (!isAlive() || !victim->IsInWorld() || !victim->isAlive())
         return false;
 
-    // player cannot attack in mount state
-    if (GetTypeId() == TYPEID_PLAYER && IsMounted())
-        return false;
-
     // nobody can attack GM in GM-mode
     if (victim->GetTypeId() == TYPEID_PLAYER)
     {
         if (((Player*)victim)->isGameMaster())
             return false;
     }
-    else
+
+    if (GetTypeId() == TYPEID_PLAYER)
     {
-        if (((Creature*)victim)->IsInEvadeMode())
+        // player cannot attack in mount state
+        if (IsMounted())
             return false;
+    }
+    else 
+    {
+        // stop creature from unsheathing their weapon
+        // also make sure to keep resetting any melee timers to avoid insta hits between mob spellcasts
+        if (IsNonMeleeSpellCasted(false))
+        {
+            SetSheath(SHEATH_STATE_UNARMED);
+            resetAttackTimer(BASE_ATTACK);
+            resetAttackTimer(OFF_ATTACK);
+            resetAttackTimer(RANGED_ATTACK);
+            return false;
+        }
     }
 
     // remove SPELL_AURA_MOD_UNATTACKABLE at attack (in case non-interruptible spells stun aura applied also that not let attack)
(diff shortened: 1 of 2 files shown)


Phatcat:
Okay, in all seriousness. To fix this properly we have to look at the AI and what is causing the creatures to keep attacking even though they are casting. This fixes the issue, but the unit is still trying to melee every chance it gets, which cannot be easy on the server, so for optimization and to avoid the 'hackfix' of unsheating and resetting attack timers, the AI has to be changed to stop trying to melee. The AI has to be rewritten, though. I will be looking into the AI when I am done with pets and engineering.

src/game/Unit.cpp
```diff
@@ -339,7 +339,7 @@ bool Unit::UpdateMeleeAttackingState()
     if (!victim || IsNonMeleeSpellCasted(false))
         return false;
 
-    if (!isAttackReady(BASE_ATTACK) && !(isAttackReady(OFF_ATTACK) && haveOffhandWeapon()))
+    if (!isAttackReady(BASE_ATTACK) && (haveOffhandWeapon() && !(isAttackReady(OFF_ATTACK))))
```
VladimirMangos: I think wrong change.  Check expected skip mo any weapon attack cases . In final form it true only if  unit have offhand weapon and so always false for unit without offhand weapon

Phatcat: Actually, the outcome of the two conditions is exactly the same, mine is just cleaner as it clearly shows what it does, where the initial one is a bit obscure.

It basically just went from a NAND to a more transparent AND NOT function instead.

katz1:
@Phatcat 
I tested this a bit. The mobs are using their casting animation now which is nice, but the attacking is still strange. They are occasionally still doing the one hit between two casts and even if they don't, their weapon is often unsheated after cast and than sheated sometime during the next cast.

If you are rewriting the AI in the future, I'd also consider lowering the priority on spell casting in general. According to most vanilla videos I've seen, they are really not throwing spells that often.

Phatcat:
@katz1 You are right this is not perfect. and you are right that spell priority is too high, under some circumstances. the thing you mention about mobs hitting between spells. they did this in vanilla as well. They went straight for meleeing when they had finished a spell, just like players do. And the attack timers are reset to the moment the spell stops so they should not be able to insta hit.
their weapon should be unsheathed after a spell cast. this is not just mob, but unit behaviour.

Phatcat:
> According to most vanilla videos I've seen, they are really not throwing spells that often.

This is because their priority for meleeing vs casting also has range factored in. Very clearly. We do not have this factored in, in any way.

katz1:
@Phatcat 

> And the attack timers are reset to the moment the spell stops so they should not be able to insta hit.

Yes, I forgot to mention that. They no longer hit instantly. In that case doing the one hit between two casts is acceptable.

> their weapon should be unsheathed after a spell cast

I find this a bit strange, but I can imagine that how it was. In vanilla they probably rarely casted 2 spells completely simultaneously, so it wasn't popping in for a moment that often? Regarding the unsheathing though, it seems to me that now there is a short time after they already started casting the spell, yet their weapon is still not hidden.

Phatcat:
@katz1 That may actually just be a visual delay. I have noticed this too, yet this behaviour is randomly occuring, and should not even be possible, so it is most likely just a visual glitch caused by the server lagging behind a bit. There is sadly not much more to do without touching up on the AI, and I rather wanna look into a complete rewrite.

Phatcat:
This is only the beginning.

PR #129, opened 2016-05-14
[MANGOS-CLASSIC] Fixing bug where spells remove themselve on cast.

SatyPardus:
It has to do with UnitAuraProcHandler->HandleRemoveByDamageChanceProc - So there is probably a better way.

This is my first fix and my first day with C++, any tip or better way would be awesome :)

### DIFF
--- a/src/game/LootMgr.cpp
+++ b/src/game/LootMgr.cpp
@@ -1352,7 +1352,7 @@ void Loot::GroupCheck()
                     continue;
 
                 if (*itr == m_masterOwnerGuid)
-                    masterLooter == looter;
+                    masterLooter = looter;
 
                 data << *itr;
                 ++playerCount;


Phatcat:
@SatyPardus Nothing wrong with the syntax, but you probably know that already.
To suggest a better solution we have to know just how and what is bugged. Could you link the issue you are fixing or perhaps one of the spells you suggest to fix with this pr?

SatyPardus:
The bugfix is for this issue: https://github.com/cmangos/issues/issues/890
But it probably should fix more stuff that works similar like Frost Nova. When there is any similar.

The problem was, that frost nova got canceled to the own damage it deals. I tried to fix it first by moving the aura to a later execution, but i didn't find a way (i don't have enough c++ knowledge for that)

Phatcat:
Since you mention yourself it has to do with function HandleRemoveByDamageChanceProc
, so why not move your condition to [the function](https://github.com/cmangos/mangos-classic/blob/22b0a55d3384cf8905f76260a8f1fd3aab7b130f/src/game/UnitAuraProcHandler.cpp#L1365) that calls RemoveAurasByCasterSpell?

Here you have the Aura* triggeredByAura - so you can get the duration and max duration from there, if you believe that is the way to go. I don't know personally.
But by adding your condition to [this](https://github.com/cmangos/mangos-classic/blob/22b0a55d3384cf8905f76260a8f1fd3aab7b130f/src/game/UnitAuraProcHandler.cpp#L1370) instead then RemoveAurasByCasterSpell is not even called unless it's needed. I don't know if -some- auras should be able to be removed just when they have been created. Doing the condition here and not in RemoveAurasByCasterSpell does allow for other auras still to be removed when just having been applied (or has max duration, to be exact).

SatyPardus:
Okay i changed it and added a better comment. (I hope it shows up here)
But i really cant figure out where "HandleRemoveByDamageChanceProc" gets called. The call hirachy is just empty.

Phatcat:
Well, you did not remove your first commit, which means that your condition is now in both functions, which is very redundant. I don't know if you are too familiar with git but if you do a git reset --soft head~2 , then remove your first changes, then only stage and commit your second changes, and then do a git push -f , then it should be updated to only be 1 commit in the pr, with your latest changes and nothing else. Like it should.

Phatcat:
Then you just need to push your changes as 1 commit. You shouldn't have pushed till you had commited your changes, though, or it wouldn't have closed. Sorry if I am unclear with the instructions here..

SatyPardus:
Haha, i normally use the GUI, but that is to limited. I try to figure it out, but now i am on a very old update :D

Phatcat:
If you want to rebase (update) your branch to include all the things from master without losing your commits you do a
> git rebase cmangos/master

SatyPardus:
I justed messed it up completly. I make it new. Git is very good, but also very hard to handle with fork stuff. First time and fail :D

Phatcat:
There is no shame in learning. We've all been there. Just rebranch from master and start over with a fresh new pr - you should also link the issue you are fixing in either your commit comment or your pr comment - that way it gets referenced, but you probably already know that.

SatyPardus:
Okay. I hope now it is okay. How do i prevent git from putting new commits to a pull request?

Phatcat:
You work on different branches. Just keep branching cmangos/master into new local branches. You cannot work on different prs on the same branch.

PR #130, opened 2016-05-15
Fixed a bug where the "Frost Nova"-Root would be canceled by own damage

SatyPardus:
Very hackish way - There is probably a better way.

Issue to this bug: https://github.com/cmangos/issues/issues/890

### DIFF
--- a/src/game/UnitAuraProcHandler.cpp
+++ b/src/game/UnitAuraProcHandler.cpp
@@ -1367,7 +1367,7 @@ SpellAuraProcResult Unit::HandleRemoveByDamageChanceProc(Unit* pVictim, uint32 d
     // The chance to dispel an aura depends on the damage taken with respect to the casters level.
     uint32 max_dmg = getLevel() > 8 ? 25 * getLevel() - 150 : 50;
     float chance = float(damage) / max_dmg * 100.0f;
-    if (roll_chance_f(chance))
+    if (roll_chance_f(chance) && triggeredByAura->GetAuraDuration() != triggeredByAura->GetAuraMaxDuration()) // TODO find a better way to handle this. Without the duration check, it will cancel the "Frost Nova"-Root imidiatly - What shouldn't happen
     {
         triggeredByAura->SetInUse(true);
         RemoveAurasByCasterSpell(triggeredByAura->GetId(), triggeredByAura->GetCasterGuid());


Phatcat:
Well, this makes sure that an aura can't dispell itself on cast, which fixes a heck of a lot more than just that one spell. I wouldn't consider it a hackfix as it doesn't create specific cases, but on the contrary it applies to every case with 2 or more effects where once is damage and the other is a root, stun or fear aura which can be dispelled by damage. A lot of goblin engineering stuff comes to mind, at least.

Could you perhaps test this with different spells that both root/stun/fear and deal damage?

SatyPardus:
We already tested it with 6 players. 2 mages, 1 warlock and pala, druid and rouge. Looks like there is no problem. Only thing is, when someone gets rooted and attacked by another player at the same time, it ignores the root and stuff. But i can be wrong there - maybe it's just luck? But i like it. Fells right

Laizerox:
This is a nice way to patch it but the issue behind this is far more global.

We handle spells wrong. If u look at spell code you will see that we first process effects, then calculate damage and deal damage after applying effects which causes the removal.

Damage and healing calculation should be done in their own effects instead after effects.

I am currently working on designing a new system with @killerwife and I've already stumbled upon that we need to rewrite likely healing / damaging part, proccing and everything else related to it.

cyberium:
Forgive me if i dont understaind correctly this PR.
I think this change will simply remove all the possibility to get any aura removed by damage.

Phatcat:
@cyberium It hopefully shouldn't. It should only bypass the if condition if the duration is the same as the maxduration, meaning no duration has passed, meaning the aura has -just- been applied. that is the theory at least. I have not personally tested this.

SatyPardus:
Works pretty good on my server^^ And no bugs so far from other players

katz1:
I guess this would not work correctly if the target received damage from another source at the exact same time, which should be able to break the effect.

I'll use this as a temporary fix until the spell system is rewritten. Thanks!

Phatcat:
@In that case, it'd not only have to recieve the damage at the same time but also having run the complete shabang to apply the aura, then run the code to apply the damage and check proc. Would it then, when fetching the duration this time, still be equal to maxduration?

katz1:
@Phatcat 

> Would it then, when fetching the duration this time, still be equal to maxduration?

Sorry but I lack the knowledge to answer that. The doubt I had was only theoretical, it may not be technically possible to make a scenario where this fix will not work.

SatyPardus:
The change for that is very minimal, as the server will process the attack from another player in the next "tick", where the duration is already a little bit smaller (enough to cancel the stun)

killerwife:
While this is the kind of fix that we can permit on a private server repo, I think adding this to cmangos would be detrimental since we are trying to do proper fixes here. That being said, keep the PR so it reminds us, and I hope it doesnt stop you from contributing more.

Phatcat:
@killerwife
So you are rather suggesting to fix up on [this function](https://github.com/cmangos/mangos-classic/blob/5355e58655119954241d8a82a09e0481abe818bf/src/game/Unit.cpp#L8116) instead, then?
If you feel like this fix is improper, which it very well may be, then let us talk about how we fix this properly then. Where should we fix this?

Phatcat:
@SatyPardus You also asked for where HandleByRemoveDamageChanceProc is called - it is [here](https://github.com/cmangos/mangos-classic/blob/5355e58655119954241d8a82a09e0481abe818bf/src/game/Unit.cpp#L8248)

cyberium:
I think the proper way should be to rewrite part of spell system to handle damage first.
I think an acceptable way is to "mark" current holder in the way that when we apply damage we recognize aura applyed in another effect index of that same holder and so we dont remove it.

killerwife:
@Laizerox  is rewriting damage processing, no ETA, and it will completely fix this problem, in addition to cleaning up core

Phatcat:
@cyberium what you are talking about should already be possible in the current core, no?

@killerwife in that case then I agree, we should not merge and just leave it open to remind ourselves to test for this when the updated damage processing gets merged.

Phatcat:
This may be interesting regarding this issue: https://github.com/anathemareborn/core/commit/439c6df27d441bb4c130dd290f27e1883e1063ea

killerwife:
Thats the same issue with a different flavor. Will be fixed by the WIP as well.

Phatcat:
Yeah I know. @robinsch commit just seemed a bit more genuine than this, so I thought it might be of interest to @SatyPardus , since he said himself his was probably a hackfix that could be fixed better. This is but one example of how it could probably be fixed with less of a hackfix. 
Glad there's a whole patch in the workings that will touch up on this as well as other issues, though. still felt it worth linking this so Saty could check it out.

PR #131, opened 2016-05-15
Fix for spell duplication after training

katz1:
Issue: https://github.com/cmangos/issues/issues/925

In the spell learning case specified in the issue two packets are sent which cause the learning of a spell:
- The first one from Player::addSpell _(SMSG_SUPERCEDED_SPELL)_
- The second one from Player::learnSpell _(SMSG_LEARNED_SPELL)_

The second packet is not needed in this case, because the first one will replace the old spell and show the "learned" message properly. As the second packet is sent based on the return value of Player::addSpell, it seemed appropriate to alter that.

There are some foggy parts here which I'd love to have verified. First of all the return value of Player::addSpell is used inside Player::removeSpell as well. I think my change stands in that case as well, but I am only about 80% sure. Secondly, I have no idea what is the purpose of this code in Player::addSpell:

```

  else if (sSpellMgr.IsHighRankOfSpell(itr2->first, spell_id))
  {
      if (IsInWorld())                // not send spell (re-/over-)learn packets at loading
      {
          WorldPacket data(SMSG_SUPERCEDED_SPELL, (4));
          data << uint16(spell_id);
          data << uint16(itr2->first);
          GetSession()->SendPacket(&data);
      }

      // mark new spell as disable (not learned yet for client and will not learned)
      newspell.active = false;
      if (newspell.state != PLAYERSPELL_NEW)
          newspell.state = PLAYERSPELL_CHANGED;
  }
```

I couldn't get the code run inside this branch even with gm learn / unlearn commands, so I can't imagine a valid use for it. Maybe I should set the _replacedOldSpell_ flag in this branch as well, but I am really not sure.

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -2982,7 +2982,13 @@ bool Player::addSpell(uint32 spell_id, bool active, bool learning, bool dependen
         }
     }
 
-    // return true (for send learn packet) only if spell active (in case ranked spells) and not replace old spell
+    if (active && !disabled && !replacedOldSpell && learning && IsInWorld())
+    {
+        WorldPacket data(SMSG_LEARNED_SPELL, 4);
+        data << uint32(spell_id);
+        GetSession()->SendPacket(&data);
+    }
+
     return active && !disabled && !replacedOldSpell;
 }
 
@@ -3011,15 +3017,7 @@ void Player::learnSpell(uint32 spell_id, bool dependent)
     bool disabled = (itr != m_spells.end()) ? itr->second.disabled : false;
     bool active = disabled ? itr->second.active : true;
 
-    bool learning = addSpell(spell_id, active, true, dependent, false);
-
-    // prevent duplicated entires in spell book, also not send if not in world (loading)
-    if (learning && IsInWorld())
-    {
-        WorldPacket data(SMSG_LEARNED_SPELL, 4);
-        data << uint32(spell_id);
-        GetSession()->SendPacket(&data);
-    }
+    addSpell(spell_id, active, true, dependent, false);
 
     // learn all disabled higher ranks (recursive)
     if (disabled)


katz1:
I am quite new to GitHub. Does anyone has a clue why did the appveyor check failed and what can I do about it?

cyberium:
>fatal: The remote end hung up unexpectedly
fatal: index-pack failed
error: RPC failed; result=18, HTTP code = 200
Command exited with code 128

its network or communication problem between appveyor and github. you can ignore that.

cyberium:
Thank you for this.
I tryed to follow a bit the code and found that 
> else if (sSpellMgr.IsHighRankOfSpell(itr2->first, spell_id))

is in error that should be fixed by removed the entire if condition.
If i understiand correctly that code is for "downgrade" a spell to its lower rank.

About the PR yes it should fix the problem i dont see regression. Btw i think the best is to rewrite some part of learnSpell and addSpell. I think we should move the handling if SMSG_LEARNED_SPELL in addSpell. If iam correct addSpell when "learning" bool is true have to fire SMSG_LEARNED_SPELL or SMSG_SUPERCEDED_SPELL if the player is in world.

what's your thought?

katz1:
@cyberium 
Thanks for looking into this.

> If i understiand correctly that code is for "downgrade" a spell to its lower rank.

Yes it is, but I didn't find a use of it. I tried unlearning talents (while having talent spells) and professions (with spells) without gm commands. And I also tried unlearning only a single rank of a spell with a GM command. In none of these cases had that block any relevance, which confused me.


> I think we should move the handling if SMSG_LEARNED_SPELL in addSpell. If iam correct addSpell when "learning" bool is true have to fire SMSG_LEARNED_SPELL or SMSG_SUPERCEDED_SPELL if the player is in world.

I think that would make sense. I can try this tomorrow (unless you already started making it).

cyberium:
That part may be called when using something like ".level -1"?

No i didnt started on it just reviewed the actual code.

katz1:
I think addSpell should not have a return value at all, but then it's use in removeSpell should be handled as well:

```
if (addSpell(prev_id, true, false, spell.dependent, spell.disabled))
{
    // downgrade spell ranks in spellbook and action bar
    WorldPacket data(SMSG_SUPERCEDED_SPELL, 4);
    data << uint16(spell_id);
    data << uint16(prev_id);
    GetSession()->SendPacket(&data);
    prev_activate = true;
}
```
This is unfortunately another part of the code I did not manage to run into with debugging. I am not confident in moving this packet sending to addSpell, addSpell would probably need another flag to differentiate it from other addSpell calls? It would be nice to clean this up but I have trouble recognizing the potential side effects.

> That part may be called when using something like ".level -1"?

I gave it a try, but reducing character levels does not affect learned spells.

VladimirMangos:
Also can related to downgrade  dependent spells when main spell/talent downgrade/unlearn.
I not remember exactly case, sorry

katz1:
@cyberium 
Just letting you know that it's unlikely that I will improve this further.

cyberium:
Thank you for your suggestion you already did it :) The rest will be "clarification/rewrite/optimisation".
I'll try to take care of it when i'll have some time.

cala:
@cyberium just assigning yourself on this because I'm looking at/sorting out the currently open PR. So, we know that this one is looked at and by who.  (no hurry ;) )

PR #133, opened 2016-05-18
Added missing include for GCC

katz1:
I mentioned this here:
https://github.com/cmangos/mangos-classic/commit/48a84fbfb6c30d81fb1b7467b071a1512112d50a#commitcomment-17476215

Figured I'll make a PR as well.

### DIFF
--- a/src/game/vmap/BIH.h
+++ b/src/game/vmap/BIH.h
@@ -25,6 +25,7 @@
 
 #include <Platform/Define.h>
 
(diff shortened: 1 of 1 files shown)


wbierman:
<stdexcept> is not used in the .h file.  It should be enough to add it to the .cpp file I think?

cyberium:
thank you this should be fixed

PR #134, opened 2016-05-19
Protecting the Shipment quest fix

Zooke:
Fixes quest 309 Protecting the Shipment by some magic of C++ which I have no knowledge of.
It was just common sense looking at the code that something wasn't right when I compared it to other code and hey presto now it works.
Used in conjunction with https://github.com/classicdb/database/pull/844

### DIFF
--- a/src/scriptdev2/scripts/eastern_kingdoms/loch_modan.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/loch_modan.cpp
@@ -133,11 +133,11 @@ struct npc_miranAI: public npc_escortAI
                 m_creature->SummonCreature(NPC_DARK_IRON_DWARF, m_afAmbushSpawn[1].m_fX, m_afAmbushSpawn[1].m_fY, m_afAmbushSpawn[1].m_fZ, m_afAmbushSpawn[1].m_fO, TEMPSUMMON_CORPSE_TIMED_DESPAWN, 25000);
                 break;
             case 23:
-				if (Player* pPlayer = GetPlayerForEscort())
-				{
-					DoScriptText(SAY_MIRAN_3, m_creature);
-					pPlayer->GroupEventHappens(QUEST_PROTECTING_THE_SHIPMENT, m_creature);
-				}
+                if (Player* pPlayer = GetPlayerForEscort())
+                {
+                    DoScriptText(SAY_MIRAN_3, m_creature);
+                    pPlayer->GroupEventHappens(QUEST_PROTECTING_THE_SHIPMENT, m_creature);
+                }
                 break;
         }
     }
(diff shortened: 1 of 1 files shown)


src/scriptdev2/scripts/eastern_kingdoms/loch_modan.cpp
```diff
@@ -133,9 +133,11 @@ struct npc_miranAI: public npc_escortAI
                 m_creature->SummonCreature(NPC_DARK_IRON_DWARF, m_afAmbushSpawn[1].m_fX, m_afAmbushSpawn[1].m_fY, m_afAmbushSpawn[1].m_fZ, m_afAmbushSpawn[1].m_fO, TEMPSUMMON_CORPSE_TIMED_DESPAWN, 25000);
                 break;
             case 23:
-                DoScriptText(SAY_MIRAN_3, m_creature);
-                if (Player* pPlayer = GetPlayerForEscort())
-                    pPlayer->GroupEventHappens(QUEST_PROTECTING_THE_SHIPMENT, m_creature);
+				if (Player* pPlayer = GetPlayerForEscort())
```
Rushor: pls fix your tabs settings :)

Phatcat:
These are under indented. also you need to use 4 whitespace characters, not tabs, or you need to change your tabs settings in your software to do 4 whitespace characters instead of a standard tab. either way no tabs.

cala:
Thank you for your contribution. However, I see no reason why the text said by the NPC should require that an escorting player is near. This is a "flavor" text and should be visible even if the escorting players died, left, disconnect.

Zooke:
Ahh, you misunderstand how the quest was not working correctly, take the quest then speed run to the hand in, you will see that you can complete it without even doing the escort.
This fix solves that issue by requiring the player to be within the stated range of the escortee, thus requiring a successful escort to complete the quest.

cala:
> This fix solves that issue by requiring the escortee to actually make it to the camp, thus requiring a successful escort to complete the quest.

This is fixed by your [other PR on Classic DB](https://github.com/classicdb/database/pull/844). You make the quest requires an external event (the quest completion sent by SD2) instead of auto-completing. I agree with this one and I should merge it soon.

However, in your PR here, you make the completion speech text of the NPC requires the presence of a nearby escorting player. The text is "*Ah, here at last! It's going to feel so good to get rid of these barrels.*" and I see no reason for requiring an escorting player for that text: no player's name or class, the text does not hint that the NPC is speaking to a player... Said otherwise, I don't think we need this change.

> By the way, I am sorry for the mess I seem to have made of this commit, it's my first time.

No worries, you have no idea how dirty my first PR was... ;-)

Zooke:
Not sure how I confused the 2, but at least it's fixed.

PR #135, opened 2016-05-22
AURA_STATE_HEALTHLESS_20_PERCENT apply only on alive targets

sidsukana:
Fixed problem when spell events by AURA_STATE_HEALTHLESS_20_PERCENT procs on died target
It also can be applied to **mangos-tbc** and **mangos-wotlk**

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -328,7 +328,9 @@ void Unit::Update(uint32 update_diff, uint32 p_time)
     // update abilities available only for fraction of time
     UpdateReactives(update_diff);
 
-    ModifyAuraState(AURA_STATE_HEALTHLESS_20_PERCENT, GetHealth() < GetMaxHealth() * 0.20f);
+    if (isAlive())
+        ModifyAuraState(AURA_STATE_HEALTHLESS_20_PERCENT, GetHealth() < GetMaxHealth() * 0.20f);
+
     UpdateSplineMovement(p_time);
     i_motionMaster.UpdateMotion(p_time);
 }


sidsukana:
any comments on this cosmetics? =)

Phatcat:
Cosmetics look fine. What exactly kind of issues does it solve? what kind of events still proc on dead targets relating to this aura specifically?

sidsukana:
Hammer of Wrath, Execute for example

Phatcat:
Actually, what I meant was, which of these kinds of spells should even be castable on a dead target to begin with? - does it really matter if the target has the aura applied or not? is the issue not with the spells being castable on dead targets?

sidsukana:
No, you can not cast on target. AURA STATE sets a flag for the client side to play the event (in this case, a pop-up label)

Phatcat:
I am not quite sure what this really fixes, then? - What event is it that procs on a dead target affected by this aura since we have to make sure not to apply it? is it that a player having this unit as a target has execute and hammer of wrath lit up as if able to cast these on the target even if it's actually dead? - Is this really an issue with the unit and not how we do initial spell checks?

sidsukana:
It seems that I need to make a video about how it was and how it became, or you do not understand what makes the aura state :)

sidsukana:
1. Target losing hp to 20%
2. Server set aura state AURA_STATE_HEALTHLESS_20_PERCENT to field UNIT_FIELD_AURASTATE
3. Client playing pop-up label "HAMMER OF WRATH" if you Paladin, and "EXECUTE" if you Warrior
4. Target died
Before:
5. Client continue playing event
After:
5. Client does not playing event

cyberium:
@Phatcat you know that execute button should be only enabled on unit less than 20% HP.
To inform client that condition is fullfiled we set a special for that when we update the unit object.
A died unit actualy keep that flag. So when ingame we select a died unit we can see execute button enabled.

@sidsukana is isAlive() is true for feign death unit? (just to be sure)

Phatcat:
@sidsukana
> Before: Client continue playing event
After: Client does not playing event

.... What event? popup-event? but shouldn't the fact that the unit is dead overwrite any other state the unit has making execute and hammer of wrath NOT play the pop-up label? It seems like the check for execute and wrath is bypassing checking if the unit is dead. Is this really how it's supposed to be?

@cala Yeah, I know this much at least. We inform the client the condition is fulfilled, sure, but shouldn't some initial check when selecting the target tell the client that the target is dead, thus not enabling pop-up events for ANY kind of non resurrective spell?

sidsukana:
@cyberium yes is true for feign death unit

PR #136, opened 2016-05-26
Set relative paths to the lib and etc folders to make server relocation possible

katz1:
Issue: https://github.com/cmangos/issues/issues/935

I am still making a few tests, but it appears to be working so far. I was able to relocate my server build and still use it.
I did this without knowing anything about CMake, so any help is appreciated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,17 +83,21 @@ if(CMAKE_CONFIGURATION_TYPES)
     FORCE)
 endif()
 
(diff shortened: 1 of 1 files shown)


cyberium:
Thank you very well for this. I have currently some time issue but i'll merge that for sure asap.

cyberium:
Thank you in b57512dd67facc364350ba16dccf536104c7367c

PR #137, opened 2016-06-01
Fix crushing blows

Phatcat:
Only mobs 3 levels higher can do crushing blows..... not crashing blows
either. Thanks for pointing @Zooke

Ref https://github.com/cmangos/issues/issues/949

~~Second commit rewrites the whole function somewhat as looking at some of the other conditions had me wonder, and after studying the function I just couldn't leave it in it's current state.~~ No second commit, will arrive in another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/137)

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -2213,27 +2213,19 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
         return MELEE_HIT_CRIT;
     }
 
-    if ((GetTypeId() != TYPEID_PLAYER && !((Creature*)this)->IsPet()) &&
-            !(((Creature*)this)->GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_CRUSH) &&
-            !SpellCasted /*Only autoattack can be crashing blow*/)
+    // mobs can score crushing blows if they're 3 or more levels above victim
+    // having defense above your maximum (from items, talents etc.) has no effect
+    // mob's level * 5 - player's current defense skill - add 2% chance per lacking skill point, min. is 15%
+    if ((getLevel() - 3) >= pVictim->getLevel() && !SpellCasted
+        && roll < (tmp = (((attackerMaxSkillValueForLevel - victimMaxSkillValueForLevel) * 200) - 1500)))
     {
-        // mobs can score crushing blows if they're 3 or more levels above victim
-        // or when their weapon skill is 15 or more above victim's defense skill
-        tmp = victimDefenseSkill;
-        int32 tmpmax = victimMaxSkillValueForLevel;
-        // having defense above your maximum (from items, talents etc.) has no effect
-        tmp = tmp > tmpmax ? tmpmax : tmp;
-        // tmp = mob's level * 5 - player's current defense skill
-        tmp = attackerMaxSkillValueForLevel - tmp;
-        if (tmp >= 15)
+        uint32 typeId = GetTypeId();
+        if ((typeId == TYPEID_UNIT && !(GetOwnerGuid() && GetOwner()->GetTypeId() == TYPEID_PLAYER)
+            && !(((Creature*)this)->GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_CRUSH))
+            || (typeId == TYPEID_PLAYER && GetCharmerGuid() && GetCharmer()->GetTypeId() == TYPEID_UNIT))
         {
-            // add 2% chance per lacking skill point, min. is 15%
-            tmp = tmp * 200 - 1500;
-            if (roll < (sum += tmp))
-            {
-                DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: CRUSHING <%d, %d)", sum - tmp, sum);
-                return MELEE_HIT_CRUSHING;
-            }
+            DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: CRUSHING %d)", tmp);
+            return MELEE_HIT_CRUSHING;
         }
     }
 


Phatcat:
We even say in a comment just below the checks that the unit must be 3 levels higher than its target, yet we never check for this... It's not like it was difficult to add, so have someone removed it for some reason or did the guy writing the comment just forget to actually add it in?

Zooke:
This is where the ambiguity really kicks in because if a mob is 3 levels above, it automatically has +15 skills on the calculation if I understand correctly that any bonuses are not applicable for crushing blows.

I just get more confused by the day with this, probably because of our other conversation where I concede that this situation may be purely hypothetical because of the artificial parameters involved in not levelling a character from scratch.

Phatcat:
The reason for the level check may be to spare mid to mid-high level casters who may be lacking behind in the defense skill, meaning that they could potentially suffer from crushing blows from yellow mobs and the like, and I have NEVER seen that happen during my over 100 /played days to boot on retail vanilla. I actually am convinced by now that there should be a level check.

Zooke:
This then leads us to something that we might struggle to find any evidence of.
And that is, were crushing blows ever allowed to be 100%. From a design point of view I would suggest that there was probably a cap, but I am really going out on a limb here.
If I can find any video of say a world boss hitting a low level character and it not being crushing then this would prove a cap.

Phatcat:
I'm actually convinced that would be somewhat moderately easy to find some videos for; just think about how many people made the Darnassus -> Ironforge trip; many were lower than level 10, getting attacked by level 20+ mobs in wetlands.
All we have to do is find a bunch of videos of people getting attacked by mobs more than 12 levels higher than themself... the problem could be that it would be hard to be sure of a 100% chance since these people would not be able to take enough hits for us to set up any kind of statistic of any worth.

The question again; does it really matter if we allow a 100% crushing blow or hardcap it at some high value like 80 or 90? It's going to be the same outcome for the player; he is going to die or have to flee, engaging a mob that much higher level than himself. One crushing blow more or less I doubt will make much of a difference given it's only 50% extra damage; less than a critical hit anyways.

Phatcat:
@Zooke I would say good luck finding videos of low level chars surviving a hit from a world boss.

sidsukana:
http://vanilla-wow.wikia.com not a proof source
For example 10 May 2010 (3.x) http://wow.gamepedia.com/index.php?title=Crushing_blow&oldid=2262119
**Information since of 3.0 patch an above**

February 2009 http://wow.gamepedia.com/index.php?title=Crushing_blow&oldid=1772881
Before 3.0, the chance of being hit by a crushing blow was 2% per point of difference minus 15%. Giving a 15% chance of being hit by a "boss" (which is 2% * 15 - 15%).

And any other article is excerpted with patches 3.x for example http://vanilla-wow.wikia.com/wiki/Seal_of_Righteousness 15% base mana?)))) AP scaling in classic?? etc

cyberium:
here is my proposal. The advantage of that code make it possible to be same at all expension.
Well i dont know why wotlk had already that check and why it is set as 4 level not 3...

```cpp
    // mobs can score crushing blows if they're 3 or more levels above victim
    if (GetLevelForTarget(pVictim) >= pVictim->GetLevelForTarget(this) + 3
        // can be from by creature (if can)
        && ((GetTypeId() == TYPEID_UNIT && !static_cast<Creature const*>(this)->IsPet() && !(static_cast<Creature const*>(this)->GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_CRUSH))
            || (GetTypeId() == TYPEID_PLAYER && GetCharmerOrOwnerGuid())) // controlled player that considered as creature
        // only autoattack can be crushing blow
        && !SpellCasted)
```

Zooke:
> Well i dont know why wotlk had already that check and why it is set as 4 level not 3

http://wowthinktank.blogspot.co.uk/2008/07/general-its-official-no-more-crushing.html

Phatcat:
@cyberium Your code does exactly the same as my code, you just put it in the beginning of the if statement and i put it last in the if statement. What other difference is there?

cyberium:
@Phatcat its not exactly the same, that code is from wotlk. The use of GetLevelForTarget implie some difference, and they also added the check for controlled players.
But basicaly the code is realy near.

Phatcat:
@cyberium while I agree doing a GetLevelForTarget may be wise to do on the target, doing it on itself seems VERY weird to not just do a simple getLevel. that one you will have to explain to me.

Now, my argument for putting it last in the if statement and not in the beginning is that we should not even do any level checks if it's not a unit (which should be faster and less consuming to find out than to compare the levels, wouldn't you agree?)

cyberium:
@Phatcat GetLevelForTarget will "compute" correct level for bosses. Iam not exactly sure if it matter or not in this case. I think it should be correct.
For the check position i think you are right.

Phatcat:
@cyberium Wouldn't the smart thing to do go ahead and make the same change for the other cores?
Or, in this case, just for vanilla and TBC? - Well, given that this is the unit, and not the player, it will compute the correct level... as it is the boss calling the function getLevel. Is the players level hidden from the boss? O.o - also, why on earth would we be restricted in the info gotten by getLevel when it's called directly on the unit we want the level from? It would very well seem like getlevelfortarget is used to get the 'unknown' if target is much lower level. I think you are confusing the two.

cyberium:
I do confuse nothing :) Has i said i proposed this change from the wotlk code. I never stated that the code is correct or not. My point was just to make them sync. So i dont know the exact role of GetLevelForTarget() here.
Well after more research i think this can be used only in one case. Pet will see all worldboss as they own level +3 (3 by default is set in mangosd.conf).
This seem strange/not correct to me. So yes i think getlevel is enough in this case. Maybe some DB devs @cala @Tobschinski  @Grz3s can explain that? Whats the level of the worldboss?

cala:
Worldboss is a rank given to a creature to make it appears as special for players ingame, like elite, rare or rare elite. There is no specific rules for level for those. But usually, what are called "worldbosses" are the raid bosses encounters where the creatures are designed to have:
Boss_level = Expansion_max_level + 3
This gives Boss_level = 63 in Classic, 73 in TBC, 83 in WotLK and so on... But you can find some NPC marked as worldbosses that do not have the max level.

> Well after more research i think this can be used only in one case. Pet will see all worldboss as they own level +3 (3 by default is set in mangosd.conf).

This is very strange and more than probably wrong. Imagine a lvl 80 hunter doing one of the worldboss from Classic. The player would face a NPC with a level of 63 while his/her pet would see the same NPC as being lvl 83? This makes no sense.

Phatcat:
@cyberium
I must have misunderstood when you said
> GetLevelForTarget will "compute" correct level for bosses

;)

Phatcat:
Is there anything you guys would like for me to change about the code then? Don't like the subtraction, want me to do an addition instead?

cyberium:
its ok for me what, only, what about controled player that should be considered like a creature?
Is that not possible in classic? I mean a controled player crushing blow.
I know that will be another PR but my point is to have the more near possible code for the 3 cores.

Phatcat:
@cyberium I forgot about that. Well, I don't know actually. Should a controlled player be able to do crushing blows? Should they even be able to do that in wrath?

cyberium:
I remember some weird hit from friends in "Skeram prophet" combat AQ40. But not sure they were from crushing blow.

Phatcat:
There's also the mind controlling boss in strath ud, but I really don't remember getting hit by crushing blows, I mostly did those at level 60 with full defense anyways...

Phatcat:
@cyberium Did you not have 300 defense during aq40 raiding back in vanilla? Damn, you guys must have had some heavy carriers.

cyberium:
I had 300 or more defence as a paly in a high level raid with an high level guid ;)

Phatcat:
Then none of your teammates could ever do a crushing blow on you, ever, since you need at least 15 more base skills, and going by the game designers word, also be 3 levels higher.

cyberium:
You probably right on that and i cant get more input about that line in WoTLK code so i more think its not right and have to be reverted.

Phatcat:
@cyberium should we maybe get a second opinion in and vouch for this before pushing and PR'ing the same change for the other cores?

cyberium:
We always can choose to wait :) We often do that in fact, but then some solution are still sleeping from long time in issue tracker or in PR list without decision :/

Phatcat:
I rather have someone complain about this later than noone ever sticking their head in and vouching for this change.

@cala can we just get you to vouch? that'd be enough for me xD

cala:
I'm busy atm. Let me a few hours to read the whole thread, if that's ok for you. :)

Phatcat:
> We often do that in fact, but then some solution are still sleeping from long time in issue tracker or in PR list without decision :/

We need more manpower.... =/

cala:
I'm ok with the change for Classic and TBC: NPCs with level + 3 >= player level can do crushing blows.
For WotLK the level difference has to be changed to 4 instead of 3: http://wow.gamepedia.com/Crushing_blow 

> its ok for me what, only, what about controled player that should be considered like a creature?
Is that not possible in classic? I mean a controled player crushing blow.

Yes, it is possible but very unlikely to occur due to the required level difference.

> I remember some weird hit from friends in "Skeram prophet" combat AQ40. But not sure they were from crushing blow.

Usually, mind-controlled players had a boost to their damage output in order to make them more threatful in raid condition. I'm thinking indeed of [Skerram](http://wow.gamepedia.com/The_Prophet_Skeram) or [Kel'Thuzad phase 2](http://wow.gamepedia.com/Kel%27Thuzad_%28original%29). That is with 100% certainty the reason of that damage burst @cyberium  is refering to because everyone had to be lvl 60 to run AQ40 in Classic so crushing blows were not possible (unless he was speaking of raids occurring in TBC or later with a group where players had different level though greater than 60).

So, that's an OK for me to merge it and change the level difference to 4 for WotLK. :+1:

Phatcat:
Level difference is already 4 in wrath core;
>Well i dont know why wotlk had already that check and why it is set as 4 level not 3...

 but the condition is not placed the same spot in the if statement as my condition (my reasoning for placement is performance, cyber seems to agree)

Phatcat:
@cyberium @cala taking another looksee at this function revealed a massive simplification now that we do a level check, what do you guys think?

cyberium:
I cannot see where it may fail. So i think its good choice.

Phatcat:
Well.... This PR grew just slightly.....

Zooke:
Phatcat, I cannot thank you enough for the way you have picked up the baton and run with this.

src/game/Unit.cpp
```diff
@@ -2099,69 +2099,91 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
...
+
+    if (from_behind)
+        DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: attack came from behind.");
+
+    if (roll < miss_chance)
     {
-        DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: MISS");
+        DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: MISS %d", miss_chance);
         return MELEE_HIT_MISS;
     }
```
cyberium: Its not good practice to have a space between else and and last bracket of previous if. Rhe comment should be moved too.

src/game/Unit.cpp
```diff
@@ -2099,69 +2099,91 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
 
     // bonus from skills is 0.04%
     int32 skillBonus  = 4 * (attackerWeaponSkill - victimMaxSkillValueForLevel);
-    int32 sum = 0;
-    int32 roll = urand(0, 10000);
-    int32 tmp = miss_chance;
+    int32 roll = urand(0, 100) * 100;
```
cyberium: using this formula will give only random number 100 by 100 are you sure about that change?

cyberium:
Ah a whole revamp. Fine. But this implie more check before merging cause it may have huge impact for combat damage.

src/game/Unit.cpp
Phatcat: Yes. Because 1% = 100. we don't need to roll percentages of 12.34, you know what I mean?

src/game/Unit.cpp
Phatcat: I was actually in doubt about that, since I have seen both practises throughout the code. I will stick to not having spaces between any if and else (and else if) from now on. (will of course also change any prs pending)

src/game/Unit.cpp
Phatcat: No wait, shit..... you are right, of course you can have .xx percentages, will revert that one.

cyberium:
Sorry for the mistake i maid by pushing wrong branch :/

Phatcat:
@Zooke also a 100% crushing is impossible by design since miss, dogdge, block and crit will all be checked before reaching crushing. It will however make normal hits very unlikely (impossible actually with this design) from mobs great many levels higher than their victims. That is unless we introduce a hardcap that is, but is a hardcap for crushing blizzlike is the question? It says 40% is hardcap for glancing, so maybe crushing indeed did have a hardcap as well?

Zooke:
@Phatcat What I meant by 100% was 100% of the autoattack damage, i.e. after the other checks if it hits it is 100% crushing. Personally I think a cap is probably the correct way but I have no proof of this. As I said before, all we need is 1 video with a much higher level mob getting a normal hit and we prove a cap. I'll have to see if 1 can be found.

Phatcat:
> Your chance to miss decreases by 0.04%.
Your chance to score a critical hit increases by 0.04%. 

We were not doing this either, now we are.

Zooke:
Another consideration, because pretty much all of the theorycrafting data is based on end game analysis where 300 Defence is your maximum against crushing blows and is a hard cap, are we applying this to lower levels ? i.e. a lvl 40 tank, 200/200 Defence but +10 from gear, making him 210. Do we cap the lvl 40 at 200 ? I don't think we should otherwise +defence enchants/gear becomes pointless.
Dunno how this is handled at the moment, just throwing it out there.

Phatcat:
Data is to be made on attacks in general; it is only crushing blows that calculate by the base value, all other skills calculate appropriately. 

Look, this is me before I knew any programming; https://github.com/ServerProject/Phoenix_1.12.1_Issue_tracker/issues/2819

These comments may be interesting;
> it doesnt show up in the spell book what is so hard to understand
(...)

>>This chance is further modified by the difference between your current Attack rating and the Defense rating of your target. The game tooltip assumes an even level opponent with a Defense rating equal to [Level * 5]. Normally, your Attack rating is equal to your skill rating with the type of weapon you are wielding. The formula is as follows:

>>Your parsed Parry won't necessarily match your tooltip if you're fighting creatures higher or lower in level to you. See the Defense formula above to understand.

>http://www.wowwiki.com/Parry?oldid=316405
>http://www.wowwiki.com/Critical_strike?oldid=119462

>Spell books never updated you recieve the +crit but the spellbook doesnt show it for + or - weaponskill

@griesakanik - just giving you a shoutout you dumb lazy prick? How's Feenix holding up? - you guys still active? of course you're not.
> (A mob's weapon skill is **assumed** to be its level * 5)

This is why it won't **necessarily** match the tooltip, but in most cases it will.

> A source related to this: http://forums.elitistjerks.com/index.php?showtopic=6547
"To compute your crit chance in the spellbook, it assumes the target has 5*level defense skill. This is why you appear to lose .2 to crit when you level; the tooltip is simply recalculated for a target with 5 more Defense."
It should in fact be considered in your spellbook crit chance.

So are we updating the spell book with chances are not?

Oh yeah, and [this](http://wowwiki.wikia.com/wiki/Parry?oldid=316405) makes me kind of sad:
>WoW, like many other games, uses a combat results table-based combat scheme (where one roll determines outcome of an attack), so percentages are absolute.

Phatcat:
@cyberium I don't know how you feel about merging only the 'crushing' commit? 
Since that does not rework the function but corrects the behaviour of crushing within the current functionality (and closes at least 1 open issue we have on the tracker)

cyberium:
+1

Phatcat:
@cala @cyberium new version up taking into account charm and pet cases.

cyberium:
Look ok except the first 'if' seem now a bit unecessary too hard to read but its not a real issue.
Good work.

Phatcat:
@cyberium the first if statement is the level check and weapon skill vs defense skill check, the very essense of crushing; I was planning to do this last but looking at all the other conditions it actually has to verify I thought that the better thing would perhaps be to place it first despite what I originally said. That first if statement is indeed needed. Sadly all of these checks are needed as far as I'm aware. Actually a bit more checks than we did before.
If you are good with this and noone else sees an issue with it (i'll change it to 4 levels for wrath) then I feel it's ready for merge, then I'll take a look at reworking the RollMeleeoutcomeAgainst in another PR.

src/game/Unit.cpp
```diff
@@ -2213,27 +2213,19 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
...
+    // mobs can score crushing blows if they're 3 or more levels above victim
+    // having defense above your maximum (from items, talents etc.) has no effect
+    // mob's level * 5 - player's current defense skill - add 2% chance per lacking skill point, min. is 15%
+    if ((getLevel() - 3) >= pVictim->getLevel() && !SpellCasted
+        && roll < (tmp = (((attackerMaxSkillValueForLevel - victimMaxSkillValueForLevel) * 200) - 1500)))
+    {
+        uint32 typeId = GetTypeId();
+        if ((typeId == TYPEID_UNIT && !(GetOwnerGuid() && GetOwner()->GetTypeId() == TYPEID_PLAYER)
+            && !(((Creature*)this)->GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_CRUSH))
+            || typeId == TYPEID_PLAYER && GetCharmerGuid() && GetCharmer()->GetTypeId() == TYPEID_UNIT)
```
Phatcat: I need to group all the conditions on this line with a ()

Phatcat:
Manual closing once again, has been merged.

PR #140, opened 2016-06-03
[RFC] Update sql

Phatcat:
involuntary courtesy of @kraminpouds

---

This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/140)


cooler-SAI:
add this guy as dev to repo already :+1:

Phatcat:
@cala this has to do with database handling and stuff, so it's for the db teams to review? - I don't know who to ask to look at this cause I'm not sure what parts of our system it affects the most.

cala:
This does not change the DB structure/content so this is not for DB devs to review.
The question is: why update MySQL ? (and no the other DB engines)
Is this fixing an issue? Is this mandatory for a forthcoming change? Or is it because the target hardware for some private server requires this?

Phatcat:
Those are some good questions. Questions I do not hold the answers to honestly. But I can counter with even more questions!

Is there a reason NOT to update MySQL, though? Will it impact compatability? Will it impact performance? Will it impact stabiliy?

It may just be me, but updating/upgrading software, dependencies and libraries should hopefully not be a bad thing.

VladimirMangos:
Update also can improve compatibility with new compilers/os headers/etc... I not see reason not update to official current versions for used external libs

Phatcat:
But an additional 11k lines for maybe no gain at all could also be a bit redundant..

Phatcat:
Where are we on this? Wait till we update the other DB engines, or..?

cyberium:
we need this to be done. Just check it wont mess any platform.

Phatcat:
probably this version of mysql is already outdated by now ^^

cala:
Probably, but the more up-to-date versions are more likely to raise issues with old plateforms (I'm looking at me). So, let's first update with this version here.
I can check that this version works on my plateform at least (this will be low priority for me, though, sorry).

Phatcat:
this whole PR is very very low priority, don't worry at all, I was ready to close it honestly.

Phatcat:
MySQL updated last month to newer version.

PR #141, opened 2016-06-07
Minor fix. Fixed some issues with Git Bash on Windows

CormanDrex:


Phatcat:
Okay, those are 2 very different, very unrelated commits in the same pr, perhaps pr them individually?

You should move the retaliation fix over to another branch, then remove it from your master branch (which you for best practice should avoid pr'ing from in the future) and then force push master to remote to remove the unrelated commit.

Then you go over to your other branch where you copied the retaliation fix over, and do a new pr from that.

Let me know if you need any help with git, it can be a nightmare getting the hang of at first.

CormanDrex:
Yep. I need some help :)

Phatcat:
Could you possibly hit me up on IRC? It will be easier with a bit of feedback going back and forth.

CormanDrex:
I've found an awesome book - "Pro Git" by Scott Chacon. I began to understand how to use Git. :)

PR #142, opened 2016-06-09
[Warrior][Retaliation] Prevent retaliation strikes while stuned.

CormanDrex:


Phatcat:
Not saying this is wrong, but is this enough? It's a good catch, cause of course warriors are not supposed to be able to hit back when incapacitated. But only stunned? how about dazed, sapped, confused, etc?

Phatcat:
I don't know why I typed dazed, of course not dazed, but there is also sleeping, polymorphed, and probably a few more. All these states should not allow you to hit back.

CormanDrex:
I looked through all patch notes but i found info only about stun.
http://wow.gamepedia.com/Patch_1.7.0
I understand that this is logical about sapped confued and etc but theres not much information about it on the Internet...

Phatcat:
The reason the others aren't mentioned is because they all break when they get attacked; but still this first hit that breaks the incapacitation effect, and should **maybe** not trigger a counter hit, or maybe it should?

The spell sure does have a lot of unknown flags associated with it....

https://gist.github.com/Phatcat/d48c65320ea6936c8a11a60e67201b93

PR #145, opened 2016-06-10
[RFC]ST scripting by kraminpounds

Phatcat:
ST ai work by @kraminpounds

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/145)

### DIFF
--- a/src/scriptdev2/scripts/eastern_kingdoms/sunken_temple/boss_atalai_defender.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/sunken_temple/boss_atalai_defender.cpp
@@ -1,0 +1,251 @@
+/* Copyright (C) 2006 - 2011 ScriptDev2 <https://scriptdev2.svn.sourceforge.net/>
+* This program is free software; you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation; either version 2 of the License, or
+* (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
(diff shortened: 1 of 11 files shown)


src/scriptdev2/scripts/eastern_kingdoms/sunken_temple/boss_atalai_defender.cpp
```diff
@@ -0,0 +1,251 @@
...
+	// Zul'Lor
+	SPELL_CLEAVE = 15496,
+	SPELL_FRAILTY = 12530
+};
+
+struct boss_atalai_defenderAI : public ScriptedAI
+{
+	boss_atalai_defenderAI(Creature* pCreature) : ScriptedAI(pCreature)
+	{
+		m_pInstance = (instance_sunken_temple*)pCreature->GetInstanceData();
```
Konctantin: Mabe can replace tabs to spaces?

Phatcat: If the code is good and we decide to do something with it I'll fix it up real fine ;)

cala:
Almost everything in this commit is about moving SD2 code to another SD2 file or moving ACID code to SD2. In overhaul, this is a development choice that will/may work but is different from CMaNGOS development philosophy. Moreover, that brings nothing new in (this is the same content reordered) except for one spot regarding this issue https://github.com/cmangos/issues/issues/469.

I tried to fix it long ago with ACID to avoid using SD2. With no success, so I moved to something else. But if I use SD2 in the end, this will be something different from this.

I'm not saying this PR is wrong (and thank you for spotting it and sharing it) but there is nothing in to take for us in my opinion. So, I suggest to close it. :smile:

xfurry:
Rejected.
We don't need to completely reword this instance because of a few minor things.
As I mentioned yesterday and as @cala pointed out, we can focus only on the broken parts, without rewriting the entire instance.
Over 90% of the script is working in SD2 and Acid already and it has been like that for about 5 years.

Phatcat:
@xfurry You don't close my PRs, I close them myself, thank you.

PR #148, opened 2016-06-11
[WIP]Anti cheat

Phatcat:
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/148)


killerwife:
I think this is one of the things that should be optional and be a module from the get go. (instead of being in the main codebase)

Phatcat:
Why not have it added and just disabled?
I'm not against having it as a module, like we have bots, but I think it's a shame we don't just have it in the core (bots included) and disabled.

xfurry:
In my opinion, I think we should have some patience with this.
We already added quite many big changes to the core. Let's give people some time to stabilize things, before we implement other major changes.

killerwife:
Warden requires a lot of work and once its in the codebase people might think that we fully support it.

Phatcat:
@xfurry - Don't worry, I'm not asking to implement this anytime soon... I got to be better with tagging, sorry.

xfurry:
I didn't ask you to close it... 
I was just saying that this won't be implemented very soon.

Phatcat:
I was informed that someone else is actually in active development of an anti cheat system for cmangos, so I linked him to the branch I pr'd from and another branch I have with a broken half-imported version of warden.

cala:
> We already added quite many big changes to the core. Let's give people some time to stabilize things, before we implement other major changes.

I agree with @xfurry : there were many heavy changes in core of late and there are still issues opened on a weekly basis about them. So in my opinion, we should wait a bit about next big changes and take that time to deal with the numerous "small" issues already open. Also, I'm not sure there was ever a warden for Classic.

Moreover, warden is relevant for big (private) servers with many people (that are not supported by CMaNGOS for obvious reasons), not for the small home server with one or two dozen of friends (that should still not be supported, but do not fool ourselves here). So, IMHO, warden is not needed if we are straight.

Phatcat:
@cala 
I really see no reason not to have warden..... aaaaaaaand.... there's nothing illegal about private servers, small or big. I've been over this countless times with people on countless forums. If you need the proof I will provide (as in court order decisions - the same decisions allowing us to emulate in the first place)

killerwife:
I can only vouch for it being a major time-sink to even make it run properly.

PR #150, opened 2016-06-11
Player bots work

Phatcat:
From Nost... I can't help it. PR'd to Portalclassic as well.

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/150)


cala:
As long as playerbots are kept separated from core, I think this kind of PR should also be kept separated (moreover since you've made one to playerbot repo ;-) )

Phatcat:
This was to spark the debate about merging playerbots into the core, since it looks like that was what Nost had planned and I don't think that may be a bad idea.

Phatcat:
@cala but the PR on playerbots should be enough to remind us about this when it becomes relevant, so I'll close this PR and just let the branch go stale for now.

PR #151, opened 2016-06-12
Pet patch 1.1: The tiny Strange one

Phatcat:
Okay, this was the stuff I could find related to pets on some of all the different repos I've been looking over. Now properly looking it over, this is what we have use of.

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/151)


Phatcat:
There wasn't much fun in the commits I removed from the PR again; some of them were things that fitted in other prs, some were things we already solved by other means and some were... hackfixes.

These two commits I've pulled in should however solve 2 open issues we have;
https://github.com/cmangos/issues/issues/551 as well as something which is unreported; that owners don't enter combat when their pets do. This is heavily abusable.

My own commit is just a friendly request for someone to be so kind and tell me how to send a message that shows up on the screen for the user so I wouldn't have to update the TODO but actually just do it..

cala:
The commit making the owner enter combat if its pets does should be enhanced to work both ways: pet should also enter combat with current owner's target if owner is in combat. Currently, this is not the case and can be tested with a NPC that summon pets but is neutral to players. Attacking the NPC will make it aggro, but the summoned pets will never attack the player unless directly engaged.
NPC 2755 is a good example for that.

cala:
Looks promising. :+1: 
I'm very busy at the moment. I'll test it when I get the chance but it won't be before next week, sorry.

Phatcat:
@xfurry - abit like this? https://github.com/cmangos/mangos-classic/pull/151/commits/91da352770b12755e10ec540cb86783d60ceac43
Or were you thinking something else?

cala:
I expect to be able to test tonight or more probably during the two following days. Any specific tests/things you want me to perform or that I should focus on?

Phatcat:
It would be nice to know if we at least close the 2 issues I've linked with this.

cala:
As promised, I tested this PR and it looks good.

I had no crash taming rare creatures, but I never had one the few times I did it before this patch so I can't confirm it's 100% fixed.

A remark, as we discussed elsewhere: https://github.com/cmangos/mangos-classic/pull/151/commits/e1076972127dd75233ba04fa5565f2e6f9231d23 should be expand to make also pets enter combat with their master aggro list *on spawn*. Without this, pets will not assist their master if the master's target is at range (outside aggro range) or if the master has a neutral faction.

An example of the later can be checked by applying (script for Classic only):
```SQL
DELETE FROM `creature_ai_scripts` WHERE `id` BETWEEN 275502 AND 275503;
INSERT INTO `creature_ai_scripts` (`id`, `creature_id`, `event_type`, `event_inverse_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action1_type`, `action1_param1`, `action1_param2`, `action1_param3`, `action2_type`, `action2_param1`, `action2_param2`, `action2_param3`, `action3_type`, `action3_param1`, `action3_param2`, `action3_param3`, `comment`) VALUES
(275502, 2755, 0, 0, 100, 1, 20000, 20000, 6000000, 6000000, 11, 4937, 0, 1, 22, 2, 0, 0, 0, 0, 0, 0, 'Myzrael - Cast Summon Prismatic Exiles'),
(275503, 2755, 0, 0, 75, 1, 26500, 27500, 6500, 7500, 11, 10388, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Myzrael - Cast Summon Prismatic Exile');
```
and then attacking Myzrael (`.add npc 2755`): her adds stand still instead of attacking the player. However, attacking one of the adds make it (and only it) react and fight back.

I did not find something wrong with the other changes.

Phatcat:
The crash has been tested and confirmed on mangos zero here: https://github.com/mangoszero/server/commit/f6731a4f9160027917617c34796d8b034db8e087#commitcomment-16760506 and we have the issue reported here; cmangos/issues#551

To reproduce it you need to try and tame specific rares, those that are pooled.

I would be surprised would this turn out to be different issues needin' different fixes.

> A remark, as we discussed elsewhere: e107697 should be expand to make also pets enter combat with their master aggro list on spawn. Without this, pets will not assist their master if the master's target is at range (outside aggro range) or if the master has a neutral faction.

I see what you mean now. I misunderstood you.... Once again..

The question is, though, to try and fix this during the different summon effects in spelleffects, or locate which AIs it is relevant for and work it into that/those... the latter would seem like the more logical, but also heavier approach...

cala:
> To reproduce it you need to try and tame specific rares, those that are pooled.

Ok, I'll try again with those and report back.

> The question is, though, to try and fix this during the different summon effects in spelleffects, or locate which AIs it is relevant for and work it into that/those... the latter would seem like the more logical, but also heavier approach.

I can't tell, sorry. I would say the former, but that's only because I think that not all summons should behave that way, only the guardians and minions. @xfurry could be of good advices here to avoid any conflict with the other AI, especially SD2 scripts where there are a lot of `Attack()` when a NPC is summoned.

cala:
Ok, I can confirm that:
- I had a crash when successfully taming a rare NPC that is pooled (I guess the issue was more related with the fact of being pool than being rare) with master branch
- I was not able to reproduce that crash with your PR (I only did a dozen of tames, but the 66% chances advertised in one of the report should have poped-in at least once)

Phatcat:
Want me to reword the commit to better suit what it actually fixes and what the issue actually was or you think that one is fine as is?

cala:
Oh no need to do that. I did not have a look at the commit before my initial comment. I think it is fine as it is.

Phatcat:
I'm gonna do it in the spell effects as I think you're right that maybe not all summons should be aggroed on summon. However! you do mention this particular gem;

> especially SD2 scripts where there are a lot of Attack() when a NPC is summoned.

So if we do this correct in the summon effects, all of those calls would be redundant (since they are happening as spell effects, right?) and can be cleaned from, thereby 'simplifying' "a lot" of scripts? :smile:

cala:
> So if we do this correct in the summon effects, all of those calls would be redundant (since the 'NPC summoning's are happening as spell effects, right?

If only that was so easy... :wink: 
Some summoned are done by spells, some are not. Some are done by spells on a core and not on another core (spell was removed) or different spells are used on different cores. I know this does not look perfect or at least clean, but Blizz did not always do perfect things and moreover they don't seem to have backward compatibility in mind when developing a new expansion. So we have to deal with that.

Said otherwise, the use of `Attack()` is handle case by case in the scripts. So yes, adding the attack in the summon effects will probably allow to clean some scripts, but not a lot, I'm afraid.

Phatcat:
>Said otherwise, the use of Attack() is handle case by case in the scripts. So yes, adding the attack in the summon effects will probably allow to clean some scripts, but not a lot, I'm afraid.

Doesn't that depend which effects they are using, and fix the appropriate effects in each and every core, because otherwise we may be leaving out pets that are not summoned via script bt by acid?

src/game/SpellEffects.cpp
```diff
@@ -3006,12 +3016,21 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
         ((Player*)m_caster)->PetSpellInitialize();
     }
     else
-    {
-        // Notify Summoner
+    {   // Notify Summoner and assist if in combat
         if (m_originalCaster && m_originalCaster != m_caster && m_originalCaster->GetTypeId() == TYPEID_UNIT && ((Creature*)m_originalCaster)->AI())
```
Phatcat: Hmmm, that m_originalCaster->GetTypeId() == TYPEID_UNIT seems a bit redundant here since this wholse else shabang will only execute if m_caster->GetTypeId() != TYPEID_PLAYER

src/game/SpellEffects.cpp
```diff
@@ -3006,12 +3016,21 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
...
+        {
             ((Creature*)m_originalCaster)->AI()->JustSummoned(NewSummon);
+
+            if (m_originalCaster->isInCombat())
+                ((Creature*)NewSummon)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
+        }
         else if (m_caster->GetTypeId() == TYPEID_UNIT && ((Creature*)m_caster)->AI())
```
Phatcat: And again here..

src/game/SpellEffects.cpp
```diff
@@ -2654,11 +2654,21 @@ void Spell::EffectSummonGuardian(SpellEffectIndex eff_idx)
...
+
+            if (m_originalCaster->isInCombat())
+                ((Creature*)spawnCreature)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
+        }
         else if (m_caster->GetTypeId() == TYPEID_UNIT && ((Creature*)m_caster)->AI())
+        {
             ((Creature*)m_caster)->AI()->JustSummoned(spawnCreature);
+
+            if (m_originalCaster->isInCombat())
+                ((Creature*)spawnCreature)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
```
Phatcat: Oh, that is wrong...

src/game/SpellEffects.cpp
```diff
@@ -3006,12 +3016,21 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
...
+
+            if (m_originalCaster->isInCombat())
+                ((Creature*)NewSummon)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
+        }
         else if (m_caster->GetTypeId() == TYPEID_UNIT && ((Creature*)m_caster)->AI())
+        {
             ((Creature*)m_caster)->AI()->JustSummoned(NewSummon);
+
+            if (m_originalCaster->isInCombat())
+                ((Creature*)NewSummon)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
```
Phatcat: And here. Nice.

killerwife:
I did some checking, and only found jewelcrafting statues example. Afaik they are guardians and they should never attack. While I understand this being a classic PR, isnt there such an example in vanilla as well? Just checking that the "attack" on summon wont break anything.

Phatcat:
@killerwife in fact, there are many. One would only have to turn to the engineering profession to find that all target dummies are guardians, alarm-o-bot is a guardian and there are a few others.

Guardians are sadly not in a pretty state atm and this may indeed not make things better for them... Any suggestion is appreciated. I'm open to just about anything besides hackfixing.

killerwife:
well then wont this
if (m_originalCaster->isInCombat())
                 ((Creature*)spawnCreature)->AI()->AttackStart(m_originalCaster->getAttackerForHelper());
break these non-attack guardians?

Phatcat:
These non-attack guardians are already broken, probably have been for long before I joined here, in the sense that all guardians behave in the same generic way, unless they are scripted, so I guess it would be a matter of scripting the missing behaviours, and in that case we could overwrite any attack behaviour in the scripts, no?

killerwife:
Yes, but it would be very unintuitive. We already have to hack creatures so that they dont go into combat. Therefore I think there should be a setting for it, or check against some flags, because this will force even a PASSIVE guardian to attack.

Phatcat:
@killerwife you are not really listening, passive guardians ALREADY DO ATTACK.

Phatcat:
Furthermore, there is no such thing as a 'passive' guardian in our system as we overwrite the reaction to aggressive. Something we have been doing for, as I said, longer than I've been here.
So unless it's scripted then it's broken; and if we have to script most 'special' guardians to make them work then whatever 'special' guardian checking you are asking for here will be redundant, am I wrong?

Phatcat:
I just remembered something about this commit as well @killerwife - it only affects creature summons, because players don't have an AI, and this is only called if the caster (owner) has an AI - so we will not have any problems with any engineering or jewelcrafting pets, or any other player pets for that matter; the question is then, are there passive creature pets out there at all? cause if not then issue solved :tada:

Phatcat:
@killerwife Been studying the TBC core specifically for a bit now and it looks the jewelcrafting statues are handled as totems for the summon, so they aren't even a problem at all actually.

I don't think there's any regressions in this, HOWEVER you are very right that we need to utilize flags for special pets since many of them are not working as they should at all, regardless of any of my work.

Phatcat:
@cala I added the check it to the relevant spellEffects, what do you think? - unless it has any apparent regressions I am thinking of merging this PR together with satys orientation commit.
@killerwife would also like your thoughts on this including any changes I might have done since you first looked it over.

cala:
I hope to have a look at this tonight or tomorrow. :)

cala:
The last commit does indeed allow "neutral" pets to assist their summoner on spawn instead of staring passively at their beloved master behind kicked down.
It does not seem to break SD2 scripts that already use "Attack()" for the gardian pet on summon.
This is nice. :+1: 

However, when the fight with the "neutral" master is reset (vanish, GM ON, death...), upon attacking again the summoner, the neutral pets do not engage again in fight to assist their master.

Phatcat:
I think it's because the checks me and cayden do in unit can't check for guardians, I think it only checks for 'pets', I don't know if you know what I mean, or if it's true at all... At least it's no regression here as it's been this way at least since pet patch 1 was merged..

Phatcat:
@cala @killerwife

Reworked most of my work in this PR - this feels a lot more right. It also solves a few issues.

Phatcat:
@cyberium @VladimirMangos What do you guys think of this?

cala:
@Phatcat Nice, I'll test it tonight or this WE. :)

Phatcat:
How lovely. I just did some hours testing and I had to reintroduce one of the commits from before the 'rework' of the PR. Looking at the weird behaviour on owner death as well as weird 'neutral' mob behaviour currently.

You should test it with some of the pets (most notably those with flag changes) that I PR'd for the DB, I really like how the target dummies turned out, too bad their loot is messed, though.

cala:
After a few tests, I see nothing wrong with this PR. I can't say that I did all possible tests on it, since I'm the only one testing your PR (aside from you, of course ;) ), I can't be sure. But as soon as you push it to master, you can expect feedback within minutes if something is even slightly wrong...

There are still three - minor- points that would need to be fixed soon or later (if you feel like doing it, but I can perfectly understand that you grow tired of working on pets):
- the minion-like pets do not use their creature type : beast pets should be sheepable/hibernatable, UD should be shackle and so on
- guardian pets summoned by players are stuck at spawn if they have EventAI set as AI (at least the one I used: Timbermaw Ancestor : NPC 15720, summoned by item 21326, exemple of AI script below) while the guardians used by NPCs work like a charm with or withouth EventAI script
- neutral guardians summoned by (neutral) NPC do attack their owner's target on summon if owner is in combat (this is fine), however if combat is reset (death, GM ON/OFF) the guardians won't enter combat again if the owner is attacked, player have to manually attack each guardian to make them enter combat

Overhaul, good job @Phatcat ! :+1: 

As promised, example of AI script I sued for the Timbermaw Ancestor:
```SQL
DELETE FROM `creature_ai_scripts` WHERE `creature_id` = 15720;
INSERT INTO `creature_ai_scripts` VALUES
(1572001, 15720, 14, 0, 100, 1, 1000, 40, 15000, 18000, 11, 11986, 6, 1, 0, 0, 0, 0, 0, 0, 0, 0, 'Timbermaw Ancestor - Cast Healing Wave on Friendlies'),
(1572002, 15720, 0, 0, 100, 1, 0, 40, 3400, 4800, 11, 9532, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Timbermaw Ancestor - Cast Lightning Bolt');

UPDATE creature_template SET AIName='EventAI' WHERE Entry=15720;
```

Phatcat:
>  But as soon as you push it to master, you can expect feedback within minutes if something is even slightly wrong...

Oh, I am very aware of that ^^

> the minion-like pets do not use their creature type : beast pets should be sheepable/hibernatable, UD should be shackle and so on

didn't I already look into this? - what are these 'minion' like pets? - Maybe the change wasn't added to all relevant summon effects.

> (if you feel like doing it, but I can perfectly understand that you grow tired of working on pets)

I am done with pets when pets are done. Simple as that.

cala:
Sorry, my mistake: the creature type is correctly handled. I did not reset my WDB before testing like I should have. Everything's fine! :+1:

Phatcat:
It looks like this is ready for merge then... think I'm gonna sneak it into the core when people are asleep ^^

cala:
Or drunk. It looks like Germans are more likely to get drunk than asleep tonight. ;)

Phatcat:
Merged.

PR #154, opened 2016-06-14
Fix display of pet mode and reaction after dismount

Oniryck:
I think this can come in handy after recent pet work if the code works for you. (before this, after unmounting, the pet mode/reaction would always be follow/defensive even if you set your pet to anything else before mount)

### DIFF
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -1953,12 +1953,9 @@ void Pet::SynchronizeLevelWithOwner()
     }
 }
 
-void Pet::ApplyModeFlags(PetModeFlags mode, bool apply)
+void Pet::SetModeFlags(PetModeFlags mode)
 {
-    if (apply)
-        m_petModeFlags = PetModeFlags(m_petModeFlags | mode);
-    else
-        m_petModeFlags = PetModeFlags(m_petModeFlags & ~mode);
+    m_petModeFlags = mode;
 
     Unit* owner = GetOwner();
     if (!owner || owner->GetTypeId() != TYPEID_PLAYER)
--- a/src/game/Pet.h
+++ b/src/game/Pet.h
@@ -49,12 +49,17 @@ enum PetSaveMode
 // There might be a lot more
 enum PetModeFlags
 {
-    PET_MODE_UNKNOWN_0         = 0x0000001,
-    PET_MODE_UNKNOWN_2         = 0x0000100,
+    PET_MODE_STAY              = 0x0000000,
+    PET_MODE_FOLLOW            = 0x0000001,
+    PET_MODE_ATTACK            = 0x0000002,
+    PET_MODE_PASSIVE           = 0x0000000,
+    PET_MODE_DEFENSIVE         = 0x0000100,
+    PET_MODE_AGGRESSIVE        = 0x0000200,
+
     PET_MODE_DISABLE_ACTIONS   = 0x8000000,
 
     // autoset in client at summon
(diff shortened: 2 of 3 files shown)


src/game/Pet.h
```diff
@@ -49,12 +49,17 @@ enum PetSaveMode
 // There might be a lot more
 enum PetModeFlags
 {
-    PET_MODE_UNKNOWN_0         = 0x0000001,
-    PET_MODE_UNKNOWN_2         = 0x0000100,
+    PET_MODE_STAY              = 0x0000000,
+    PET_MODE_FOLLOW            = 0x0000001,
+    PET_MODE_ATTACK            = 0x0000002,
+    PET_MODE_PASSIVE           = 0x0000000,
```
Phatcat: STAY and PASSIVE are not the same, they should probably not have the same OP-code.

Oniryck: This OP is never used and could be commented out. I left it to pinpoint this logic behind the modes.

Phatcat:
I just tested this and it seems to acomplish what we want.

Furthermore we even get a function call simplified which is never a bad thing.

I don't see any regressions with this one, yet I'm still very new here so I'd like for one of the older devs to give their opinion on this. @cyberium would probably not be a terrible guess for a proper code reviewer for this but if he doesn't have the time then @xfurry or @VladimirMangos would not be terrible guesses either.

Phatcat:
As a sidenote, we have this exact same visual error when summoning as well; nomatter the state of the pet before, it will **display** itself as defensive and following, even if it in fact has other pet states.

src/game/Pet.cpp
```diff
@@ -1953,12 +1953,9 @@ void Pet::SynchronizeLevelWithOwner()
     }
 }
 
-void Pet::ApplyModeFlags(PetModeFlags mode, bool apply)
+void Pet::ApplyModeFlags(PetModeFlags mode)
```
VladimirMangos: Please rename to Set then

src/game/Pet.h
Phatcat: Nah, I get it now.

0x0000100
Follow + passive

0x0000000
stay + passive

These op-codes are probably correct, yea.

Phatcat:
Would you mind squashing the commits? Don't think the changes need each their own commit.

Oniryck:
Sure, here you go.

Phatcat:
LGTM

Phatcat:
Going to merge this when the PR reaches it's 14 day mark unless someone else has any objections before then.

src/game/Unit.cpp
```diff
@@ -6257,7 +6257,10 @@ void Unit::Unmount(bool from_aura)
         SendMessageToSet(&data, true);
 
         if (Pet* pet = GetPet())
-            pet->ApplyModeFlags(PET_MODE_DISABLE_ACTIONS, false);
+        {
```
Phatcat: There's actually no need for this line

src/game/Unit.cpp
```diff
@@ -6257,7 +6257,10 @@ void Unit::Unmount(bool from_aura)
...
+            if (CharmInfo* charmInfo = pet->GetCharmInfo())
+                pet->SetModeFlags(PetModeFlags(charmInfo->GetReactState() | charmInfo->GetCommandState() * 0x100));
+        }
```
Phatcat: if you remove this line as well

Phatcat:
Gonne give people until tonight to tell me why this shouldn't be slapped onto all 3 cores.

cala:
ok for me.

Phatcat:
Merged.

PR #156, opened 2016-06-15
Suppress default greeting text for quest dialog when interaction wit…

lduguid:
Suppress default greeting text for quest dialog when interaction with quest giver game object is not appropriate. 

See [WANTED Poster Greets You #142](https://github.com/classicdb/database/issues/142#issuecomment-226333443) and [Gadgetzan #860](https://github.com/classicdb/database/issues/860) for more history and details motivating this pull request.

### DIFF
--- a/src/game/SpellHandler.cpp
+++ b/src/game/SpellHandler.cpp
@@ -285,6 +285,10 @@ void WorldSession::HandleGameObjectUseOpcode(WorldPacket& recv_data)
         return;
     }
 
+    // Supress default greeting text for quest dialog when interaction with quest giver game object is not appropiate.
+    if (obj->GetGoType() == GAMEOBJECT_TYPE_QUESTGIVER && !obj->ActivateToQuest(_player) && !obj->GetGOInfo()->GetGossipMenuId())
+        return;
+
     obj->Use(_player);
 }
 


lduguid:
I am becoming more convinced that this is **not** the correct place to do this (HandleGameObjectUseOpcode in SpellHandler.cpp), 

I have since discovered other quest game objects that will also show the default  'Greetings <playername>' message when you are not on the appropriate part of that quest that can **not** be handled here. For example there is the 'Old Lion Statue'  (http://db.vanillagaming.org/?object=31), this quest game object is handled differently via SpellEffects.cpp, Spell::EffectOpenLock(...)-> Spell::SendLoot(...). When the player interacts with this GO, you are given a timer/progress bar (like opening a chest) and then once the progress is up the greetings message is displayed.

Following my current 'tact', I could also place similar logic as above in the appropriate location of either _EffectOpenLock_ or _SendLoot_ but that is even 'hackier' with the feel of putting out fires just before they occur rather than generic handling it in one place upfront (is this even possible?).

I also revisited @Cala original investigation at [WANTED Poster Greets You #142](https://github.com/classicdb/database/issues/142#issuecomment-226333443) and placed this logic above inside void GameObject::GameObject::Use(Unit* user). i.e.

```
}
        case GAMEOBJECT_TYPE_QUESTGIVER:                    // 2
        {
            if (user->GetTypeId() != TYPEID_PLAYER)
                return;

            Player* player = (Player*)user;

            if (!sScriptMgr.OnGossipHello(player, this))
            {
                player->PrepareGossipMenu(this, GetGOInfo()->questgiver.gossipID);
                // From here --->
                if (!ActivateToQuest(player) && GetGOInfo()->GetGossipMenuId())
                    return;
                // <--- to here
                player->SendPreparedGossip(this);
            }

            return;
        }
        case GAMEOBJECT_TYPE_CHEST:       

```
This has the advantage of suppressing the greeting message (when appropriate) regardless of how the quest displays it and does away with handling is different locations, however it does feel wrong to also be just suppressing the greeting message at the point of displaying it. In the case of the 'Old Lion Statue' with the progress bar, the user will wait for the progress to end, only to see nothing, that feels wrong.

I believe what it really required is suppressing interaction (cannot click it at all) with the quest object altogether, like you do with say quest GO's that you have to collect. i.e. you cannot interact with them in anyway unless you are on the collection quest. I have zero experience with  cmangos core and am not sure if this is possible with quest giver game object types?  It is too bad I could not get the suggestion to use GO_DYNFLAG_LO_NO_INTERACT in GameObject::BuildValuesUpdate to work as it seems like the right place to handle this...

I am under the gun to finish a work related project before the weekend, so I will not be able to revisit it until then, feel free to reject this pull request in the meantime. I will leave it open for now in case it jogs someone with more experience on perhaps a better approach.

Phatcat:
@lduguid Thank you for your efforts, your experimentation and detailed explaination of the situation as it is currently.

I think you are right that this is not the solution we are looking for,also in part of the limitations you report and the overall wrong approach to the issue.

You may be on to something, as my initial thought as well was that these objects should not be highlightable by the mouse at all if not on the quest, just as with collectiable quest object like you mention. It is not about them greeting you, it is about you being able to click them in the first place that is the issue.

Zooke:
Do we have any source for not highlighting, or the greetings messages (didn't some objects even have unique texts for when you were not on the quest ?)

Phatcat:
No, you should be able to highlight and get the tooltip showing the name of the object when mousing over it ingame, but you should not be able to interact with the object.

Gerhood:
Interaction with a Questgiver-Object should be possible if (and only if) the player is on a quest for that object (the object will display quest texts) OR the object has a default gossip assigned.
This is how it works on official servers, I cannot confirm for Vanilla WoW.

Phatcat:
So our problem may in fact be with us having default greetings set for a bunch of items that shouldn't have?

Gerhood:
The default greeting is showed if none is set in DB. If the object cannot even be interacted with this is no problem.

cala:
Sorry, I'm still very busy.
I confirm what Gerhood says: this issue comes from the fact that players car interact with quest GOs at a time where the quest(s) associated with the GOs is(are) not available or in progress. Thus, there is no gossip to create (because no quest content is returned), hence the default gossip returned instead.

To fix this issue, players should be prevented to interact/activate with such GOs when they are not on the quest. Gerhood suggested a fix (maybe two) I did not have the time to test.

lduguid:
Closing this pull request, as further discussion in [WANTED Poster Greets You #142](https://github.com/classicdb/database/issues/142#issuecomment-226333443) has suggested a better approach by @Gerhood . If no on else beats me to it, I will make another pull request with said suggestions later tonight.

lduguid:
Please see [flag (no interact) quest giver GO when player not eligible for that quest #157](https://github.com/cmangos/mangos-classic/pull/157) for new pull request replacing this one.

PR #158, opened 2016-06-23
Crash fixes

Phatcat:
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/158)

### DIFF
--- a/src/game/Object.h
+++ b/src/game/Object.h
@@ -519,7 +519,7 @@ class MANGOS_DLL_SPEC WorldObject : public Object
         float GetDistanceZ(const WorldObject* obj) const;
         bool IsInMap(const WorldObject* obj) const
         {
-            return IsInWorld() && obj->IsInWorld() && (GetMap() == obj->GetMap());
+            return obj && IsInWorld() && obj->IsInWorld() && (GetMap() == obj->GetMap());
         }
         bool IsWithinDist3d(float x, float y, float z, float dist2compare) const;
         bool IsWithinDist2d(float x, float y, float dist2compare) const;
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -2511,6 +2511,10 @@ void Spell::SpellStart(SpellCastTargets const* targets, Aura* triggeredByAura)
     m_spellState = SPELL_STATE_STARTING;
     m_targets = *targets;
 
+    if (m_CastItem)
+        m_CastItemGuid = m_CastItem->GetObjectGuid();
+
+
     m_castPositionX = m_caster->GetPositionX();
     m_castPositionY = m_caster->GetPositionY();
     m_castPositionZ = m_caster->GetPositionZ();
@@ -2930,6 +2934,12 @@ void Spell::update(uint32 difftime)
     UpdatePointers();
 
     if (m_targets.getUnitTargetGuid() && !m_targets.getUnitTarget())
+    {
+        cancel();
+        return;
+    }
+
+    if (m_CastItemGuid && !m_CastItem)
     {
         cancel();
         return;
@@ -3747,6 +3757,7 @@ void Spell::TakeReagents()
                 }
 
                 m_CastItem = nullptr;
+                m_CastItemGuid.Clear();
             }
         }
 
@@ -5899,6 +5910,11 @@ void Spell::UpdatePointers()
     UpdateOriginalCasterPointer();
 
     m_targets.Update(m_caster);
+
+    if (m_caster->GetTypeId() == TYPEID_PLAYER)
+        m_CastItem = ((Player *)m_caster)->GetItemByGuid(m_CastItemGuid);
+    else
+        m_CastItem = nullptr;
 }
 
 bool Spell::CheckTargetCreatureType(Unit* target) const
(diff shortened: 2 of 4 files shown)


cala:
They look ok to me. :+1: 
Just two remarks:
- if possible, the chinese characters in the second commits (I assume they are chinese) should be changed to an english title so everyone can guess at a glance what this commit does without having to look at the code
- in the same commit it would be nice, but not mandatory to add a empty line after the pVictim check to make the whole check more easily spotable (yes, I'm being really picky here)

Aside from these, that's a go for the code.

Phatcat:
@cala I couldn't help but give the function a small shake-up ..

cala:
:+1:

Phatcat:
Alright, unless anyone has any issues with these I'mma merge them to all 3 cores tonight.

src/scriptdev2/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
```diff
@@ -598,8 +598,11 @@ struct boss_arugalAI : public ScriptedAI
         {
             if (!m_bAttacking)
             {
-                DoCastSpellIfCan(m_creature->getVictim(), SPELL_VOID_BOLT);
-                m_uiVoidboltTimer = urand(2900, 4800);
+                if (Unit* victim = m_creature->getVictim())
```
xfurry: This is already checked in the script:

```
// Check if we have a current target
        if (!m_creature->SelectHostileTarget() || !m_creature->getVictim())
            return;
```

I don't think that this change is relevant for the crash.

xfurry: Eventually we can do another redundant check for "pTarget" in this function: 
https://github.com/cmangos/mangos-wotlk/blob/master/src/game/CreatureAI.cpp#L84

So then it will be clear for any spell cast from script.

Phatcat: Alright, thanks.

I'm going to try and look into that one in detail then.

Phatcat:
I gotta learn to do better reporting:
The crash that @mityada s commit fixes can be reproduced by using your hearthstone and throwing it away before the cast is finished.

src/scriptdev2/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
Phatcat: But your comment about adding it to CreatureAI, would that then mean we could remove it from all the different scripts? If so, and not much heavier on the server then it may actually not be a bad idea, what do you think?

Phatcat: @xfurry But on line 564 in this script we do a 'StopAttacking()' - doesn't that nullify the victim? Then we would in fact have to recheck it here.

Shouldn't it be a ((Unit*)this)->AttackStop(); if we indeed want to keep the victim but stop meleeing?

xfurry: StopAttacking() is a custom function in this script.
It shouldn't impact the spell casting.

Actually that function should be called StopMeleeAttack and StartMeleeAttack, because it allows the creature to start or stop movement to the target.

Anyway, that script looks pretty old. I think it was one of the firsts to be developed, so the logic of the event looks kind of weird to me.

However, I still think it shouldn't crash on spell cast.
Usually the target pointer during the CastSpell function is checked multiple times, and this wasn't a problem before.

Phatcat: I have not spend a great deal on that one. It sounds like the whole script is maybe due for a workover making this commit even more redundant, I'm going to pull it from the lot since I don't think this commit is worth defending any further.

Phatcat:
Gonna merge these remaining 2 since they indeed do fix some crashes producable by players, sometime when the checks have passed or tomorrow, depends how long I stay up tonight.

VladimirMangos:
Do you think you fix problem. No. Before you have bug in code + crash. Now you have bug + hack  and possible no crash. I think this way to nowhere...

Phatcat:
@VladimirMangos Actually I am on the school of thought that we should allow for crashing when we have a bug, but @cyberium told me otherwise, he indeed told me to 'hide' a crash until we could solve it properly; read here: https://github.com/cmangos/issues/issues/969#issuecomment-227457768

I'm just trying to be somewhat consistent...

cyberium:
Ok its not same case but i keep same attitude.
I prefer a spell that will not work than a crash.
So iam sure for spell that were caused those crash we will/have new opened issues and then we'll fall back to that invalid pointer.
When you dont have the ability to fix all, fixing a crash is a good enough step.

Phatcat:
> but i keep same attitude.

That's what I mean by trying to be consistent.

And while I think cyberiums is the safer approach, I do agree with vladimir that hiding and 'fixing' bugs like this may lead to us hiding the issue for ourselves making it harder to spot and fix the actual problems, while also greatly limiting ourselves in the bug reports we will recieve since people are more likely to report crashes than anything else.

I am a bit torn here on what approach to follow now to be honest.

PR #119, opened 2016-04-26
Implemented school immunity for creature from database

sidsukana:
There are many creatures that are immune to a particular school of magic (such as bosses) patch makes it possible to establish immunity for school in the database

PR #159, opened 2016-06-24
Fixing a bug, where the hunter pet doesnt look at his target

SatyPardus:
This bug was leading to an pet, that didnt attack it's target, making the pet nearly unuseable and worthless. This will help your little buddy.

### DIFF
--- a/src/game/PetAI.cpp
+++ b/src/game/PetAI.cpp
@@ -335,6 +335,16 @@ void PetAI::UpdateAI(const uint32 diff)
             return;
         }
 
+        if (!m_creature->HasInArc(2 * M_PI_F / 3, victim))
+        {
+            m_creature->SetInFront(victim);
+            if (victim->GetTypeId() == TYPEID_PLAYER)
+                m_creature->SendCreateUpdateToPlayer((Player*)victim);
+
+            if (owner && owner->GetTypeId() == TYPEID_PLAYER)
+                m_creature->SendCreateUpdateToPlayer((Player*)owner);
+        }
+
         // required to be stopped cases
         if (m_creature->IsStopped() && m_creature->IsNonMeleeSpellCasted(false))
         {


SatyPardus:
@Phatcat i hope i got it right with the branch stuff. I messed up everything again and had to do everything again. I hope i didnt destroy my local changes on other forks >.<

src/game/PetAI.cpp
```diff
@@ -335,6 +335,16 @@ void PetAI::UpdateAI(const uint32 diff)
             return;
         }
 
+        if (!m_creature->HasInArc(2 * M_PI_F / 3, victim))
+        {
+	        m_creature->SetInFront(victim);
```
Phatcat: over-indented

only 4 whitespace chars per 'tab'

src/game/PetAI.cpp
```diff
@@ -335,6 +335,16 @@ void PetAI::UpdateAI(const uint32 diff)
...
+	        if (victim->GetTypeId() == TYPEID_PLAYER)
+		        m_creature->SendCreateUpdateToPlayer((Player*)victim);
```
Phatcat: This is really over-indented.

src/game/PetAI.cpp
SatyPardus: My settings are set to 4 spaces per tab o.o Why did it push more inside it?

Phatcat: Maybe you hit the space bar and didn't notice.

It is no biggie, you can do a fixup commit and then squash them and it'll be all good =)

SatyPardus: In my local git and in visual basic it's only 4 tabs. Idk why here are 8

Phatcat: It looks like you may have hit the tab twice and thought that you could delete the whole tab by pressing backspace but in fact you only deleted a single whitespace? That would explain the odd number of 7, and the 'double tab' you do in line 342 and 345 ?

SatyPardus: This is how it looks for me, and no, they are not tabs. I only work with spaces (It's by default 4 spaces per tab press in my settings, no actual tabs are added)
http://imgur.com/fLUVJIe - Did git maybe change something? Is there a setting?

Sorry for the nooby questions^^ Learning this is harder than c++

Phatcat:
overall its looking fine (apart from the 'overuse' of whitespacing ^^)

SatyPardus:
Now it seems to be fixed. Seems like visual studio added one invisible tab before the lines i copied of my other branch. Fixed it by Shift tabbing everything and added the tabs again. Stupid tabs^^

Phatcat:
Okay, are you ready for some of the more complicated stuff?

Now, on your branch, open whatever client you want to use for writing git commands, I personally use powershell. Enter the command:

`git rebase -i head~2`

This will open up a txt document allowing you to do all kinds of things with the latest 2 commits on the branch you're currently on; in this case your 2 commits.

What you want to do here is look at the second line of that document, this is your latest commit, your 'fixup' commit. What you want to do with the line is replace the `pick` with an `f`, that way, the commit will merge itself into the commit above it. If you want to keep the title and message for the commit you don't replace with an f but with an `s` instead.

When you are done save the document and close it. If you did a squash (`s`), then a new document will open allowing you to configure the title and the message of the commit you just merged into, since you told it you wanted to keep it (by doing a squick and not a fixup)

Hope it makes sense =)

Phatcat:
Oh yeah, when you do squashes and fixups you HAVE to force push it afterwards or you'll have 'merge' issues.

You do that with a simple

`git push -f`

Here is some more info I wrote about merge issues you may find useful: https://github.com/cmangos/mangos-classic/pull/119#issuecomment-223863490

Phatcat:
:tada:

SatyPardus:
Wow first try! :D Thanks.

Phatcat:
Pet patch 1.1 is getting merged. This Pr includes flag handling in PetAI, so you need to add an additional check to the beginning of the overlaying if statement you are introducing;

`!m_creature->HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_DISABLE_ROTATE)`

Phatcat:
Okay, I just tested a modified version of this and it works very well for all kinds of pets, not just hunters pets.

I am going to merge this, but I would like for you to rebase and add that extra check and move the whole functionality inside 'CanReachwithMelee' a bit further down; there is no need to turn to the target until we're close enough to do a melee attack.

Phatcat:
You know what, I'm cleaning up the PetAI anyway, I'll do any moving around I find needed, I'll merge this now, this is fine.

Phatcat:
Merged.

PR #160, opened 2016-06-25
Ensured correct unit facing server side after Distract

katz1:
Issue: https://github.com/cmangos/issues/issues/981

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -2393,11 +2393,15 @@ void Spell::EffectDistract(SpellEffectIndex /*eff_idx*/)
     if (unitTarget->hasUnitState(UNIT_STAT_CAN_NOT_REACT))
         return;
 
-    unitTarget->SetFacingTo(unitTarget->GetAngle(m_targets.m_destX, m_targets.m_destY));
+    float orientation = unitTarget->GetAngle(m_targets.m_destX, m_targets.m_destY);
+    unitTarget->SetFacingTo(orientation);
     unitTarget->clearUnitState(UNIT_STAT_MOVING);
 
     if (unitTarget->GetTypeId() == TYPEID_UNIT)
         unitTarget->GetMotionMaster()->MoveDistract(damage * IN_MILLISECONDS);
+
+    // This is needed to change the facing server side as well (and it must be after the MoveDistract call)
+    unitTarget->SetOrientation(orientation);
 }
 
 void Spell::EffectPickPocket(SpellEffectIndex /*eff_idx*/)


Phatcat:
Is there any particular reason you don't wanna set orientation both server and client side at the same time but rather do other stuff in between?

Besides that it looks fine code-wise.
I am going to do a couple of tests either later tonight or during the weekend and report back.

katz1:
@Phatcat 
There was no particular reason. I just gave a few tests with this code and it seemed OK:

```
void Spell::EffectDistract(SpellEffectIndex /*eff_idx*/)
{
    // Check for possible target
    if (!unitTarget || unitTarget->isInCombat())
        return;

    // target must be OK to do this
    if (unitTarget->hasUnitState(UNIT_STAT_CAN_NOT_REACT))
        return;

    unitTarget->clearUnitState(UNIT_STAT_MOVING);

    if (unitTarget->GetTypeId() == TYPEID_UNIT)
        unitTarget->GetMotionMaster()->MoveDistract(damage * IN_MILLISECONDS);

    float orientation = unitTarget->GetAngle(m_targets.m_destX, m_targets.m_destY);
    unitTarget->SetFacingTo(orientation);
    // This is needed to change the facing server side as well (and it must be after the MoveDistract call)
    unitTarget->SetOrientation(orientation);
}
```

Phatcat:
@cyberium Do you have any problems with this or can I slap it on all 3 cores? - Going to apply the code from the comment and the code from the commit itself. with katz1 as the author, of course.

cyberium:
i think its somewhat redundant and may work because of something weird cause this issue elsewere.
But i have no objection to fix it that way for now.

Phatcat:
I'm with you on that one. It seems like we have to force unit facing quite a few places so something is clearly not working properly somewhere.

killerwife:
As of right now, I do not think such an even should occur, where facing should be desynced server/client wise.

Phatcat:
Which is why we have to merge this, as we cannot do one without doing the other, and currently we are only setting orientation for one without the other, which may be contributing to desyncing.

Phatcat:
unitTarget->SetFacingTo(orientation) **should** be setting the orientation to whatever we tell it, shouldn't it?

Phatcat:
https://github.com/cmangos/mangos-classic/blob/f33e834b852ee08ac4cb54b7868825adcc8b8e07/src/game/Unit.cpp#L3181

that is interesting.

killerwife:
That I have no clue. I was lucky to find it mentioned somewhere that this occurs at all, and why it occurs. When @katz1 reported the issue, i knew what was up.

That being said, from a pure logic standpoint you are right. Has to be tested across expansions tho.

Phatcat:
We set orientation with SetInFront, but not with SetFacingTo 

Why?

killerwife:
I'd try git blame if u have time.

Phatcat:
You're gonna have to explain to me how that works then as I've never done that. Let's take that one over IRC.

Phatcat:
Merged with https://github.com/cmangos/mangos-classic/commit/c642c662b8ad8990cc097f8ad911955f74d92179

PR #164, opened 2016-07-08
Clean up a bit in spell checkcast and spelleffect dummies

Phatcat:
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/164)

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -395,55 +395,45 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
             {
                 case 3360:                                  // Curse of the Eye
                 {
-                    if (!unitTarget)
-                        return;
+                    if (unitTarget)
+                        m_caster->CastSpell(unitTarget, ((unitTarget->getGender() == GENDER_MALE) ? 10651 : 10653), true);
 
-                    uint32 spell_id = (unitTarget->getGender() == GENDER_MALE) ? 10651 : 10653;
-
-                    m_caster->CastSpell(unitTarget, spell_id, true);
                     return;
                 }
                 case 7671:                                  // Transformation (human<->worgen)
                 {
-                    if (!unitTarget)
-                        return;
+                    if (unitTarget) // Transform Visual
+                        unitTarget->CastSpell(unitTarget, 24085, true);
 
-                    // Transform Visual
-                    unitTarget->CastSpell(unitTarget, 24085, true);
                     return;
                 }
                 case 8063:                                  // Deviate Fish
                 {
-                    if (m_caster->GetTypeId() != TYPEID_PLAYER)
-                        return;
-
-                    uint32 spell_id = 0;
-                    switch (urand(1, 5))
+                    if (m_caster->GetTypeId() == TYPEID_PLAYER)
                     {
-                        case 1: spell_id = 8064; break;     // Sleepy
-                        case 2: spell_id = 8065; break;     // Invigorate
-                        case 3: spell_id = 8066; break;     // Shrink
-                        case 4: spell_id = 8067; break;     // Party Time!
-                        case 5: spell_id = 8068; break;     // Healthy Spirit
+                        // 8064 = Sleepy | 8065 = Invigorate | 8066 = Shrink | 8067 = Party Time! | 8068 = Healthy Spirit
+                        const uint32 spell_list[5] = {8064, 8065, 8066, 8067, 8068};
+                        m_caster->CastSpell(m_caster, spell_list[urand(0, 4)], true, nullptr);
                     }
-                    m_caster->CastSpell(m_caster, spell_id, true, nullptr);
+
                     return;
                 }
                 case 8213:                                  // Savory Deviate Delight
                 {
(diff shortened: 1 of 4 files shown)


src/game/SpellEffects.cpp
```diff
@@ -1332,12 +1273,12 @@ void Spell::EffectTriggerSpell(SpellEffectIndex eff_idx)
         }
         // Terrordale Haunting Spirit #2, special case because chance is set to 101% in DBC while description is 55%
         case 23209:
-            if (urand(0, 100) < 55)
+            if (urand(0, 20) < 11)
```
Phatcat: these should have been 19 really.

src/game/SpellEffects.cpp
```diff
@@ -1332,12 +1273,12 @@ void Spell::EffectTriggerSpell(SpellEffectIndex eff_idx)
...
                 m_caster->CastSpell(unitTarget, triggered_spell_id, true);
             return;
         // Terrordale Haunting Spirit #3, special case because chance is set to 101% in DBC while description is 35%
         case 23253:
-            if (urand(0, 100) < 35)
+            if (urand(0, 20) < 7)
```
cala: Though I perfectly agree with this change code-wise, I thought the prior values to be more understandable/related to the comment explaining the exception (that's why I used them).

Phatcat: They however all stack up more resource usage. I'll add a comment explaining why I use the numbers that I do where I use 'obscure' numbers.

killerwife: Well, from my limited understanding, urand does %21 instead of %101 and all of those variables are int anyway so you didnt save anything IMO, this is most likely all dealt with compiler side.

Phatcat: What? - no, it has to find a random number between x and y, the range was 0 -> 100 before, now it's 0 -> 20.

so it would also be the same if we did urand(0, 1000), then?

killerwife: it still calls C function which generates from 0 - 65535 and then fits its using modulo operation

Phatcat: I did not know that, that is good to know; how about when we do a urand(0, 1) - is that a bool or created the same way?

killerwife: same, even some c++ random header can be implemented using C rand

evil-at-wow: Actually urand is something defined in the cmangos sources (in src/shared/Util.{h,cpp}). It always returns an uint32, so urand(0,1) will either return 0 or 1, as an uint32 (you know that's not the same as a bool, but it's convertible to bool just like any integer type).

Internally, in the implementation, the numbers are generated using a [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister). The range doesn't really matter performance wise and you won't really notice a difference between urand(0,1) and urand(0,100000).

killerwife: I knew it was something like that, i looked at it a few months ago, but I guess I was mistaken to think it was done using C rand, anyway, there you have it @Phatcat

Phatcat:
@VladimirMangos I know you are good with logic; does this look all horrible to you?

sidsukana:
@Phatcat operator '&&' checks ALL expressions  unlike '||'.
What an example you would call a good 'source code' here? :)

Phatcat:
Yeah, gonna leave this up for a few days then I'm gonna merge it; these commits shouldn't change functionality in any way what-so-ever, only simplify codeflow.

Blumfield:
did a quick look, seems ok to me

Phatcat:
@Blumfield Thank you, I appreciate that.
@dagochen Would you possibly look this over for me as well? with a couple of people like you having looked it over I would be comfortable pushing it.

Also any comment about me possibly regressing code style or readability in some areas (like @cala 's comment (which should have been addressed, by the way)) is also greatly appreciated.

Lillecarl:
*[src/game/MotionMaster.cpp, line 365 at r4](https://reviewable.io:443/reviews/cmangos/mangos-classic/164#-KSjYOu_BN8OgJagZFjB:-KSjYOu_BN8OgJagZFjC:bz662rv) ([raw file](https://github.com/cmangos/mangos-classic/blob/979a9a54e5e94e0f034b6de1f027c57e9b2aa46d/src/game/MotionMaster.cpp#L365)):*
> ```cpp
>     DEBUG_FILTER_LOG(LOG_FILTER_AI_AND_MOVEGENSS, "%s flee from %s", m_owner->GetGuidStr().c_str(), enemy->GetGuidStr().c_str());
> 
>     if (m_owner->GetTypeId() == TYPEID_PLAYER)
> ```

Are the only options TYPEID_PLAYER and TYPEID_UNIT?

The reason i'm asking is because as shown here, players are threated one way, and everything else another.

After your change Creatures are threated one way, and everything is threated another way.

Did you consider this while changing this bit? :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cmangos/mangos-classic/164)*

Phatcat:
I didnt merge the motionmaster nor the aura changes, no :)

PR #168, opened 2016-07-14
Fix for the so-called "target bug"

Warlockbugs:
Inspect handler should not silently (server-side only) change player's current selection. This behaviour leads to unwanted side-effects such as summons landing on wrong group members. Its intended for inspect target and real player target to be separate, as many addons rely on background inspect queries to function properly.

### DIFF
--- a/src/game/MiscHandler.cpp
+++ b/src/game/MiscHandler.cpp
@@ -962,8 +962,6 @@ void WorldSession::HandleInspectOpcode(WorldPacket& recv_data)
     recv_data >> guid;
     DEBUG_LOG("Inspected guid is %s", guid.GetString().c_str());
 
-    _player->SetSelectionGuid(guid);
-
     Player* plr = sObjectMgr.GetPlayer(guid);
     if (!plr)                                               // wrong player
         return;


Phatcat:
And so just not doing it will have the desired effect?
I'm not saying what we're doing currently is right, but by removing this, would that not result in the inspected target within the inspect window changing when the player changes target?

also, there's no reason to PR the same change for more than 1 core, if it's applicable to more than 1 core we just port it when merging.

Phatcat:
you can ignore that the check is failing by the way, it's unrelated to these changes, it's to do with appveyor not being a terrific system all the time.

Warlockbugs:
The change has nothing to do with inspect frame. Inspect frame is not a problem. What this action really does is it desyncs player's target on client and player's target on the server by changing player's target on any inspect query, not just inspect frame.
There is a WoW API function NotifyInspect, which is often used by many addons to scan players in background.
For example, a tooltip addon displays talent spec data of the player you are hovering your mouse on, it calls this function every time mouseover unit changes. And every time server assigns your target to the mouseover, no matter what is your real target! This causes many delayed spells (like summon effect, which procs 5 sec after the cast is done) to land on last inspected target (which is 99% of the time is done by some addon, not inspect frame, i.e. not target you really inspected or really intended to target), instead of your real current target client-side.
I was a regular warlock raider on a private TBC server for 4 years with 150+ days played, and I can say with all my responsibility that this little thing wasted so many soul shards and ate so much nerves, that it is above and beyond my ability to express in english.
What i was doing to fight this issue from client side (fighting the windmill, i know, but...) is to create a script, which hooks NotifyInspect function and effectively supresses any inspect calls to the real Blizzard NotifyInspect function if summoning spell is currently casting or ended less than 7 seconds ago (giving the time to the server to apply that 5s delay).
Also i had to re-macro Ritual of Summoning to pair it with forced target update (explicitly clear target and then target last target) to make server dump last inspected target and target my real instead.
This script and macro were able to circumvent 95% of the wrongly summoned players, but sometimes addons were able to sneak an inspect call in between summon cast sent and server response for the spellcast start and... :rage:
Sorry for the wall of text, i wanted to give a full insight on the issue instead of regular "hurr durr summon bugged" you usually see in bug reports.

Phatcat:
The question is, though, if we rely on getting that selection somewhere else in our system, which we very likely do (or at least used to) since it was added by someone; we need to make sure things don't fall apart if we merge this (which I very much think we should; this is actually abusable, as I have just realized now how people were able to abuse meeting/summon stones on a private server I used to play on a couple of years ago.)

Warlockbugs:
I am not a mangos dev by any margin, but from what it looks like, this was a hack to not alter another "convenient" function for forming something that was sent in the response or fundamental misunderstanding that inspect can be used outside of inspect frame for any unitid WoW UI sees without "target" attachment.

Phatcat:
Merged.

PR #169, opened 2016-07-18
Implement SMSG_PVP_CREDIT opcode

sidsukana:
It's PvP frame on honorable kills

### DIFF
--- a/src/game/BattleGround/BattleGround.cpp
+++ b/src/game/BattleGround/BattleGround.cpp
@@ -1169,7 +1169,7 @@ void BattleGround::UpdatePlayerScore(Player* Source, uint32 type, uint32 value)
             break;
         case SCORE_BONUS_HONOR:                             // Honor bonus
             // reward honor instantly
-            if (Source->AddHonorCP(value, HONORABLE, 0, 0))
+            if (Source->AddHonorCP(value, HONORABLE))
                 itr->second->BonusHonor += value;
             break;
         default:
--- a/src/game/OutdoorPvP/OutdoorPvPEP.cpp
+++ b/src/game/OutdoorPvP/OutdoorPvPEP.cpp
@@ -175,7 +175,7 @@ void OutdoorPvPEP::HandleObjectiveComplete(uint32 eventId, const std::list<Playe
         if ((*itr) && (*itr)->GetTeam() == team)
         {
             (*itr)->KilledMonsterCredit(credit);
-            (*itr)->AddHonorCP(HONOR_REWARD_PLAGUELANDS, HONORABLE, 0, 0);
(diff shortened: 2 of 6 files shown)


Phatcat:
Could you possibly go a bit more into detail with this does exactly? How can I best see the changes in effect?

dagochen:
From what the code changes I would try to do a dishonorable kill ...?

Phatcat:
is there something wrong with the current way we do dishonorable kills? - but thanks, I'll try that.

dagochen:
Thats something you should ask @sidsukana :) Maybe dishonorable kills were not shown in combatllog or so....? Dunno...never checked that^^

Phatcat:
would be nice to know exactly what I was looking at, the information is a bit sparse on this one.

sidsukana:
Kill honorable target in PvP - popup frame with icon and name of target rank.
For example https://youtu.be/IyeoT1owdFo?t=508 HK:Senior Sergeant

Phatcat:
> popup frame with icon and name of target rank.
Say no more, I don't even have to look at the video, I know exactly what you are talking about now. Don't we do those at all on cmangos? If that is the case then, nice job, sid!

How was this handled for TBC? It showed the rating of your opponent if your opponent was a player, no? Or am I just imagining things?

sidsukana:
> Say no more, I don't even have to look at the video, I know exactly what you are talking about now. 
> Don't we do those at all on cmangos? If that is the case then, nice job, sid!

PvP system in ТВС and wotlk was greatly changed in comparison with the classic. I know that in the TBС and wotlk also have a pop-up frames, but not sure that they have the same structure in the classics.

(If I understand you :))

killerwife:
We do not need this :) The packets version in TBC is implemented properly.

Phatcat:
I'm guessing it didn't change much between TBC and Wrath? (that it's properly implemented for wrath as well)

killerwife:
Well the only reason this isnt implemented is because of wotlk branch. So nope.

Phatcat:
Alright so I tested this out, and sadly I did not get the pvp pop-up...

I took my freshly created warrior from both factions, gave them a few levels, teleported the to the same area, set them both to pvp, then I set one to kill the other; no pop-up..

It is very much supposed to pop up even if your kill has no rank at all https://www.youtube.com/watch?v=FoAsyv6QpSs (0:30)

Phatcat:
No wait scratch that, it works fine, there was a level difference between them making the target gray....

Phatcat:
Merged with https://github.com/cmangos/mangos-classic/commit/979a9a54e5e94e0f034b6de1f027c57e9b2aa46d

PR #171, opened 2016-07-18
fix compile on linux

shagu:
add missing include which is required when compiling on linux.

See http://cmangos.net/thread-7398-post-45792.html and https://github.com/cmangos/issues/issues/1001 for more information.

### DIFF
--- a/src/game/vmap/BIH.cpp
+++ b/src/game/vmap/BIH.cpp
@@ -19,6 +19,7 @@
 #include "BIH.h"
 
 #include <cmath>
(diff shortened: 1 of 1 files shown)


Muehe:
Hey, nice to see you here!

Could you add the fourth include I noted in the issue please? The first two have conflicting commits on Wotlk branch, but the `src/game/vmap/WorldModel.cpp` one should be fine.

shagu:
Sorry, It didn't built for win32 and I have no motivation nor windows to work on that any further.

PR #172, opened 2016-07-18
fix crash when using eyes of the beast

shagu:
fix server crashes whenever a hunter uses "eyes of the beast"

see https://github.com/cmangos/issues/issues/968 for more information.

### DIFF
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -2127,7 +2127,7 @@ void Aura::HandleModPossessPet(bool apply, bool Real)
 
     if (apply)
     {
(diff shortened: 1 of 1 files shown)


cyberium:
Thank for your efforts.

This will not fix the issue, its an hack to avoid using that part of the code. You better comment it or replace the if condition by 'if (0)'....

This will need more effort to be realy fixed. :/

shagu:
It wasn't a big effort, I just committed and PR'ed solutions I found in the issue tracker which solved my problems and stabilized the server, because it looked like that nobody else does it. 
I didn't looked closely enough at it, Sorry.

Thanks for your answer :)

PR #174, opened 2016-07-22
Rewrite Config to "pimpl" idiom

sidsukana:
It solve warning C4251

### DIFF
--- a/src/shared/Config/Config.cpp
+++ b/src/shared/Config/Config.cpp
@@ -17,33 +17,56 @@
  */
 
 #include "Config.h"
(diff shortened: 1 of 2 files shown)


sidsukana:
Restore STD using.

Phatcat:
So... what does this do exactly? besides solving a warning?

sidsukana:
Only C4251, but only for Config class (as example for resolve some warnings wihout disable)

cyberium:
no more needed Thanks @sidsukana for the effort

PR #175, opened 2016-07-22
Remove external Mersenne Twister

sidsukana:
Replace to internal boost::random::mersenne_twister


killerwife:
Dont even use boost, use STD mersenne twister.

Phatcat:
Why?

We are dependant on boost, no? - so why not just embrace that?

Is it worth the 450ish lines to have our own solution when we can just pull one from boost?

killerwife:
Because boost can be removed in the future, with newest c++ standards stuff. We are using STD everywhere we can, c++11 rewrites were the reason for that, so in my opinion STD>Boost>other stuff. I used #include <random> for all I needed for doing very heavy simulations, hence I can vouch for its functionality.

TL;DR in the future we might get rid of boost, but we will never get rid of STD

sidsukana:
Hmm, I had a different opinion on this)
Boost optimized library that promotes new standards for C++. It is more updated and optimized. When will be the next new STD version?

sidsukana:
I would like to know what the code development policies adhere to Mangos?

killerwife:
Well, lets summon the higher ups @cyberium @namreeb since they are the ones working with this stuff. In the end both generators are most likely the same. I just prefer to use STD when I can.

sidsukana:
@killerwife i agree) i create topic on forum http://cmangos.net/thread-7528.html

sidsukana:
Rework for std::mt19937
I have my doubts about the seed. I used seed_seq of 2 values: time() and clock(). Is this enough to be?

killerwife:
I used std::random_device, but this also has its drawbacks. As it is implementation dependant, on MSVC it gives a higher precision seed than clock(), but may very well be different on GCC. Generally you want nanotime, time and clock are seconds afaik.

sidsukana:
@killerwife 
I read about the advantages and disadvantages random_device and decided to stay until the district is that used Mangos (but without the hash values itih) instead I decided to use seed_seq.
Alternatively, you can add additional source - random_device

Phatcat:
So what is the verdict on this one? Is it good to merge or how are we? @sidsukana @killerwife @namreeb @evil-at-wow ?

Phatcat:
@cyberium @VladimirMangos @xfurry last callout before merging this.

PR #177, opened 2016-07-27
threat spells added to combatlog

dagochen:
And numero duo ;) I am so happy its working^^

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3371,6 +3371,13 @@ void Spell::SendLogExecute()
         {
             switch (m_spellInfo->Effect[EFFECT_INDEX_0])
             {
+                case SPELL_EFFECT_DISPEL:
+                case SPELL_EFFECT_THREAT:
+                    if (Unit* unit = m_targets.getUnitTarget())
+                        data << unit->GetObjectGuid();
+                    else
+                        data << uint8(0);
+                    break;
                 case SPELL_EFFECT_POWER_DRAIN:
                     if (Unit* unit = m_targets.getUnitTarget())
                         data << unit->GetPackGUID();


dagochen:
1. get a rogue
2. learn faint
3. go to a mob and hit it
4. check combatlog, it should say something like "you perform faint on <target>."

same for decurse by mages for dispells.

src/game/Spell.cpp
```diff
@@ -3371,6 +3371,13 @@ void Spell::SendLogExecute()
         {
             switch (m_spellInfo->Effect[EFFECT_INDEX_0])
             {
+                case SPELL_EFFECT_DISPEL:
+                case SPELL_EFFECT_THREAT:
```
Phatcat: You may want to try and move these two cases down here: https://github.com/cmangos/mangos-classic/blob/0ed121835e5b967f4976ec2d20531529be5243f0/src/game/Spell.cpp#L3452 because then you should be able to remove the functionality you added in the following lines, as those cases by the link go through basically the same.

Phatcat: I see it uses
`data << unit->GetPackGUID();`
where you use
`data << unit->GetObjectGuid();`

Hmmm, my comment may be bullocks then, but I still think it's worth the try if you havn't already.

dagochen:
Nope. GetPackGuid does not work.

Phatcat:
Small issue with this one (may be a general issue actually)

We inform the player twice about a melee spell if it misses (at least on this PR with feint)
feint

dagochen:
Oh. That should not happen I guess. I ll check out how to avoid it.

I found this piece of code where SendLogExecute is called. I think the boolean should be false in cases where we already have send some log for that spell before. Whats your call on that?

`void Spell::_handle_finish_phase()
{
    // spell log
    if (m_needSpellLog)
        SendLogExecute();
}`

Phatcat:
Are there examples of other spells getting suppressed this way, perhaps on crit or on dodge or likewise cases?

How does it currently look with other spells that are reported currently when they miss... I'll go check real quick.

Phatcat:
That's weird, I simply can't get offensive spells to miss on cmangos classic, I can only get them to either hit or resist (which makes me think of partial resisting and if that is implemented) - but on resist only the resist message is reported, the normal spell message is not sent.

It however looks like we also double report in other cases, though a bit differently; on hitting an evading mob we report thse two: (which may be correct, though, I'm just pretty sure an amount should be reported on evade (or absorbed) as well..

evade
resist

Phatcat:
That's weird and slightly conflicting on those 2 cases, one showing how 'absorbed' is sent to the player and then the next screenshot showing something else. ´Same spell. Great.

Phatcat:
Yeah, spells missed are only reported once;

sinister

It did however reveal to me also that rogues when doing sinister strikes and other instant melee attacks (melee attacks that are not 'on next swing' it will automatically proc weapon hit at the same time, without resetting normal attacking time either, making the rogue the meanest, fastest killer ever... that seems.... broken... It's hard capturing that with a screenshot, just make a level 1 rogue and go fight some mobs and you'll see and feel it instantly.

also this seems like a weird 'dodge on normal hit' message to me that we send out; it seems very narritive and very much out of place with the rest of the combat messages which carry themselves more straight forward and basic.
dodged

Sorry for all of this semi-unrelated stuff..

Phatcat:
Also our messages don't follow a consistent concept of time, switching between past and present tense. (you miss (present) / you missed (past))

dagochen:
I think the time thing about present and past cant be controlled by the server. I would guess its clientside...?

dagochen:
The time thing should not be changed as long as there are no indices by combatlog parsing addons which proove it to be wrong or we might stop them from working correctly :)

Phatcat:
Indeed. if these are handled client side well then yes, my comment about the style of the messages can be ignored.

> And if we make something like "if hitinfo != SPELL_OK return" inside sendspelllog?

Question is what other effects, if any at all, on that case list also could be reporting double, and may benefit from this change.

Phatcat:
Also I am not able to tell you wether this is the right direction to go with the broader issue at hand here or not.

You most likely want to get a hold of one of the developers who are more familiar with the spell system or server/client communication I'd think. Not exactly sure who in this case.

dagochen:
Also for me it somehow feels wrong that there is Unit::SendSpellMiss, which handles something spelllrelated in Unit, while most of the other Sendxy methods are in Spell::.... Shouldn't this be handled all in Spell and not in Unit?

Phatcat:
Exactly my thought as well, but I do not know enough about the system to be sure; i think it has to do with the fact that auto attack is not considered a spell, I'm not sure.

dagochen:
I also believe that there are several cases in sendlogexecute which are not working completly. SPELL_EFFECT_INTERRUPT_CAST e.g. should propably send into combatlog something like 
"you interrupted target's firebolt." or so. Not sure if that is working at the moment...but i guess not, since we dont send the spellid: data << uint32(0);                      // spellid

Laizerox:
I am working already on fixing majority of SendLogExecute and order is shown in how are the packets sents.

What comes to messages its all client side and all the texts are hardcoded into client.

Also ObjectGuid() and Pack() are 2 different things and fixing packets would be nice if you checked them also in client to be 100% sure. Yes it wont work until we send proper data.

Misses and other hit info is written is Spell:: using packets SMSG_SPELL_GO or SMSG_SPELL_START since TBC.

Afaik Unit function is used in special cases, I could be wrong here since I don't remember now but I am certain miss info atleast in TBC is sent in above packets. In classic I reckon we needed to send explicit miss info (Absorsb, resist, evade, miss).

Also Unit function is given a caster perspective. If you moved it into Spell you would also need to pass realCaster or store it and since currently all calls happen using realCaster.

dagochen:
@Laizerox
Anything that can be done here to make your work easier? Or shall we close this PR then and wait for your changes  ?

Laizerox:
@dagochen probably no because we need to build SendLogExecute in completely different manner than we're doing now.

I'm only now struggling with summonable objects guids (Pets etc).

Also with my small test results I was able to show extra attacks in combat log also. "You gain x extra attack from %spell"

dagochen:
So we can close this one here?

Phatcat:
I would say leave this open as a reminder, even if we don't end up merging this PR I see no reason to close it until a better substitute has been PR'd / merged.

mityada:
There is a bigger problem here. The packet structure for SMSG_SPELLLOGEXECUTE is from TBC and is wrong for 1.12.1 client. First of all, 1.12.1 client expects guids as ObjectGuid, not PackedGuid. Then, the set of effects is quite different. There are no SPELL_EFFECT_SUMMON_*, but there are some others, including SPELL_EFFECT_DISPEL and SPELL_EFFECT_THREAT @dagochen wanted to add.

I have corrected the structure here: https://github.com/mityada/mangos-classic/commit/68ae21a5d089ef1872f8dd9528056adde0ed6543. I'm not creating a PR yet, as logs still need the correct data instead of zeros to actually work.

cyberium:
is replaced by https://github.com/cmangos/mangos-classic/pull/298

PR #178, opened 2016-08-08
Fix some texts (Majordomo and server string)

Phatcat:
from @sidsukana and @Tobschinski 

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/178)


sql/scriptdev2/scriptdev2.sql
```diff
@@ -2038,7 +2038,7 @@ INSERT INTO script_texts (entry,content_default,sound,type,language,emote,commen
 (-1409004,'The runes of warding have been destroyed! Hunt down the infedels my bretheren.',8039,6,0,0,'majordomo SAY_SPAWN'),
 (-1409005,'Ashes to Ashes!',8037,1,0,0,'majordomo SAY_SLAY'),
 (-1409006,'Burn mortals! Burn for this transgression!',8036,1,0,0,'majordomo SAY_SPECIAL'),
-(-1409007,'Impossible! Stay your attack mortals! I submitt! I submitt!',8038,1,0,0,'majordomo SAY_DEFEAT_1'),
+(-1409007,'Impossible! Stay your attack, mortals... I submit! I submit! Brashly, you have come to wrest the secrets of the Living Flame! You will soon regret the recklessness of your quest. I go now to summon the lord whose house this is. Should you seek an audience with him, your paltry lives will surely be forfeit! Nevertheless, seek out his lair, if you dare!',8038,1,0,0,'majordomo SAY_DEFEAT_1'),
```
Tobschinski: This is different, I'm afraid. Although the soundfile contains all three strings, they are indeed split into 3 different ones.

Phatcat: So it's more a case of missing the last string altogether which should then contain what has been added to the second string here? - or is there a 1409008 ?

Tobschinski: They exist: -1409020 & -1409021. However, all of the used strings (even the ones not mentioned here) have either spelling of punctuation mistakes.

Phatcat: Ah, so a much larger update is in need actually. Well, our raiding group sucked and we only ever got to do raggy a couple of times, I think I only experienced that encounter once.

Unless the strings have been changed on current retail I guess the correct spellings could be gotten from there.

Phatcat:
@cala @X-Savior @Grz3s

This update is relevant to the dbs.

killerwife:
Afaik these things are kept in core file so DB providers can be chill about it. Just like the coeffs stuff.

Phatcat:
https://github.com/cmangos/mangos-classic/pull/178/commits/702a044539ae79f0d4a0b118af9bbba2669273ee#diff-19d5e01df7226bec976ff28e83a4bbefL3637 as well ?

killerwife:
y

PR #179, opened 2016-08-08
Honor

Phatcat:
Honor log work by @kraminpouds

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/179)


Phatcat:
This may honestly be a bit much, sending all that honor info into the server log. I don't really feel that strongly about this change to be honest.

PR #180, opened 2016-08-08
Game/Spell: Handle SPELL_FAILED_NOTHING_TO_DISPEL

Phatcat:
By @Caydan

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/180)

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4041,8 +4041,7 @@ SpellCastResult Spell::CheckCast(bool strict)
         // totem immunity for channeled spells(needs to be before spell cast)
         // spell attribs for player channeled spells
         if (m_spellInfo->HasAttribute(SPELL_ATTR_EX_CHANNEL_TRACK_TARGET)
-                && target->GetTypeId() == TYPEID_UNIT
-                && ((Creature*)target)->IsTotem())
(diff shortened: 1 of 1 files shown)


Phatcat:
Anyone else care to look at this? In my opinion this looks decent.

dagochen:
What is the problem which shall be fixed by this?

Phatcat:
As I understand it currently the player is not informed properly if there is no no debuff to remove when using dispel effects, as in that error message is not implemented anywhere in the code, it's only defined:

https://github.com/cmangos/mangos-classic/search?utf8=%E2%9C%93&q=SPELL_FAILED_NOTHING_TO_DISPEL

We do get a message that 'Nothing to dispel', so....

Phatcat:
This seems to be an issue exclusive to classic, even though SPELL_FAILED_NOTHING_TO_DISPEL is nowhere to be found besides in the defines.
It seems to be handled fine on tbc and wrath (perhaps the check was moved client side, or I am missing something obvious perhaps)

you can reproduce an error by making a druid
`.learn 8946`
cast it on yourself
it shouldn't allow it to be cast, but rather display 'Nothing to disspel'
With this patch it displays that when appropriate.
On tbc and wrath branches, again, it seems to already act appropriate, even though that spell cast result is not handled anywhere serverside..

PR #182, opened 2016-08-09
Fix double login crash, issue #979

dufernst:


dufernst:
Note that I don't check `PlayerLoading()` as is currently the case in `RemoveSession`.
For as far as I can see `PlayerLoading()` will always be false anyway.

sidsukana:
works fine!

dufernst:
I think my code makes more sense than the previous code and it solves at least one issue.
@namreeb do you know if the `PlayerLoading()` check was required?
When I am looking at the code surrounding `m_playerLoading` in `CharacterHandler.cpp` I get the feeling that there are some threading issues that went unnoticed.

For example the async tasks during player login. The callback functions do check if a session exists for a given account ID but not if it is the same session that sent the character login packet.

impowski:
Didn't fix anything for me. I always get this exception.
` terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  close: Bad file descriptor

impowski:
It's start to drop after socket closing in Socket.cpp.

dufernst:
@impowski could you provide me with the complete callstack?
@Phatcat  they indeed most definitely won't solve the issue completely, I hope @namreeb can give me some insights on his thoughts on dealing with `PlayerLoading`.

impowski:
@dufernst Just debug it and you will see.

dufernst:
Unfortunately I don't use mangos classic myself, my "fix" is solely based on me looking at the code.

dazidkk:
http://pastebin.com/LndeML2Q

Phatcat:
@dufernst 
Do you want to continue working on this considering the changes recently pushed to master? - The issue should be closed and fixed, but that doesn't mean what you were going for here isn't an improvement.

Either way I wanted to tell you that you have merge conflicts due to recent changes

dufernst:
If the changes to master now fix the issue I am probably not going to continue.
The way I was considering fixing it requires quite a bit of refactoring so would not happen/be finished any time soon.

Phatcat:
That is quite alright, I hope you don't mind me closing this for now, then.

Feel try to re-open this or re-PR something whenever you feel it's ready if you do decide to work on it.

PR #183, opened 2016-08-09
Lightning Shield Fix

dazidkk:
Fixes Lightning Shield on Lightning Shield not proccing all the charges
at once.


killerwife:
Tabs

PR #184, opened 2016-08-09
Lightning Shield Fix

dazidkk:
Fixes Lightning Shield on Lightning Shield not proccing all the charges
at once.

### DIFF
--- a/src/game/UnitAuraProcHandler.cpp
+++ b/src/game/UnitAuraProcHandler.cpp
@@ -1085,6 +1085,9 @@ SpellAuraProcResult Unit::HandleProcTriggerSpellAuraProc(Unit* pVictim, uint32 d
             // Lightning Shield (overwrite non existing triggered spell call in spell.dbc
             if (auraSpellInfo->IsFitToFamilyMask(uint64(0x0000000000000400)) && auraSpellInfo->SpellVisual == 37)
             {
+                if (procSpell && procSpell->SpellFamilyFlags & uint64(0x0000000000000400))
+                    return SPELL_AURA_PROC_FAILED;
+
                 switch (auraSpellInfo->Id)
                 {
                     case 324:                           // Rank 1


Phatcat:
@cyberium @killerwife @Laizerox Any one of you guys familiar enough with this part of the spell system (well, more aura system, I wouldn't know honestly) to be able to either yay or nay the method of this tiny change here?

Phatcat:
It addresses this issue; https://github.com/cmangos/issues/issues/1021

Phatcat:
you sure it's enough to check the family flag and that it's a procced spell to know that it's lightning shield triggering lightning shield? are there no other spells in that spell family?

killerwife:
Actually, I might have a Spell Attribute blanket fix that will make this obsolete :)

killerwife:
But It will take me 1-2 weeks to get around to it.

killerwife:
@dazidkk can you please check this query in your DB?
select * from spell_proc_event where entry=324;

Phatcat:
@killerwife well, we are in no hurry on this one, I guess, so if you can fix this in a cleaner way than this then sure, go for it, or you could give dazidkk some pointers and he could try to improve his PR? up to you.

killerwife:
I took a closer look at the issue and my research found the following:
1. the fix is there because of missing proc spell in tbc and wotlk spell.dbc, the whole lightning shield handler is basically one hack, it will be possible to remove the hack once we migrate spell.dbc to SQL and can insert this missing spell on tbc/wotlk branches
2. the issue arises from proc system proccing on lightning shield in a chain reaction for some reason. Chain lightning should only proc on DAMAGE_TAKEN (i.e. unit takes damage from outside source) and never on DAMAGE_DONE, and since damage is dealt to original damage dealer, i do not see how this can be reasonable behaviour

Therefore the advised solution is to dig deep into proc system and find out why there is a chain proc reaction, and not merely hide the problem :) As this is classic core, I only know how it should work (proc logic is same on TBC where i work)

cyberium:
Should be fixed

PR #185, opened 2016-08-14
Fix RA input processing

mityada:
There is a bug in RASocket introduced in eeaf6d6, which causes the last character of the input to be truncated. It makes RA unusable unless you append an extra character to your input.

It would be enough to just fix this line
https://github.com/cmangos/mangos-classic/blob/master/src/mangosd/RASocket.cpp#L91
But I've found another flaw in this code, so I've decided to refactor the input processing to make it more clear.

### DIFF
--- a/src/mangosd/RASocket.cpp
+++ b/src/mangosd/RASocket.cpp
@@ -40,8 +40,6 @@ RASocket::RASocket(boost::asio::io_service &service, std::function<void(Socket *
     m_secure(sConfig.GetBoolDefault("RA.Secure", true)), MaNGOS::Socket(service, closeHandler),
     m_authLevel(AuthLevel::None), m_accountId(0), m_accountLevel(AccountTypes::SEC_PLAYER)
 {
-    m_commandBuffer.reserve(InitialBufferSize);
-
     if (sConfig.IsSet("RA.Stricted"))
     {
         sLog.outError("Deprecated config option RA.Stricted being used.  Use RA.Restricted instead.");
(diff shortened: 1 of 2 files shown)


src/mangosd/RASocket.cpp
```diff
@@ -76,48 +74,48 @@ bool RASocket::Open()
...
-        std::copy(buffer.cbegin(), buffer.cend(), std::back_inserter(m_commandBuffer));
-        return true;
+        if (!HandleInput())
+            return false;
     }
 
+    return true;
+}
+
+bool RASocket::HandleInput() {
```
Phatcat: `{` should have its own line =)

Phatcat:
Besides the one tiny note about the style I think this looks good and should be merged. Anyone who doesn't agree should let it be known within a couple of days or they very well find that I have merged this.

src/mangosd/RASocket.cpp
```diff
@@ -76,48 +74,48 @@ bool RASocket::Open()
...
-    {
-        if (buffer[newLine] == '\r' || buffer[newLine] == '\n')
-        {
-            if (newLine > 0)
-                std::copy(buffer.cbegin(), buffer.cbegin() + newLine - 1, std::back_inserter(m_commandBuffer));
+    static const std::string NEWLINE = "\n\r";
+
+    auto pos = 0;
+
+    while (pos != std::string::npos) {
```
mityada: @Phatcat, you haven't noticed this one =)

Phatcat: you're right. I didn't. this { would of course also need its own newline ^^

Phatcat:
so, since you referenced this in that issue on the wrath branch (https://github.com/cmangos/issues/issues/877), have you tested this code against the wrath core and confirmed that it solved the issue? Or is this issue also present on classic and it solves it here? (I have not looked at the issue at all, just seeing the ref)

mityada:
All three cores (classic, tbc & wotlk) have exactly the same code for RASocket, so this patch can be applied to all without any changes. But to be honest, I tested it only for mangos-classic.

PR #190, opened 2016-09-03
Add script for Gates traps in Stratholme

cala:
The two gateways in Stratholme (in Scarlet side and in Undead side) are
now triggering the traps as expected:
- the gates close around the player(s)
- the plagued critters spawn and attack
- the gates reopen after 20 secs
- traps are on 30 min cooldown

### DIFF
--- a/src/scriptdev2/scripts/eastern_kingdoms/stratholme/instance_stratholme.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/stratholme/instance_stratholme.cpp
@@ -35,7 +35,8 @@ instance_stratholme::instance_stratholme(Map* pMap) : ScriptedInstance(pMap),
     m_uiMindlessCount(0),
     m_uiPostboxesUsed(0),
     m_uiAuriusSummonTimer(0),
-    m_bIsSlaughterDoorOpen(false)
+    m_bIsSlaughterDoorOpen(false),
+    m_uiGateTrapTimers{{0,0,0},{0,0,0}}
 {
     Initialize();
 }
@@ -188,6 +189,10 @@ void instance_stratholme::OnObjectCreate(GameObject* pGo)
             break;
         case GO_PORT_ELDERS:
         case GO_YSIDA_CAGE:
+        case GO_PORT_TRAP_GATE_1:
+        case GO_PORT_TRAP_GATE_2:
+        case GO_PORT_TRAP_GATE_3:
+        case GO_PORT_TRAP_GATE_4:
             break;
 
         default:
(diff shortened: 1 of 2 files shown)


cala:
@xfurry I'd like to have your opinion here.
This is the first time I script a server-side trap and I don't know if this is the best way to do it. Maybe, you will have a better idea.

In every call of Update(), I test the two traps by:
- checking if they are on cooldown (timer != 0). If not, I check if a player or pet is in range of the center of each trapped hallway. If so, I "trigger" the trap by closing the doors of the hallway and setting the timers for this trap (1 for spawning the critters, 1 for reopening the gate and 1 for resetting the trap)

I'm wondering if I should really do this at every call of Update() or set a timer (another one) to do this less often. What do you think?

The in-game behaviour is ok (timers, radius, NPCs/GO entries) though.

Phatcat:
I know you asked for xfurrys opinion explicitly, but I can't help myself. Also I don't think we should do this check on every update() call no, I think indeed a timer should be in place, checking every x amount of seconds (1 - 2) - this also seems to be fit better with videos as this fits nicely with the fact that you were able to bypass them if you moved fast, and it doesn't appear to trigger at the exact same time in all the videos.

xfurry:
Well, while I was trying to implement this event myself, I found the same problem
There is no way to check when a player is in the right spot, other than checking the distance between the player and some GO or NPC or map. Doing this is kind of ugly, but I guess we have no other choice.

Your code looks good. :+1:

cala:
Thank you both of you for this quick feedback ! If only I could it be that fast at work... :wink:

@Phatcat I asked for @xfurry's feedback because he is the one with the most experience with SD2 and maybe he would have already see/done something similar. But everyone's feedback is welcome of course ! :smile: 

In fact, the feedback you've provided in the issue thread and the videos were very useful. My first idea was to trigger the trap when the player was between the two gates but after seeing the videos you linked, I realised that what I was seeing was the result of the interaction of the trajectory of a player in a rectangular gateway with the circular activation zone of a trap.
For example, the gateway is 18.5 yards long and 10 yards wide. In the middle, I placed a trap with a 5 yards radius. When players enter in the middle of the gateway width, they trigger the trap after 4 yards in. But, if they enter by staying close to one of the side wall, they trigger the trap after walking more than 9 yards in. If they have the sprint effect, they can even go out before the gates are finishing closing (they need 1 second).

I tried to use a timer to not check on every update, but it allowed some players to cross the gateway without even triggering the trap.

Thanks again for your feedback, I did a few more tuning after that and this is now merged and working as shown in @Phatcat  videos. :+1:

PR #191, opened 2016-09-05
Implement PIN-based two-factor authentication

Chaosvex:
_Update: TOTP ported._
_Note: This PR is currently for reviewing/feedback rather than merging._

This is an implementation of the two-factor PIN system that's available in the 1.xx clients. It should also be available in TBC, WotLK and possibly beyond.

I've ported it directly from my own original implementation (https://github.com/EmberEmu/Ember/blob/spark-new/src/login/PINAuthenticator.cpp)/

So you can see how it behaves without running it yourself:
https://www.youtube.com/watch?v=_9IS44BGgJU      - Fixed PIN
https://www.youtube.com/watch?v=VXAOxSBGFXg   - TOTP

I'll squash the commits before the final PR. :)

Thanks!

### DIFF
--- a/sql/base/realmd.sql
+++ b/sql/base/realmd.sql
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.6.17, for Win64 (x86_64)
+-- MySQL dump 10.13
 --
--- Host: 127.0.0.1    Database: realmd3
+-- Host: localhost    Database: realmd
 -- ------------------------------------------------------
--- Server version 5.6.21
+-- Server version	5.6.21
 
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
@@ -40,8 +40,7 @@ CREATE TABLE `account` (
   `expansion` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `mutetime` bigint(40) unsigned NOT NULL DEFAULT '0',
   `locale` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `security` int(11) DEFAULT '0',
-  `pin` int(11) DEFAULT '0',
+  `security` varchar(16) DEFAULT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `idx_username` (`username`),
   KEY `idx_gmlevel` (`gmlevel`)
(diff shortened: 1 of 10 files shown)


sql/base/realmd.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13
+-- MySQL dump 10.13  Distrib 5.6.17, for Win64 (x86_64)
```
Phatcat: This

sql/base/realmd.sql
```diff
@@ -1,8 +1,8 @@
...
 --
--- Host: localhost    Database: realmd
+-- Host: 127.0.0.1    Database: realmd3
```
Phatcat: this

sql/base/realmd.sql
```diff
@@ -1,8 +1,8 @@
...
 -- ------------------------------------------------------
--- Server version	5.5.32
+-- Server version 5.6.21
```
Phatcat: and this, probably has to go.

sql/base/realmd.sql
```diff
@@ -16,30 +16,12 @@
 /*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
 
 --
--- Table structure for table `realmd_db_version`
```
Phatcat: Why did you remove all of this?

sql/base/realmd.sql
```diff
@@ -16,30 +16,12 @@
...
-(NULL);
-/*!40000 ALTER TABLE `realmd_db_version` ENABLE KEYS */;
-UNLOCK TABLES;
-
---
 -- Table structure for table `account`
 --
 
 DROP TABLE IF EXISTS `account`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
```
Phatcat: you've added a lot of these kind of lines which are prety much all commented out ?

src/realmd/AuthSocket.cpp
```diff
@@ -511,6 +513,13 @@ bool AuthSocket::_HandleLogonProof()
     if (!Read((char*)&lp, sizeof(sAuthLogonProof_C)))
         return false;
 
+    PINData pinData;
+    if(lp.securityFlags)
+    {
+        if(!Read((char*)&pinData, sizeof(pinData)))
+            return false;
+    }
+
```
Phatcat: could maybe simplify into;

`if (lp.securityFlags && !Read((char*)&pinData, sizeof(pinData)))`
`    return false;`

?

src/realmd/AuthSocket.cpp
```diff
@@ -511,6 +513,13 @@ bool AuthSocket::_HandleLogonProof()
     if (!Read((char*)&lp, sizeof(sAuthLogonProof_C)))
         return false;
 
+    PINData pinData;
```
Phatcat: maybe add an empty newline between this and the line below?

Chaosvex: That'd be my preference but I just followed the example set above by "sAuthLogonProof_C lp;".

sql/base/realmd.sql
Chaosvex: It's an autogenerated file.

sql/base/realmd.sql
Chaosvex: It's an autogenerated file. I guess the exporter changed slightly since the table was last updated in 2013. ;)

sql/base/realmd.sql
Chaosvex: These details don't really matter. As said elsewhere, autogenerated file. I'll change the database name at least before it's final. ;)

src/realmd/AuthSocket.cpp
Phatcat: I'd say add an empty line to the example above as well, this is not one of those files being several thousand lines long so we don't have to feel bad using a bit of space to keeping it better visually organized in my opinion.

src/realmd/AuthSocket.cpp
Phatcat: no a fan of the idea of turning the two if statements into a single one? or not possible within this scope?

Chaosvex: It's possible although I think it decreases readability _slightly_.

Phatcat: Not if you slap a comment on there as well ;)

PR #192, opened 2016-09-06
Make spell miss visuals work

mityada:
For reasons unknown to me, Spell::WriteSpellGoTargets in mangos-classic writes all targets as if they were hit, even if they were not. Because of this, all spells visually hit the target in the game.

I've checked SMSG_SPELL_GO packet handler in 1.12.1 client. It's exactly the same in this part as in WotLK and even later patches. So I took Spell::WriteSpellGoTargets code from mangos-wotlk and now spell resists, absorbs, reflects, etc. are shown correctly.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3335,31 +3335,52 @@ void Spell::WriteAmmoToPacket(WorldPacket* data)
 
 void Spell::WriteSpellGoTargets(WorldPacket* data)
 {
+    size_t count_pos = data->wpos();
+    *data << uint8(0);                                      // placeholder
+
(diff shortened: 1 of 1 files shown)


Phatcat:
Not that it really matters that much but did you by chance take a look to see if the TBC handles this wrongfully as well? I'm guessing it either has the same implementation as classic or as wrath most likely.

mityada:
Yep, TBC has exactly the same code as WotLK. So now all three cores will be in sync in that part.

Phatcat:
so basically the classic branch is just behind on this one, seems like it's pretty straight forward. gonna merge this soon (giving at least few days for people to react with objections)

PR #193, opened 2016-09-07
Implementation of time-based one-time password & fixed PIN authentication

Chaosvex:
_This is continued from https://github.com/cmangos/mangos-classic/pull/191 now that TOTP has been implemented._

TOTP is now implemented on top of the PIN system and is backwards compatible with the 'old' IP locking system. 

An example can be viewed here: https://www.youtube.com/watch?v=VXAOxSBGFXg

**_Database changes:_**

1) I've 'extended' the realmd _'locked'_ column in the realm database to support additional flags. These are:

```
NONE           = 0x00,
IP_LOCK        = 0x01, // for backwards compatibility
FIXED_PIN      = 0x02,
TOTP           = 0x04,
ALWAYS_ENFORCE = 0x08
```

This allows a fair degree of flexibility with the upgraded lock system. Through setting flags, the user can:
- Continue to use the existing IP lock system, as is (value: 1)
- IP lock & 2FA to unlock the account (only prompt for PIN on IP change) (value: 3 or 5)
- IP lock & 2FA to unlock the account (_always_ prompt for PIN) (value: 11 or 13)
- No IP lock & 2FA (_always_ prompt for PIN) (value: 10 or 12)

The fixed PIN and TOTP flags are mutually exclusive.

2) The realmd database has a single new column, 'security'. This contains a string representing either a fixed PIN or a 16-character base32 encoded string, depending on whether the user is using a fixed PIN or TOTP.

_**Shared secret generation:**_

When generating a TOTP key, you should generate a random 10-byte array, base32 encode it and display the 16-character result to the user. The user will then add this key to their mobile authenticator to 'link' it with their account. If you want to learn more about the specifics, I'd recommend having a look at [https://tools.ietf.org/html/rfc6238](https://tools.ietf.org/html/rfc6238).

If you don't have a mobile, you can use an application called WinAuth, otherwise you can use Google Authenticator (or any other app implementing the same algorithm).

### DIFF
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -21,6 +21,8 @@
 */
 
 #include "Common.h"
+#include "Auth/Hmac.h"
+#include "Auth/base32.h"
 #include "Database/DatabaseEnv.h"
 #include "Config/Config.h"
 #include "Log.h"
@@ -30,7 +32,7 @@
 #include "Util.h"
 
 #include <openssl/md5.h>
-//#include "Util.h" -- for commented utf8ToUpperOnlyLatin
+#include <ctime>
 
 extern DatabaseType LoginDatabase;
 
@@ -142,6 +144,14 @@ struct PINData {
     uint8 hash[20];
 };
 
+enum LockFlag {
+    NONE           = 0x00,
+    IP_LOCK        = 0x01,
+    FIXED_PIN      = 0x02,
+    TOTP           = 0x04,
+    ALWAYS_ENFORCE = 0x08
+};
+
 typedef struct XFER_INIT
 {
     uint8 cmd;                                              // XFER_INITIATE
@@ -167,7 +177,8 @@ typedef struct AuthHandler
 
 /// Constructor - set the N and g values for SRP6
 AuthSocket::AuthSocket(boost::asio::io_service &service, std::function<void (Socket *)> closeHandler)
-    : Socket(service, closeHandler), _authed(false), _build(0), _accountSecurityLevel(SEC_PLAYER)
+    : Socket(service, closeHandler), _authed(false), promptPin(false), _build(0), gridSeed(0),
+      _accountSecurityLevel(SEC_PLAYER)
 {
     N.SetHexStr("894B645E89E1535BBDAD5B8B290650530801B18EBFBF5E8FAB3C82872A3E9BB7");
     g.SetDword(7);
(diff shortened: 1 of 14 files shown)


src/realmd/AuthSocket.cpp
```diff
@@ -373,19 +390,26 @@ bool AuthSocket::_HandleLogonChallenge()
...
             {
                 DEBUG_LOG("[AuthChallenge] Account '%s' is locked to IP - '%s'", _login.c_str(), (*result)[3].GetString());
                 DEBUG_LOG("[AuthChallenge] Player address is '%s'", m_address.c_str());
                 if (strcmp((*result)[3].GetString(), m_address.c_str()))
                 {
                     DEBUG_LOG("[AuthChallenge] Account IP differs");
-                    pkt << (uint8) WOW_FAIL_SUSPENDED;
+
+                    if (((lockFlags & TOTP) != TOTP && (lockFlags & FIXED_PIN) != FIXED_PIN))
+                        // account is IP locked and the player does not have 2FA enabled
```
Phatcat: I would move this comment one line up so that is comes before the if statement, slightly better style in my opinion, very minor thing, though.

src/realmd/AuthSocket.cpp
```diff
@@ -458,27 +482,29 @@ bool AuthSocket::_HandleLogonChallenge()
...
-                    pkt << uint8(securityFlags);            // security flags (0x0...0x04)
 
-                    if (securityFlags & 0x01)               // PIN input
-                    {
-                        pkt << uint32(0);
-                        pkt << uint64(0) << uint64(0);      // 16 bytes hash?
+                    // figure out whether we need to display the PIN grid
+                    promptPin = locked; // always prompt if the account is IP locked & 2FA is enabled
+
+                    if (!locked && ((lockFlags & ALWAYS_ENFORCE) == ALWAYS_ENFORCE)) {
```
Phatcat: { should have its own line

src/realmd/AuthSocket.cpp
```diff
@@ -600,8 +635,44 @@ bool AuthSocket::_HandleLogonProof()
...
+        }
+        else if ((lockFlags & TOTP) == TOTP)
+        {
+            for (int i = -2; i != 2; ++i)
+            {
+                auto pin = generateTotpPin(securityInfo, i);
+
+                if (pin == uint32(-1))
+                {
+                    break;
```
Phatcat: no reason for brackets around this, no?

src/realmd/AuthSocket.cpp
```diff
@@ -600,8 +635,44 @@ bool AuthSocket::_HandleLogonProof()
...
+                }
+                
+                if ((pinResult = VerifyPinData(pin, pinData)))
+                {
+                    break;
```
Phatcat: and this.

src/realmd/AuthSocket.cpp
```diff
@@ -987,4 +1058,106 @@ bool AuthSocket::_HandleXferAccept()
...
+    sha.Initialize();
+    sha.UpdateData(clientData.salt, sizeof(clientData.salt));
+    sha.UpdateData(hash.AsByteArray(), hash.GetNumBytes());
+    sha.Finalize();
+    hash.SetBinary(sha.GetDigest(), sha.GetLength());
+
+    return !memcmp(hash.AsByteArray(), clientData.hash, 20);
+}
+
+uint32 AuthSocket::generateTotpPin(const std::string& secret, int interval) {
```
Phatcat: { should have it's own line

src/realmd/AuthSocket.cpp
```diff
@@ -987,4 +1058,106 @@ bool AuthSocket::_HandleXferAccept()
...
+    std::vector<uint8> decoded_key((secret.size() + 7) / 8 * 5);
+    int key_size = base32_decode((const uint8_t*)secret.data(), decoded_key.data(), decoded_key.size());
+
+    if (key_size == -1) {
```
Phatcat: here as well

src/shared/Auth/base32.cpp
```diff
@@ -0,0 +1,95 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string.h>
+
+#include "base32.h"
+
+int base32_decode(const uint8_t *encoded, uint8_t *result, int bufSize) {
```
Phatcat: All thoughout this file { is gonna want its own new line. not gonna comment it all the way down there as it seems to be every use of { in this file.

Chaosvex:
I can add the other methods back if you wish but this change doesn't break compatibility with TBC or WotLK because those methods were never implemented.

namreeb:
That's fine.  If you want to submit another commit whose purpose is to remove unsupported features, feel free.  The point is that its bad style to solve two different problems with one commit/PR.

Lillecarl:
https://github.com/TrinityCore/TrinityCore/commit/ba22baebbd1394cc69366d7a19d879da43885430 This related in some way?

cyberium:
Does someone really need this? I mean its nice proof of concept but its a big amount of code to add for this feature that i myself doubt anyone will use.
I will go to close it if no answer.

Phatcat:
I would certainly appreciate this feature but it kind of needs some commands to go along with it for setting it in the mangos server console like we set up other account things in there using commands..
As for someone actually using it (besides me), I think that relies a bit about how accessible it is to use, which, falls back slightly on presence or absence of commands to go along with it..

Warlockbugs:
I believe it has full potential be more popular than AHBot or playerbot. This PR needs to be processed manually:

- First commit needs to go, it seems, as we agree to keep realmd in sync (Remove security methods not available in vanilla)
- Squash code changes
- Generate new DB revision
- Something else?

Inifield:
I don't know about others but I use it because I think it's a great system

PR #194, opened 2016-09-08
Fix delayed spell handling

mityada:
Spell::DoSpellHitOnUnit returns immediately for delayed spells, which have all the effects handled at launch, thus preventing some needed code from execution. The check remains since the time when the method had nothing but a loop with HandleEffets and served only for optimization. Now it can be safely removed.

One of consequences of this check is the possibility do damage the target with some spells (e.g. Shadow Bolt), after it becomes friendly or non-attackable. For example, it is possible to kill a player after duel ended.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -1098,7 +1098,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
 
 void Spell::DoSpellHitOnUnit(Unit* unit, uint32 effectMask, bool isReflected)
 {
-    if (!unit || !effectMask)
+    if (!unit)
         return;
 
     Unit* realCaster = GetAffectiveCaster();


mityada:
Here is a short demonstration.
Before: https://www.youtube.com/watch?v=y9rr4OoaTtk
After: https://www.youtube.com/watch?v=4abdsnDaDuc

For @Phatcat, regarding cmangos/issues#1015:
In the first video, spell visually hits the player. Player receives damage and dies.
In the second video, spell still visually hits the player (it's impossible to change _visual_ behaviour). But player doesn't receive any damage and doesn't die (it's possible to change _real_ behaviour).
So, it's all about matching visual and real. In first video, they do match, but real behaviour is obviously wrong. In the second video, they don't, but real behaviour is correct and we don't have another choice.

Phatcat:
Yeah. But notice how the spell hit visual is different in those 2 cases, in video 1 it's supressed (no visual impact) in video 2 it's not supressed (clear visual impact)

So the visual can't change depending on the flags/state of the unit you say?

dagochen:
The evade message is wrong or?

Phatcat:
I think it should be 'immune' or 'absorbed' but I'm not sure. I don't think it should be evade either, no, but there's no point guessing, let's find a source, preferably video, saying/showcasing what it really should say.

Phatcat:
> In the first video, spell visually hits the player. 

Not as it does in the second video, you gotta admit that.

Phatcat:
which removes the entire foundation this conclusion is build upon;

> (it's impossible to change visual behaviour).

dagochen:
It shouldn't show up anything at all (no immune, no absorb, no evade), same as it currently does on retail. There is still some similar problem for creatures dying while projectiles flying. Not sure if that was adressed before.

mityada:
@Phatcat, it's because death animation interrupts spell hit animation. If player didn't die, is would be exactly the same.

Phatcat:
That's funny, because that didn't happen back on retail; https://www.youtube.com/watch?v=4sj21TAmQ-M
(2:32) the client can handle it all fine, so no, it's us doing it wrong server side.

mityada:
Could you explain a bit more, what didn't happen? I see the frost bolt hitting the target. Target is not friendly, there are no immunities or other special cases.

Phatcat:
https://www.youtube.com/watch?v=Eh2JvonoUlg the spell visuals still happen during dying animation on retail. Sorry about that.

also https://www.youtube.com/watch?v=z45814mZTUo - 4:35 set speed to 0.25 First spell that connects ends the duel and makes his target friendly, but the second spell that hits still shows as if it did damage, it just doesn't. Make of that what you will. All I'm saying is you people are making a lot of conclusive statements based on assumptions.

dagochen:
still we fix an error here, so its ok ;) better than keeping players die after duels....^^

mityada:
Yes, visual effects disappearing on death is another problem. It affects not only missiles, but all spell visuals, including auras. I remember trying to solve it some time ago for some creatures that cast a spell on death.

As for evade message, this PR have nothing to do with it. This code is already in the core, the patch just makes sure it's get executed for all spells. It wouldn't be difficult to replace that message with whatever you'll find more correct.

Phatcat:
> Before: https://www.youtube.com/watch?v=y9rr4OoaTtk

You are doing the command very very close to the spell hitting, it's not really possible to tell if the visual from the spell was actually about to kick in before the death animation.

mityada:
Here is a video in which player doesn't die: https://www.youtube.com/watch?v=_KZT6fMJ64k

Phatcat:
That looks good. Looks like it's supposed to.

mityada:
Sorry, I think I wasn't clear enough. It is a new "before" video, in which death doesn't interfere with spell impact. In the rest it shows the same incorrect behaviour, as the player takes damage after duel end.

dagochen:
So to conclude this: 
Wrong animations are still a problem, but not tried to fix in this PR.
This PR wants to fix wrong behaviour where people take damage, while they should not for reasons like immune (paladin bubble) or being friendly or...whatever.

Phatcat:
Yeah, I am totally behind this change, I was just trying to get a clear understanding on where we were at.

> Now it can be safely removed.

Without any bad side effects? - Since you seem to know what the check was about once it had a purpose, could you make sure it doesn't break anything at all removing this? Sorry if you already have.

I will get to merging some of your PRs soon, I promise.

mityada:
At first, Spell::DoSpellHitOnUnit had only this code:
https://github.com/cmangos/mangos-svn/blob/e587b1e29fea2c7006ad9ab5b925830b8543a907/trunk/src/game/Spell.cpp#L686-L700
It's clearly seen that the check for effectMask has no purpose but optimization.

Since then more code was added. Now the check prevents it from executing for spells, all of whose effects satisfy IsEffectHandledOnDelayedSpellLaunch. For example, the method will always return for Shadow Bolt or Lightning Bolt, as they have only SPELL_EFFECT_SCHOOL_DAMAGE, but for Frostbolt it'll work fine, as it has SPELL_EFFECT_APPLY_AURA. I think this is also the reason this bug remained unnoticed for a long time.

PR #195, opened 2016-09-13
Spell log fixes

mityada:
SMSG_SPELLLOGEXECUTE now has the correct structure and data for all effects and targets.

SMSG_SPELLDISPELLOG and SMSG_DISPEL_FAILED now have the correct structures.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3391,76 +3391,60 @@ void Spell::WriteSpellGoTargets(WorldPacket* data)
 
 void Spell::SendLogExecute()
 {
-    Unit* target = m_targets.getUnitTarget() ? m_targets.getUnitTarget() : m_caster;
-
     WorldPacket data(SMSG_SPELLLOGEXECUTE, (8 + 4 + 4 + 4 + 4 + 8));
 
-    if (m_caster->GetTypeId() == TYPEID_PLAYER)
-        data << m_caster->GetPackGUID();
-    else
-        data << target->GetPackGUID();
-
+    data << m_caster->GetPackGUID();
     data << uint32(m_spellInfo->Id);
-    uint32 effectCount = 1;
-    data << uint32(effectCount);
-    for (uint32 i = 0; i < effectCount; ++i)
+
+    uint32 effectCount = 0;
+
+    for (auto effectInfo : m_executeLogInfo)
     {
-        data << uint32(m_spellInfo->Effect[EFFECT_INDEX_0]);
-        uint32 targetCount = 1;
-        data << uint32(targetCount);
-        for (uint32 j = 0; j < targetCount; ++j)
+        if (!effectInfo.empty())
+            effectCount++;
+    }
+
+    if (!effectCount)
+        return;
+
+    data << uint32(effectCount);
+
+    for (uint32 i = 0; i < MAX_EFFECT_INDEX; ++i)
+    {
+        if (m_executeLogInfo[i].empty())
+            continue;
+
+        data << uint32(m_spellInfo->Effect[i]);
+        data << uint32(m_executeLogInfo[i].size());
+
+        for (ExecuteLogInfo info : m_executeLogInfo[i])
         {
-            switch (m_spellInfo->Effect[EFFECT_INDEX_0])
+            switch (m_spellInfo->Effect[i])
             {
                 case SPELL_EFFECT_POWER_DRAIN:
-                    if (Unit* unit = m_targets.getUnitTarget())
-                        data << unit->GetObjectGuid();
-                    else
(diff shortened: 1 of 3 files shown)


Laizerox:
I will review this today. As for design I have own opinions how you could do it more clear.

I will also cross-check this with client because in TBC blizzard made a small shift for effects.

I will get back on PC.

mityada:
With the packets corrected, client shows the following messages for dispel:
> Kobold Geomancer's Frost Armor is removed.
You cast Purge on Kobold Geomancer.

The order doesn't look natural. I think SMSG_SPELLDISPELLOG and SMSG_DISPEL_FAILED should be moved out of the effect handler to, say, Spell::SendDispelLog, which then will be called after Spell::SendLogExecute.

Laizerox:
In all of the packets sniffed from retail spell log is always last.

Phatcat:
Looking natural is not always the same as being blizzlike, would just like to throw that out there ^^

mityada:
SMSG_SPELLINSTAKILLLOG had the wrong structure too. Fixed it.

I'd say too much wrong packets.

Laizerox:
https://gist.github.com/Laizerox/d32f4d57e2f8304d047bcf2713903e32

Here's the whole structure for anyone intrested.

Laizerox:
@mityada why don't we just use ByteBuffer and append to packet data in a loop? For example trinity does it that way and to preserve compatibility with TBC & WOTLK it would be quite mandatory. Because in those 2 client we're sending SpellLog for all summoned guids also. (For example 3x guardians)

Also the packets follow almost generic structure.

src/game/SpellEffects.cpp
```diff
@@ -4894,7 +4940,14 @@ void Spell::EffectDurabilityDamage(SpellEffectIndex eff_idx)
         return;
 
     if (Item* item = ((Player*)unitTarget)->GetItemByPos(INVENTORY_SLOT_BAG_0, slot))
+    {
         ((Player*)unitTarget)->DurabilityPointsLoss(item, damage);
+
+        ExecuteLogInfo info(unitTarget->GetObjectGuid());
+        info.durabilityDamage.itemEntry = item->GetProto()->ItemId;
+        info.durabilityDamage.unk = -1;
```
Laizerox: This is item slot

src/game/Spell.cpp
```diff
@@ -3391,106 +3391,78 @@ void Spell::WriteSpellGoTargets(WorldPacket* data)
...
+
+    data << uint32(effectCount);
+
+    for (uint32 i = 0; i < MAX_EFFECT_INDEX; ++i)
+    {
+        if (m_executeLogInfo[i].empty())
+            continue;
+
+        data << uint32(m_spellInfo->Effect[i]);
+        data << uint32(m_executeLogInfo[i].size());
```
Laizerox: This is actually target counter

mityada:
@Laizerox, do you mean using ByteBuffer instead of ExecuteLogInfo structure?

Anyway, why the current way is not compatible?
> Because in those 2 client we're sending SpellLog for all summoned guids also. (For example 3x guardians)

It's possible to have any number of ExecuteLogInfo per effect, so I don't see a problem here.

src/game/SpellEffects.cpp
mityada: How can you tell? That value is not actually used in the client.

Laizerox:
@mityada yes there's no problem.

I had my own version based on trinitys now I mixed some of your ideas and will post a PR for TBC soon so we can see where we are at.

src/game/SpellEffects.cpp
Laizerox: It is. Take a closer to look to a structure here: https://gist.github.com/Laizerox/d32f4d57e2f8304d047bcf2713903e32#file-spelllogexecute-cpp-L195 and here https://gist.github.com/Laizerox/d32f4d57e2f8304d047bcf2713903e32#file-spelllogexecute-cpp-L196 and here https://gist.github.com/Laizerox/d32f4d57e2f8304d047bcf2713903e32#file-spelllogexecute-cpp-L200

mityada: Of course I've seen that. I meant it's not used in any way that can tell it's an item slot.

mityada:
I looked at the TrinityCore's code for this and I find it very similar. The only difference is that they have a separate structure for each case, while I use union.

Laizerox:
@mityada I posted my solution written few weeks back and refactored with your idea here cmangos/mangos-tbc#126

Laizerox:
@mityada could you please separate the SpellLogExecute and other packets into 2 different PR's?

PR #196, opened 2016-09-14
Fix spell execute log

mityada:
SMSG_SPELLLOGEXECUTE now has the correct structure and data for all effects and targets.

See #195.


cyberium:
how i missed that one when i rewrited it?
i should learn to do search more :/
Facepalm...
Anyway replaced by https://github.com/cmangos/mangos-classic/pull/298

PR #197, opened 2016-09-14
Fix spell instant kill log

mityada:
SMSG_SPELLINSTAKILLLOG now has the correct structure.

See #195.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -252,8 +252,7 @@ void Spell::EffectInstaKill(SpellEffectIndex /*eff_idx*/)
 
     WorldObject* caster = GetCastingObject();               // we need the original casting object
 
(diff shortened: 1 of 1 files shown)


Phatcat:
This is for classic only? the current structure for tbc and wrath is correct already?

mityada:
Yep.

Phatcat:
Alright, so what difference to master does this one do exactly? Not entirely sure how to test this one out to see the effects of it.

Warlockbugs:
Perhaps, you can use Warlock's Demonic Sacrifice on pet. This is an instakill effect.

Phatcat:
you're right, didn't think of that one at all.

on master it simply goes

"you have slain <pet name>!"

On this patch it says 

"<pet name> is killed by <spell name>"
"you have slain <pet name>!"

:+1:

Merged.

PR #198, opened 2016-09-14
Fix spell dispel log

mityada:
SMSG_SPELLDISPELLOG and SMSG_DISPEL_FAILED now have the correct structures.

See #195.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -2371,17 +2371,14 @@ void Spell::EffectDispel(SpellEffectIndex eff_idx)
         if (!success_list.empty())
         {
             int32 count = success_list.size();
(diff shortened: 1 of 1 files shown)


Phatcat:
This is for classic only? the current structure for tbc and wrath is correct already?

mityada:
Yep.

Phatcat:
when a druid uses cure poison it isn't stated like you point to that it is with purge; on current master it just says that the poison fades, like this:

"Deadly Poison fades from you"

 with your patch it says it like this;

"Your Deadly Poison is removed"
"Deadly Poison fades from you"

Shouldn't it say who and what effect dispelled it?

mityada:
No, it shouldn't. Client uses one of AURADISPELOTHER and AURADISPELSELF strings. For enUS locale, they are "%s's %s is removed." and "Your %s is removed." respectively.

Phatcat:
Merged then.

PR #199, opened 2016-09-21
Allow 8 CPUs to be used for mmap extraction

lolagram:
This commit: https://github.com/cmangos/mangos-classic/commit/0ed121835e5b967f4976ec2d20531529be5243f0 allowed the MoveMapGen.sh script to use 8 CPUs, this just allows the ExtractResources.sh script to take advantage of that.


cyberium:
Thank you, merged another way

PR #202, opened 2016-10-03
Securize EffectMultipleValue usage for pet/summon level

cyberium:
@Phatcat please check this.


Phatcat:
I think it looks fine but indeed more research needs to be done into guardians especially and how their level is set, because loading it from the db like I was doing doesn't seem correct at all, the spell itself must hold some info about the level for at least guardians as well (especially since all summons from items have a static level independant from the caster) - exception is engineering trinkets which have a level based off your profession skill, where other engineering items follow the same pattern as other item summons (static level)

PR #203, opened 2016-10-15
The Stones That Bind Us

Jaypy:
* Added support for 'The Stones That Bind Us' (Quest ID: 2681)
Not handling "death immunity" properly (like Wrath training dummies). For now just nullifying damage if the stone isn't destroyed.

related to this DB issue: https://github.com/classicdb/database/issues/107

### DIFF
--- a/src/scriptdev2/scripts/eastern_kingdoms/blasted_lands.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/blasted_lands.cpp
@@ -27,6 +27,114 @@ EndContentData */
 
 #include "precompiled.h"
 
+enum
+{
+	NPC_RAZELIKH = 7668,
+	NPC_SEVINE = 7671,
+	NPC_ALLISTARJ = 7670,
+	NPC_GROL = 7669,
+
+	GOB_RAZELIKH = 141812,
+	GOB_SEVINE = 141859,
+	GOB_ALLISTARJ = 141858,
+	GOB_GROL = 141857,
+
+	SPELL_SELFSTUN = 9032
+};
+
+bool m_uiStoneDestroyed = false;
+bool m_uiStoneEnabled = false;
+GameObject* m_uiStoneGameObject;
+uint32 m_uiStunTimer = 0;
+
+struct npc_stone_servantAI : public ScriptedAI
+{
+	npc_stone_servantAI(Creature* pCreature) : ScriptedAI(pCreature) { Reset(); }
+
+	void Reset() override
+	{
+		m_uiStoneDestroyed	= false;
+		m_uiStoneEnabled	= false;
+		m_uiStunTimer = 0;
+	}
+
+	GameObject* FindStoneOfBinding()
+	{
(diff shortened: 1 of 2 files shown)


Phatcat:
You've indeed come to the righr place.

however we don't use tabs here, we use 4 whitespaces.

you can read more about our coding guidelines here: https://github.com/cmangos/issues/wiki/Coding-Standards

Warlockbugs:
Nice, these mobs were original target dummies before it was cool. 😄
I wonder if there are occasions of this behavior, in this case we can implement it directly into the core and as "Freeze health" functionality in the future.

cala:
Thank you. Assigned myself on this.

src/scriptdev2/scripts/eastern_kingdoms/blasted_lands.cpp
```diff
@@ -27,6 +27,114 @@ EndContentData */
...
+	void DamageTaken(Unit* /*pDoneBy*/, uint32& uiDamage) override
+	{
+		if (uiDamage < m_creature->GetHealth())
+			return;
+
+		if(m_creature->GetHealth() >= 1)
+			m_creature->SetHealth((uint32)1);
+
+		if (!m_uiStoneDestroyed)
+		{
```
Phatcat: these brackets could also be omitted since the contents is a one liner, just like you've done above.

src/scriptdev2/scripts/eastern_kingdoms/blasted_lands.cpp
```diff
@@ -27,6 +27,112 @@ EndContentData */
...
+        if (!m_uiStoneDestroyed)
+            uiDamage = 0;
+    }
+    
+    void UpdateAI(const uint32 uiDiff) override
+    {
+        if (!m_creature->SelectHostileTarget() || !m_creature->getVictim())
+            return;
+        
+        if ((m_creature->GetHealthPercent() < 5.0f))
```
Phatcat: an excess set of brackets here

cala:
I may propose an enhanced version of this soon, but this looks good already. :+1: 

> I wonder if there are more occasions of this behavior, in this case we can implement it directly into the core as "Freeze health" functionality in the future.

@Warlockbugs there is already such a functionality with EventAI where you can make a NPC immune to damage. There are also many spells that add this same behaviour for specific type of damages.

cala:
Actually, when researching for the various spells used in the event, I found that this could be done solely via a mix of DB and AI scripts.
In such cases, our policy is to use SD2 when DB or AI fixes are not enough. Thank you anyway for this PR that led me on the right path. :+1: 

Closed by: https://github.com/cmangos/classic-db/commit/e2e1fb24a953c04f2cfda8ea9fc32da49569bf86 and https://github.com/cmangos/classic-db/commit/f0ae5145811c3c555436d24be90556ab424361c9 and https://github.com/cmangos/classic-db/commit/adcf18487e207e777cd8dcffdc989316732e5093.

PR #204, opened 2016-10-16
Fix engineering transporters

Phatcat:
Basically involves spell ids 23441 through 23457 , each transporter is supposed to have 1 ultimate malfunction which can kill you, both of them are supposed to be able to polymorph and evil twin you.

Before, by casting spell 23446 expecting us to afterwards be in control of which malfunction, is not how that works, so whoever programmed it in was expecting to use 36902 to do the transport. I don't even know what that spell is, and neither does the classic core cause it's not in spell.dbc - if that person would have tested it he would have experienced getting the wrong destination and it also attempting to polymorph him (which it would fail and say interrupted, since that spell can not be casted while falling)

23450 wasn't used at all, this is the dummy effect for arrival, that could almost not be more clear.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -803,24 +803,20 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
 
                     return;
                 }
-                case 23448:                                 // Transporter Arrival - Ultrasafe Transporter: Gadgetzan - backfires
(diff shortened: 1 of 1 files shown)


Phatcat:
https://www.youtube.com/watch?v=fMkT4gHmeuY

Phatcat:
Gonna merge this in about... 12 days. should give more than enough time for people to counter this, should someone have any input.

Phatcat:
Merged.

PR #205, opened 2016-10-19
Fix memory leak in WorldSocket.cpp

Jose123456:
A dynamically-allocated WorldPacket instance was not always deleted

### DIFF
--- a/src/game/WorldSocket.cpp
+++ b/src/game/WorldSocket.cpp
@@ -183,23 +183,31 @@ bool WorldSocket::ProcessIncomingData()
 
     try
     {
+        bool l_success;
         switch (opcode)
         {
             case CMSG_AUTH_SESSION:
                 if (m_session)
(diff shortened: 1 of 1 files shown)


DanielOtto:
Why not use unique_ptr?

Jose123456:
That is too much magic for my taste, but here's an implementation using std::unique_ptr:
https://github.com/Jose123456/mangos-classic/commit/248d5c1135d2d7d62cb55254253145ac36fdfdff

Here's another alternate fix that I prefer. I replaced all the short-circuit return statements with break statements
https://github.com/Jose123456/mangos-classic/commit/8bbd627112608f7a3df009c53ad3b57bf76065fe

Jose123456:
It looks like folks here prefer the unique_ptr approach. Closing this pull request and opening a new one with the unique_ptr version.

PR #207, opened 2016-10-22
Fix memory leak in WorldSocket.cpp

Jose123456:
A dynamically-allocated WorldPacket instance was not always deleted.
Alternative fix using std::unique_ptr.


namreeb:
Nice find.  I will probably adjust your fix a bit but will plan to credit you for it.

namreeb:
What do you think about this? https://github.com/namreeb/mangos-wotlk/commit/e76c7cdb7d2c1b13dc4cbbfc7f341658baebf96c

Jose123456:
Well, you've already closed this 😄, but it looks good to me.

PR #215, opened 2016-11-06
Fix typo in VMap BIH generation

mityada:
Due to this typo some objects could be left out during BIH generation and thus never checked for intersection. This breaks height and LoS calculations in some areas.

The occurrence of this bug can be seen in "Md_Mushroomcave.wmo". It can be found in several locations, for example (-6140, -2960, 400), map 0. In some parts of the cave mobs will either fall through the ground or fly up into the sky when chasing the player. LoS is not obstructed there as well.

P. S. This typo presents in all MaNGOS cores.

### DIFF
--- a/src/game/vmap/BIH.cpp
+++ b/src/game/vmap/BIH.cpp
@@ -44,7 +44,8 @@ void BIH::subdivide(int left, int right, std::vector<uint32>& tempTree, buildDat
         return;
     }
     // calculate extents
-    int axis = -1, rightOrig;
+    int axis = -1;
+    int middle;
     float clipL;
     float clipR;
     float prevClip = G3D::fnan();
@@ -72,10 +73,12 @@ void BIH::subdivide(int left, int right, std::vector<uint32>& tempTree, buildDat
         // partition L/R subsets
         clipL = -G3D::inf();
         clipR = G3D::inf();
-        rightOrig = right; // save this for later
         float nodeL = G3D::inf();
         float nodeR = -G3D::inf();
-        for (int i = left; i <= right;)
+
+        middle = right;
+
+        for (int i = left; i <= middle;)
         {
             int obj = dat.indices[i];
             float minb = dat.primBound[obj].low()[axis];
@@ -92,9 +95,9 @@ void BIH::subdivide(int left, int right, std::vector<uint32>& tempTree, buildDat
             {
                 // move to the right most
                 int t = dat.indices[i];
-                dat.indices[i] = dat.indices[right];
-                dat.indices[right] = t;
-                --right;
+                dat.indices[i] = dat.indices[middle];
+                dat.indices[middle] = t;
+                --middle;
                 if (clipR > minb)
                     clipR = minb;
             }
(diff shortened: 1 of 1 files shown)


sidsukana:
Good job, my friend! :) Thank you!

killerwife:
@mityada You seem to be familiar with MMAPs, do you have an idea how to make transports work for example?

mityada:
@killerwife, are you talking about pathfinding on transports?

killerwife:
yes

mityada:
I think I do. Extra navmeshes for transport models should be generated and then used for any movement within the transport. Making path from/to the transport can be more tricky.

killerwife:
And client sends same coords even if transport is moving, which means transport is somewhere in the corner in the navmesh. But I think even if you implemented the transport that would be helpful, the transition would be something I have an idea or two about, but not the generation :/ (since transition is just a corner case in core but navmesh is in extractor)

zamalaev:
I do not understand what the logic is, if rightOrig = right;

mityada:
@zamalaev, `right` can then be changed [here](https://github.com/mityada/mangos-classic/blob/f68416c9a13da4d116ce38ed1d6b3173f4021976/src/game/vmap/BIH.cpp#L93-L99).

Anteara:
Can you give me an example of what this aims to fix (like an example in-game that happens without this, and is not present with it?)

I am looking to merge this in to the project I volunteer at. We're considering transitioning to cmangos entirely, but the process will be tough.

For now, I will look at some commits and differences (e.g. ACE vs BOOST) and manually add some of these... I will include credits of course in the commits

Anyway, back to my question? Thanks!

mityada:
@Anteara, this solves height and LoS issues in _some_ parts of _some_ world models (e.g. buildings, caves, dungeons, etc.). I'm not sure about Karazhan as I don't have TBC data right now. [Here](http://pastebin.com/kHPKcLFH) is the list of models for Classic, where you _may_ encounter that issue.

cyberium:
Thank you @mityada i got time to test this. I cannot confirm the fix :/ The creature at that place still dont chase me.
I may forget something btw.
Recompiled all extractings tools, rebuild vmaps/mmaps and recompiled cmangos core.


However as it seem to work for some. Could you go to
> .go 1518.3234 -4710.5771 11.2992
Normal
Abnormal

As you can see there is something wrong at that place for me. Could you confirm it?
Do you have any idea where is the problem?
Is it fixed for you by your patch?

Phatcat:
I got another spot as well referenced in one of my joana's issues; https://github.com/cmangos/issues/issues/1054 ()

> [34:22](https://youtu.be/SwLjJH8LXfc?t=34m19s)
Around this area there is a bugged spot on cmangos classic (mobs will constantly keeps going in combat with you just to go evading right after) - seems to be related to pathing inside that cave on some spots as moving away from that spot allowed the mobs to path to me (I have mmaps enabled)

bad

on cmangos you can stand here, meleeing the mob right there, without it being able to path to you.

Phatcat:
Also these pets here we see here:

bad2

outside, on top of Skull rock, should rather be inside Skull rock. I can only imagine bad pathing got them here to begin with.

ryancheung:
I've tested this version(with rebulding maps, vmaps and mmaps), which is working. The path generated with this patch is normal in the caves. @cyberium 

```diff
--- a/src/game/vmap/BIH.cpp
+++ b/src/game/vmap/BIH.cpp
@@ -155,8 +155,8 @@ void BIH::subdivide(int left, int right, std::vector<uint32>& tempTree, buildDat
             if (prevAxis == axis && G3D::fuzzyEq(prevSplit, split))
             {
                 // we are stuck here - create a leaf
-                stats.updateLeaf(depth, right - left + 1);
-                createNode(tempTree, nodeIndex, left, right);
+                stats.updateLeaf(depth, rightOrig - left + 1);
+                createNode(tempTree, nodeIndex, left, rightOrig);
                 return;
             }
             right = rightOrig;
```

killerwife:
WMOs have a different problem as well afaik. But that is related to wrong Z being used. (causes players to behave weirdly when blinking in undercity for example)

PR #217, opened 2016-11-10
UpdateAutoRepeatSpells

Phatcat:
Interrupt all auto-casting on movement (hunters and casters cannot run-and-gun)
Properly switch target for auto
repeat spells on target switch for units

Ref: https://github.com/cmangos/issues/issues/489

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -19103,3 +19103,65 @@ void Player::ForceHealAndPowerUpdateInZone()
         }
     }
 }
+
+void Player::_UpdateAutoRepeatSpell()
+{
+    // check "realtime" interrupts
+    if (IsMoving() || IsNonMeleeSpellCasted(false, false, true))
+    {
+        // cancel wand shoot
+        if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
+            InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
+
+        m_AutoRepeatFirstCast = true;
+        return;
+    }
+
+    ObjectGuid targetGuid = GetSelectionGuid();
+
+    if (!targetGuid.IsEmpty() && targetGuid.IsUnit())
+    {
+        Unit* unitTarget = GetMap()->GetUnit(targetGuid);
+        // Only do real time target updating if target doesn't remain the same
+        if (unitTarget != m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_targets.getUnitTarget())
+        {
+            m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_targets.setUnitTarget(unitTarget);
+
+            // Interrupt immediately on if new target fails the castcheck, and don't allow shooting of yourself.
+            if (unitTarget == this || m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->CheckCast(true) != SPELL_CAST_OK)
+                InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
+
+            m_AutoRepeatFirstCast = true;
+            return;
+        }
+    }
+
+    // apply delay
+    if (m_AutoRepeatFirstCast && getAttackTimer(RANGED_ATTACK) < 500)
+    {
+        setAttackTimer(RANGED_ATTACK, 500);
+        m_AutoRepeatFirstCast = false;
+    }
+
+    // castroutine
+    if (isAttackReady(RANGED_ATTACK))
+    {
+        // Check if able to cast
+        if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->CheckCast(true) != SPELL_CAST_OK)
+        {
+            InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
+            return;
+        }
+
+        // Always stand before shooting, at least players
+        if (getStandState() != UNIT_STAND_STATE_STAND)
+            SetStandState(UNIT_STAND_STATE_STAND);
+
+        // we want to shoot
+        Spell* spell = new Spell(this, m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo, true);
+        spell->SpellStart(&(m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_targets));
+
+        // all went good, reset attack
+        resetAttackTimer(RANGED_ATTACK);
+    }
+}
--- a/src/game/Entities/Player.h
+++ b/src/game/Entities/Player.h
@@ -2082,6 +2082,8 @@ class Player : public Unit
         virtual CombatData* GetCombatData() override { if (m_charmInfo && m_charmInfo->GetCombatData()) return m_charmInfo->GetCombatData(); return m_combatData; }
         void ForceHealAndPowerUpdateInZone();
 
+        virtual void _UpdateAutoRepeatSpell() override;
+
 #ifdef BUILD_PLAYERBOT
         // A Player can either have a playerbotMgr (to manage its bots), or have playerbotAI (if it is a bot), or
         // neither. Code that enables bots must create the playerbotMgr and set it using SetPlayerbotMgr.
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -3556,7 +3556,7 @@ void Unit::_UpdateSpells(uint32 time)
 void Unit::_UpdateAutoRepeatSpell()
 {
     // check "realtime" interrupts
-    if ((GetTypeId() == TYPEID_PLAYER && ((Player*)this)->IsMoving()) || IsNonMeleeSpellCasted(false, false, true))
+    if (IsNonMeleeSpellCasted(false, false, true))
     {
         // cancel wand shoot
         if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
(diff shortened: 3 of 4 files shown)


Phatcat:
+ properly delay auto-repeat cast interrupts to next cast, they do not happen instantly, but on next auto repeat cast.
this can be seen throughout joana's leveling videos; https://youtu.be/NrSdjsrhX0s?t=1h18m25s

Phatcat:
Can you read? Then this change is for you!

Are you tired of having such a crooked aim that even though you start aiming at a new enemy your shots just keep sticking to your old target?

And how about stepping forward and trying to put your weapon away to simply find that you can't stop shooting?

Fear not, for there is a solution to your prayers! It requires very little of you, well, __practically__ nothing actually!

But such as in life, there's no such thing as a free dinner, I don't want your money, though, I want a bit of your time. Your time and cold-heartedly honest opinion.

I want **YOU** to playtest this patch! That's right! You, `<kindly insert your own name>`!

You have been personally hand-picked in this once-in-a-lifetime opportunity to be a part of something beatiful, something that very well could go down in history as one of the greatest patches ever created* - nay, ever to __be__ created. Period.

All that is asked of you, is that you equip yourself with your most coldheartedly honest opinions, oh, and that you take on the hunting profession, you must let it become you for a time, in that time you must practise; you must practise and pick up skills you never would have imagined yourself able of performing.

It is then that you utilize those skills, it is now that you, much like Rambo,, will put those very same skills to the test; the ultimate test: shooting at one target and then switching to another target! 

So are you ready to become the new Rambo?** I think that you are!

disclaimer:
`*` the writer of this patch and comment does not nessesarily hold the opinion that this patch is anything out of the ordinary for any other purpose than writing this comment **neither does he garantee any kind of rambo'ness will apply to you for testing this

Rushor:
@Phatcat you watched too much usa-advertisments xDD

Phatcat:
No one plays a hunter around here?

This should only take like 5 minutes to try out, there's only 2 things to test, which can be tested on basically any mobs. (switching targets while having auto shoot on, and moving to cancel auto shooting)

HanFox:
This was bugging me. Might have been something I'd have looked in to if I didn't hack code so badly (re Imp Mend Pet :P).

Seems to work as expected in regards to cancelling Auto Shot on movement and switching targets.

What does "repeat spells on target switch for units" mean exactly?

dagochen:
several more things to check:
- running in and out of maxrange
- getting interrupted by crowd controlling effects
- start autoshoot and start eating/drinking at same time
- using rangespell like arcaneshot without mana
- switching ammo while autoshot is active
- start autoshoot and mount up to become mounted archer

Phatcat:
> What does "repeat spells on target switch for units" mean exactly?

It means I'm bad with words. what I tried to say was:

~~Properly switch target for auto
repeat spells on target switch for units~~

Properly start shooting at the new target with autoRepeatspells on target switch to another unit, (if you just switch to terrain, you will still keep shooting at your old target (this is classic behaviour at least, dunno about tbc and later))


thank you dago, you are right all those things should get checked for behaviour as well.

first one should be handled somewhat, at least.
running in closer than min range or running outside of max range should make the auto shooting cancel by itself on next checkcast, which is always done before shooting.

The rest of those right there, I'll see what I can do about some of that.

dagochen:
>>running in closer than min range or running outside of max range should make the auto shooting cancel by itself on next checkcast, which is always done before shooting.

and running back into correct range should start next autoshot without the need of clicking it again. spell spell should still be active.

all the things i listed before were problems or bugs we had on our server ;) I'd like to test this PR, but...how do i do that?

Phatcat:
Actually, that depends if it's the unit that ran in and out of range, if you, the hunter yourself, moved at all, then the autoshooting will be cancelled. There is no running and gunning in wow vanilla.

there are a few ways you can test other people's prs. I keep forgetting the command to pull in work that has already been PR'd, since that can be done easy, so I do it another way which takes a bit more time, but I think it's worth the extra trouble.

you open up your cmangos classic repository in git. If you are on the github windows application, then you can get there by navigating to the repo inside the application, then clicking the gear in the upper right corner and choose 'open in git shell'

then you add my repository as a remote by entering the command

`git remote add phatcat https://github.com/Phatcat/mangos-classic.git`

then you fetch it

`git fetch phatcat`

then inside the github application you can now navigate to all of my branches on my repo, and you can find the branch with this commit on it.

If you want to pull in changes I make later to test out other Prs I do and stuff, you can now pull them in simply by doing another fetch. If you however no longer want access you can remove it again with

`git remote remove phatcat`

If you have navigated to any of the branches locally you may however have to delete them locally as well, and then you may have to repeat the remove from remote command again.

hope it made sense =)

Phatcat:
Hey @dagochen @HanFox

I fixed all of the stuff mentioned, all besides this one

> switching ammo while autoshot is active

I have fixed up the commit and force-pushed to this branch so if anyone wants to try out the new commit that would be just wonderful.

Phatcat:
Completely refactored the function to get rid of some bugs I discovered while testing.

It wouldn't always change target for autorepeat spell when the caster changed his target. This can still happen if you change target immediately after an arrow has been fired. I need to find out what is really going on, and where. I think it has to do with the fact that it forces the target of the autorepeatspell to be the target of the caster whenever it fires, it should only do that if the caster has no target (like getting struck), not if the caster has another target. It seems to be doing that currently, though, and that seems to be messing things up a bit.

It would allow you to shoot yourself by first targeting an enemy, then by timing the targeting of yourself right you would get to shoot yourself, for some reason checkcast doesn't return with cast fail in that case. That should no longer be possible either.

HanFox:
Is there a good, **easy** to access place to test crowd control?

I can change ammo whilst shooting.

Everything else _seems_ to be fixed.

HanFox:
If you change to a target that's behind you you keep shooting at the original target.

Also, (I'm not sure if this is Classic behaviour or not) if you try to cast Auto Shoot on something behind you you get an interrupted message instead of "Target needs to be in front of you".

Phatcat:
> you get an interrupted message instead of "Target needs to be in front of you"

Currently we don't have a system in place to differantiate between the messages sent when interrupting autorepeatspells. I'll look into that. Requires a bit more work, though. We will most likely have to handle it much like we handle error messages for hunter pet related stuff.

Dunno if I can do much about the ammo thing. I work almost exclusively with unit stuff, dunno much about the item system and stuffs.

> If you change to a target that's behind you you keep shooting at the original target.

Yes, the whole getting the right target seems a bit tricky. I'll see if I can do anything to improve that, will have to look deeper into that as it seems to have some weird behaviour currently, targeting and autorepeatspells, that is.

Phatcat:
@HanFox 
Yeah, fixed the target thing, it had to do with not getting the right guid, fetching selectionguid instead solved it. It seems to always want to shoot the target the player selects now, also adjusted all of the commit, it should function slightly different, but a lot better now.

Havn't looked into the error messages yet. I bet @killerwife could fix it up real good, he fixed the error messages for hunter pets real nice (it's on a PR by cyberium on both mangos-classic and tbc)

I'll try looking into the ammo thing. I never played a hunter much so I need a bit of info (and preferably some rewsearch/proof to back it up) so, what should happen, should the attack be reset or completely interrupted? - also is it the same behaviour as changing ranged weapon, or is changing ranged weapon supposed to behave differently than changing ammo, and if so, how they should behave differently?

HanFox:
So I've noticed something a bit odd with this, probably down to it waiting to check on next cast?

1) Start shooting something and move; Auto - Shot keeps flashing; stop moving it takes until the next cast check to stop. Really it needs to check if you start moving and then interrupt.

2) Do as above, but click/press Auto Shot as soon as you stop moving. It'll stop casting with an Interrupt, but then when you start shooting again it usually interrupts straight away again.

I think 2 is related to some odd behaviour I have when sometimes after the first attack the Auto Shot gets interrupted. I'm neither moving nor is the target moving out of range or getting too close.

Phatcat:
> Really it needs to check if you start moving and then interrupt.

Not true, read the comment i put in the code, and look at this; https://www.youtube.com/watch?v=NrSdjsrhX0s&feature=youtu.be&t=1h18m25s - probably it needs to be adjusted a bit to stop instantly when you stop moving. But clearly it keeps flashing while joana is moving.

> but then when you start shooting again it usually interrupts straight away again.

Yeah, I forgot to take a few things into account, should be taken care of now.

dagochen:
https://youtu.be/NrSdjsrhX0s?t=1h11m4s

the flashing turns off for what reason?

Phatcat:
@dagochen Should be taken care of now. It should interrupt on next check, I see that now.

> some odd behaviour I have when sometimes after the first attack the Auto Shot gets interrupted.

I found the issue, I wasn't setting both of the new bools to false on init Taken care of.

dagochen:
When does the next check happen? In this video the behaviour looks different to me compared with the previous link I did post. Did blizzard maybe change something ...?

https://www.youtube.com/watch?v=OrmljErZt-A&t=47s

Phatcat:
The check happens whenever you're about to shoot, so every tick of the ranged attack timer based on your weapon and quiver and all that jazz, or if you change target, then a check is done as well.
Whenever you use a ranged ability the interrupt is cancelled however, and in the video you linked he does turn around a lot and use ranged abilities.

The thing is that it looks like some of those ranged attacks should reset the attack timer as well, we do not do this on cmangos currently, we do however set the time for the next attack to be at least 0.5 seconds later. (which I do think is correct for some ranged attacks, just not all... we need research on this, though...)

PR #218, opened 2016-11-10
Fix CMake warnings about missing dependencies

Jose123456:
The revision.h file is gone since 8c94d21


Jose123456:
Revision changes were reverted.

PR #219, opened 2016-11-17
Some Naxx fixes and changes.

Qingle:
Added some changes to Naxx bosses and texts.


Qingle:
Can anyone work out why the hell the Gothik update is causing errors?

src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_razuvious.cpp
```diff
@@ -161,3 +159,5 @@ void AddSC_boss_razuvious()
     pNewScript->GetAI = &GetAI_boss_razuvious;
     pNewScript->RegisterSelf();
 }
+Contact GitHub API Training Shop Blog About
```
cyberium: dont know where those line come from but its pretty obvious its wrong and cause build fail

Qingle: Oops, that's probably me C&P from my TBC fork.

Qingle: Still seems to be a problem with Gothik, any ideas?

killerwife:
👍 keep it up, and one day we might actually have a working vanilla naxx :)

Qingle:
I'm going to try sort out the DBScriping too, I'm just a newbie to all of this so I'm learning as I go along.

Jose123456:
Your problem is here: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_anubrekhan.cpp#L139

You removed `SAY_GREET` for Anub'Rekhan here: https://github.com/cmangos/mangos-classic/pull/219/files#diff-796b2b1ce2f08c7e4a897e5f421a2a87L29

Qingle:
woops

src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_gluth.cpp
```diff
@@ -150,7 +150,9 @@ struct boss_gluthAI : public ScriptedAI
 
                 // Devour a Zombie
                 if (pZombie->IsWithinDistInMap(m_creature, 15.0f))
+                    m_creature->SetFacingToObject(pZombie);
                     m_creature->DealDamage(pZombie, pZombie->GetHealth(), nullptr, DIRECT_DAMAGE, SPELL_SCHOOL_MASK_NORMAL, nullptr, false);
+                    m_creature->SetFacingToObject(m_creature->getVictim());
```
xfurry: The SetFacingToObject() is not required. It won't have any impact on the behavior.
Since the boss it's in combat, it will update the facing to the victim anyway.

Qingle: Alright, just followed the notes at the top, if this has been sorting corewise I'll delete it.

Qingle: I'll do the TBC version when I've actually fixed this one.

cala:
Assigning myself on this. I will review it. @Qingle once this is merged to Classic, I'll do the port to TBC: the code is the same between the two cores, so let's use our backport/upport tools for that.

Qingle:
I seem to have pulled up a lot of errors atm, so I'm working through them

cala:
Please, let me know once you're done so I can look at the changes you've made. :)

Qingle:
Done.

Qingle:
@cala if you could give me an update of how anub'rehkan's intro speech works atm that would be great, because I have no idea how it functions right now.

cala:
Currently, the intro speech is triggered on Line of Sight: as soon as a player is seen by Anub'Rehkan, a SD2 function designed to handle timed speeches is called.
The structure is described here: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_anubrekhan.cpp#L56 (Text ID, NPC saying the text and timer before it says it)
The call is made here: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_anubrekhan.cpp#L135

cala:
Basically, what you need to do is:
- script the entrance door gameobject to send a EventAI to the NPC ANub'Rehkan on use
- change in Anub'Rehkan script the part triggering the dialogue on LoS to a trigger on EventAI receive

Examples of how to throw/receive EventAI
Throw: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/kalimdor/desolace.cpp#L641
Receive: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/kalimdor/desolace.cpp#L471

Qingle:
So basically it would be along lines of 

if (GO_ARAC_ANUB_DOOR == GO_STATE_ACTIVE)

?

killerwife:
void onGoUse(random params i dont remebers)
{
if(Anubrehkan* anub = getobject(rekhan))
{
anub->SendAiEvent(more random parameters, eventTypeA);
}
}

and then in anubrehkan

void ReceiveAiEvent(random params)
{
if(eventtype == eventTypeA)
trigger dialogue;
}

Qingle:
Ah ok. Thank you, I'll get it sorted today.

Qingle:
Something like this?

Qingle:
Nvm, working out which event it actually is

Qingle:
Ok I still have no idea how to do this.

Qingle:
I'll have to do some tests on this, should be good to go though

killerwife:
I suggest you squash the commits, then you need to wait for @xfurry or @cala approval/ you will need to explain your changes.

Qingle:
@killerwife I have no idea how to squash commits, help please. Github noob.

killerwife:
combination of git reset HEAD^ (to previous commit, number of carets is number of commits reverted) and then you git commit at once, meaning all commits are self contained instead of 5 commits for same issue. If your repo is not linear (same issue doesnt follow in commits), then you will need to do git cherry-pick git-hash in order of which commit fixes what, and do the reset/commit combo on those. Sadly you will need to play around with it a bit to get the hang of it.

Qingle:
So I'll have to do some Git command line work

Phatcat:
git rebase -i head~N (where N is the number of commits to work with.) is the command which gives you the most control, but is the least direct (and most time-consuming)

if you do that command a txt document should open giving you a bit of info and a list of the N number of most recent commits to the branch.
then for those commits to squash change the 'pick' to either f or s (for fix or squash), save the document and close it. Should be done. If it throws an error instead simply do git rebase --abort and ´no changes will be applied at all. This normally can't happen simply doing squashes and fixups, though.

either that or as killerwife said, do git reset --soft head~N (again, where N is the number of commits to work with)
Doing the latter you'll have all your changes, but unstaged, uncommited. So you'll have to commit them again before updating your remote.

when you've done either of the two, you do a git push -f

you have to force push because you've already published changes which you'll now be removing again.

Blumfield:
Doesn't github automaticly squash commit now ?

Phatcat:
No, that would also be a horrible nightmare, taking away squashing control from the user.

Qingle:
Is there any way to squash using the github website?

Phatcat:
Not that I know of.

cyberium:
@Phatcat give you enough tips to use it. Do the effort to learn it. Git is a power tools that can help you in many other project.
I found a video here -> https://www.youtube.com/watch?v=qh9KtjfjzCU
Well hard to listen but its basicaly what you have to do.

Qingle:
Alright I'll have to sort it when I can

cala:
Again, thank you very much @Qingle for all those changes and the research that comes with. I reviewed the scripts and below are my remarks:

* Patchwerk: OK
* Loatheb: OK
* Gothik: OK
* Gluth: OK
* Faerlina: OK
* Horsemen: I'm 99% sure that the marks timers should remain the same even after hitting the 100 marks, so the 4 related checks you added should be removed, the rest is OK
* Grobbulus:
  - I think you used the wrong inequality check in `if (m_creature->GetHealthPercent() < 30.0f)` because currently you are saying that the injection should be slower *after* Grobbulus reached the 30% HP mark
  - the timers you are using are probably those from WotLK, I did a quick timers benchmark on Classic videos of the encounter and it was 6 - 13 sec range above 30% and 3 - 7 sec below 30% HP
* Kel'Thuzad: looks very good but some timers need to be adjusted
  - Chains of Kel'Thuzad: every 60 - 120 sec (described in wiki, phase 2 strategy)
  - Frost Blast 30 - 40 sec (the lower value comes from wiki, where you probably find it, the higher is a estimation as so tough abilities were almost never on a flat timer)
  - Shadow Fissure 10 - 12 sec (same reasoning as for Frost Blast)
* Instructor Razuvious:
  - the logic of moving some of the yell to a triumphant shout check is ok, however most of the yells for the whole encounter are not as described in the wiki (but that was already that way before your patch)
  - @xfurry will confirm this, but I'm almost sure that `void SpellHit(Unit* /*pCaster*/, const SpellEntry* pSpellEntry) override` will never be called because Razuvious can not be a target of his own spell (Disruptive Shout) and to my knowledge, SpellHit will only be called if the holder is the target, not the caster. This means that the triumphant shout yells will never be performed. Unfortunately, we can not add a dummy effect for the spell because its DBC description does not support it. I don't know how to fix this for now
* Anub'Rehkan: I don't think this would work. You have to script the GO entry itself to make it use a script OnUse that would: get Creature* Anub'Rehkan from instance storage, dynamically cast it as a Anub'Rehkan AI (you've already done that) and then send the event AI to it. You can find an example of how to script a GO [here](https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/eastern_kingdoms/blackrock_depths/blackrock_depths.cpp#L1798)

cala:
If you are ok, I can merge the OK scripts soon.

If you want to fix the remaining one, please do so, or I can fix them myself if you prefer.

There is only the Razuvious part that I'm still unsure of...

cala:
And thank you for taking care of my beloved (Classic) Naxxramas! ❤️

xfurry:
> I'm almost sure that void SpellHit(Unit* /*pCaster*/, const SpellEntry* pSpellEntry) override will never be called because Razuvious can not be a target of his own spell (Disruptive Shout) and to my knowledge, SpellHit will only be called if the holder is the target, not the caster. This means that the triumphant shout yells will never be performed. Unfortunately, we can not add a dummy effect for the spell because its DBC description does not support it. I don't know how to fix this for now

I can see a comment that says ```// For Triumphant Shout, should be used anytime Disrupting Shout hits a target```
So why not use SpellHitTarget() for this? Instead of SpellHit() which is used only when target is self.
Also, please be careful on this, as if the boss hits more target, it will spam with shouts.

killerwife:
Sounds like a job for scripting a spell instead of scripting SD2 if its supposed to happen once.

Qingle:
It is meant to happen once, I have no idea why I used SpellHit().
I also have no idea how or where to script a spell.

killerwife:
Download QSW. find the effect on which the roar should trigger, and add it in spelleffects.cpp or spellauras.cpp on proper place.

Qingle:
Update from my side:
- I have no idea how the GO would work here, feel free to script it if you want.
- Don't have access to QSW atm

cala:
Thank you for the update. I'll push them soon and script the Anub'Rehkan's door part.

For Razuvious, I already looked at the spell DBC and its only effect is SPELL_EFFECT_POWER_BURN which basically does:
```
If target has mana then
   remove mana_amount
   do damage=f(mana_amount)
end
```
No way to script the spell itself. I looked for a parent spell that would trigger spell 29107 but found none.
So, I'm wondering how we could inform the boss that its spell hit as we can't script the targets because they are players... And that's sounds very complicated for only having the boss perform an emote...

killerwife:
Sounds like something I encountered when making my spell scripting implementation. Basically the script belongs to spell_finish_phase, so its only run once on m_caster, but without my system its extremely ugly.

cala:
Unless there is a way to inform a caster that its spell hit (and from what you said, there is none currently), I think I will make the NPC emote on every spell cast and add a TODO in the code for later update if that sounds good to everyone.

src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_anubrekhan.cpp
```diff
@@ -132,15 +128,24 @@ struct boss_anubrekhanAI : public ScriptedAI
         DoCastSpellIfCan(m_creature, SPELL_DOUBLE_ATTACK, CAST_TRIGGERED | CAST_AURA_NOT_PRESENT);
     }
 
-    void MoveInLineOfSight(Unit* pWho) override
+    void OnAnubDoorGo (Player* pPlayer, Creature* pCreature)
     {
-        if (!m_bHasTaunted && pWho->GetTypeId() == TYPEID_PLAYER && m_creature->IsWithinDistInMap(pWho, 110.0f) && m_creature->IsWithinLOSInMap(pWho))
+        if (GO_ARAC_ANUB_DOOR == GO_STATE_ACTIVE)
```
xfurry: This is very ugly... And also doesn't work.
If you need to make the creature to say the intro when the door changes the state, you will need to do this via the instance script.

When the door state is changed here: https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/eastern_kingdoms/naxxramas/instance_naxxramas.cpp#L288
then you should make send the AI event to the boss, using itself as the sender and receiver.

Also, I'm not sure why you don't like the way I implemented this, with the IsWithinDistInMap(). Technically the boss should start yelling when the player enters the room. The door should be already open, so I don't understand what is your logic behind it.

PS. the line below ```if (boss_anubrekhanAI* pAnub = dynamic_cast<boss_anubrekhanAI*>(pCreature->AI()))``` should be replaced simply with ```SendAIEvent(id, invoker, m_creature)```

killerwife:
@cala if it will work sounds good to me. I can deal with it later when I export my latest spellscripting draft. Kinda have been dealing with one apocalypse after another so couldnt find time for it. (namely WDB team running into limitations)

xfurry:
@killerwife let's deal with spell scripts separately. Your ideas is good, but let's not mix the script with spells. :smile:

src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_anubrekhan.cpp
Qingle: The reasoning behind it is that we can just hook it to an event instead of having to do distance/LoS checks. I thought it would also look cleaner and/or make the event smoother.

xfurry: That is fine, but the thing is that the door is already opened before the encounter. So there won'be any event triggering this.

killerwife:
its a discussion for another time and another thread, aye

cala:
> Also, I'm not sure why you don't like the way I implemented this, with the IsWithinDistInMap(). Technically the boss should start yelling when the player enters the room. The door should be already open, so I don't understand what is your logic behind it.

@xfurry possibly because currently, the speech is said when the players are already near the middle of the room (tested a few minutes ago, the LoS check may depend on server settings).  So the idea was to bind the speech to the door because this is at this moment that it should be said. My idea is to trigger the speech on GoUse.

cala:
> That is fine, but the thing is that the door is already opened before the encounter. So there won'be any event triggering this.

I missed that line, sorry. The door is not open (at least it was not in Classic and TBC). So, it is possible to script the GO.

cala:
Merged in core from revision https://github.com/cmangos/mangos-classic/commit/4c7b75326d80fd891650ff783f447f8e0b73b3b5 to https://github.com/cmangos/mangos-classic/commit/b818bf3a5bbdd6a4db2b2efae996038f99d0ae43
Thanks to everyone.

dagochen:
Grobbulus Poison Cloud not working :( Any ideas on that?

Rushor:
https://github.com/Qingle/mangos-classic/blob/0b4c2a993583db082c18a2adbb0bbf64ebcd0716/src/scriptdev2/scripts/eastern_kingdoms/naxxramas/boss_grobbulus.cpp#L20

dagochen:
thx :(

PR #220, opened 2016-11-18
Lazypeonquest - Fix peon behaviour

BareRose:
Peons will chop at the trees, not their piles, and will sleep at their spawns.
It is no longer possible to hit awake peons.
Peons are now playing sounds as they should & using the correct classic text.

Meant to go together with [my PR to cmangos/classic-db](https://github.com/cmangos/classic-db/pull/18).

Posting so @Phatcat can do his testing, probably shouldn't be merged until he's done double-checking.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4690,6 +4690,14 @@ SpellCastResult Spell::CheckCast(bool strict)
         {
             case SPELL_EFFECT_DUMMY:
             {
+                //By Spell ID
+                if (m_spellInfo->Id == 19938)               // Awaken Lazy Peon
+                {
+                    Unit* target = m_targets.getUnitTarget();
+                    if (!target || !target->HasAura(17743 /*SPELL_PEON_SLEEP*/) || target->GetEntry() != 10556 /*NPC_SLEEPING_PEON*/)
+                        return SPELL_FAILED_BAD_TARGETS;
+                }
+                //By SpellIconID
                 if (m_spellInfo->SpellIconID == 1648)       // Execute
                 {
                     Unit* target = m_targets.getUnitTarget();
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4690,15 +4690,16 @@ SpellCastResult Spell::CheckCast(bool strict)
         {
             case SPELL_EFFECT_DUMMY:
             {
-                //By Spell ID
+                // By Spell ID
                 if (m_spellInfo->Id == 19938)               // Awaken Lazy Peon
                 {
                     Unit* target = m_targets.getUnitTarget();
-                    if (!target || !target->HasAura(17743 /*SPELL_PEON_SLEEP*/) || target->GetEntry() != 10556 /*NPC_SLEEPING_PEON*/)
+                    // 17743 = Lazy Peon Sleep | 10556 = Lazy Peon
+                    if (!target || !target->HasAura(17743) || target->GetEntry() != 10556)
                         return SPELL_FAILED_BAD_TARGETS;
                 }
-                //By SpellIconID
-                if (m_spellInfo->SpellIconID == 1648)       // Execute
+                // By SpellIconID
+                else if (m_spellInfo->SpellIconID == 1648)       // Execute
                 {
                     Unit* target = m_targets.getUnitTarget();
                     if (!target || target->GetHealth() > target->GetMaxHealth() * 0.2)
--- a/src/scriptdev2/scripts/kalimdor/durotar.cpp
+++ b/src/scriptdev2/scripts/kalimdor/durotar.cpp
@@ -61,12 +61,14 @@ struct npc_lazy_peonAI : public ScriptedAI
     bool m_bIsAtPile;
     float m_fSpawnO;
 
-    void RestartWakeTimer() {
+    void RestartWakeTimer()
+    {
         m_uiWakeTimer = urand(100000, 120000); //Random wakeup timer 1m40s to 2m, retail times unknown, 2m is duration of sleep aura
         m_uiSleepTimer = 0; //Ensure only one of the two timers is running at any given time
     }
 
-    void RestartSleepTimer() {
+    void RestartSleepTimer()
+    {
         m_uiSleepTimer = 80000; //Go to sleep after 1m20s, confirmed blizzlike
         m_uiWakeTimer = 0; //Ensure only one of the two timers is running at any given time
     }
@@ -113,9 +115,6 @@ struct npc_lazy_peonAI : public ScriptedAI
 
     void MovementInform(uint32 uiMotionType, uint32 uiPointId) override
     {
-        /*if (!uiPointId)
-            return;*/
-
         if (uiMotionType == POINT_MOTION_TYPE && uiPointId)
         {
             switch (uiPointId)
--- a/src/scriptdev2/scripts/kalimdor/durotar.cpp
+++ b/src/scriptdev2/scripts/kalimdor/durotar.cpp
@@ -127,9 +127,10 @@ struct npc_lazy_peonAI : public ScriptedAI
                 }
                 case 2: //spawn
                 {
-                    m_creature->SetFacingTo(m_fSpawnO);
-                    m_creature->GetMotionMaster()->MoveTargetedHome(); //hacky way of getting the peon to sleep
+                    m_creature->GetMotionMaster()->MoveIdle();
                     m_creature->HandleEmote(EMOTE_STATE_NONE);
+                    m_creature->SetFacingTo(m_fSpawnO);
+                    DoCastSpellIfCan(m_creature, SPELL_PEON_SLEEP);
                     RestartWakeTimer();
                     break;
                 }
@@ -152,6 +153,7 @@ struct npc_lazy_peonAI : public ScriptedAI
             if (m_uiGetUpTimer <= uiDiff)
             {
                 m_uiGetUpTimer = 0;
+                m_creature->HandleEmote(EMOTE_STATE_NONE);
                 m_creature->GetMotionMaster()->MoveWaypoint();
                 m_bIsAtPile = false;
             }
@@ -209,7 +211,13 @@ struct npc_lazy_peonAI : public ScriptedAI
             if (m_uiWakeTimer <= uiDiff)
                 Awaken(m_creature);
             else
+            {
+                //reapply aura if it fell off while the peon was unloaded
+                if (!m_creature->HasAura(SPELL_PEON_SLEEP))
+                    DoCastSpellIfCan(m_creature, SPELL_PEON_SLEEP);
+                //decrement timer
                 m_uiWakeTimer -= uiDiff;
+            }
         }
     }
 };
@@ -219,22 +227,18 @@ CreatureAI* GetAI_npc_lazy_peon(Creature* pCreature)
     return new npc_lazy_peonAI(pCreature);
 }
 
-bool EffectDummyCreature_lazy_peon_awake(Unit* pCaster, uint32 uiSpellId, SpellEffectIndex uiEffIndex, Creature* pCreatureTarget, ObjectGuid /*originalCasterGuid*/)
+bool EffectDummyCreature_lazy_peon_spell(Unit* pCaster, uint32 uiSpellId, SpellEffectIndex uiEffIndex, Creature* pCreatureTarget, ObjectGuid /*originalCasterGuid*/)
 {
-    // always check spellid and effectindex
+    //make sure it's the right spell
     if (uiSpellId == SPELL_AWAKEN_PEON && uiEffIndex == EFFECT_INDEX_0)
     {
-        //TODO some of these checks are now redundant, should be removed by someone who knows how to do so without breaking things
-        if (!pCreatureTarget->HasAura(SPELL_PEON_SLEEP) || pCaster->GetTypeId() != TYPEID_PLAYER || pCreatureTarget->GetEntry() != NPC_SLEEPING_PEON)
-            return true;
-
+        //awaken peon
         if (npc_lazy_peonAI* pPeonAI = dynamic_cast<npc_lazy_peonAI*>(pCreatureTarget->AI()))
             pPeonAI->Awaken(pCaster);
-
-        // always return true when we are handling this spell and effect
+        //right spell, return true
         return true;
     }
-
+    //wrong spell, return false
     return false;
 }
 
@@ -245,6 +249,6 @@ void AddSC_durotar()
     pNewScript = new Script;
     pNewScript->Name = "npc_lazy_peon";
     pNewScript->GetAI = &GetAI_npc_lazy_peon;
-    pNewScript->pEffectDummyNPC = &EffectDummyCreature_lazy_peon_awake;
+    pNewScript->pEffectDummyNPC = &EffectDummyCreature_lazy_peon_spell;
     pNewScript->RegisterSelf();
 }
--- a/src/scriptdev2/scripts/kalimdor/durotar.cpp
+++ b/src/scriptdev2/scripts/kalimdor/durotar.cpp
@@ -51,131 +51,6 @@ struct npc_lazy_peonAI : public ScriptedAI
     npc_lazy_peonAI(Creature* pCreature) : ScriptedAI(pCreature)
     {
         Reset();
-        //tree waypoint
-        //TODO do this via invisible creatures once they're added in the db (or actual waypoints)
-        switch (m_creature->GetGUIDLow())
-        {
-            case 6527:
-            {
-                m_fTreeX = -754.196;
-                m_fTreeY = -4141.18;
-                m_fTreeZ = 39.4156;
-                m_fTreeO = 1.54425;
-                break;
-            }
-            case 7373:
-            {
-                m_fTreeX = -775.806;
-                m_fTreeY = -4196.64;
-                m_fTreeZ = 44.5941;
-                m_fTreeO = 2.26289;
-                break;
-            }
-            case 6525:
-            {
-                m_fTreeX = -757.896;
-                m_fTreeY = -4324.19;
-                m_fTreeZ = 45.1184;
-                m_fTreeO = 4.3012;
-                break;
-            }
-            case 7374:
-            {
-                m_fTreeX = -622.255;
-                m_fTreeY = -4348.61;
-                m_fTreeZ = 41.0623;
-                m_fTreeO = 4.83115;
-                break;
-            }
-            case 6524:
-            {
-                m_fTreeX = -634.882;
-                m_fTreeY = -4480.31;
-                m_fTreeZ = 46.2619;
-                m_fTreeO = 4.71333;
-                break;
-            }
-            case 7376:
-            {
-                m_fTreeX = -498.072;
-                m_fTreeY = -4455.45;
-                m_fTreeZ = 51.0777;
-                m_fTreeO = 3.27999;
-                break;
-            }
-            case 3348:
-            {
-                m_fTreeX = -510.965;
-                m_fTreeY = -4372.57;
-                m_fTreeZ = 45.6692;
-                m_fTreeO = 0.480042;
-                break;
-            }
-            case 3346:
-            {
-                m_fTreeX = -334.424;
-                m_fTreeY = -4439.34;
-                m_fTreeZ = 54.6275;
-                m_fTreeO = 4.45416;
-                break;
-            }
-            case 3347:
-            {
-                m_fTreeX = -229.428;
-                m_fTreeY = -4449.48;
-                m_fTreeZ = 63.3728;
-                m_fTreeO = 0.016658;
-                break;
-            }
-            case 3345:
-            {
-                m_fTreeX = -225.614;
-                m_fTreeY = -4284.88;
-                m_fTreeZ = 65.0956;
-                m_fTreeO = 5.00395;
-                break;
-            }
-            case 7372:
-            {
-                m_fTreeX = -213.599;
-                m_fTreeY = -4220.75;
-                m_fTreeZ = 62.2392;
-                m_fTreeO = 1.93932;
-                break;
-            }
-            case 7375:
-            {
-                m_fTreeX = -265.354;
-                m_fTreeY = -4145.13;
-                m_fTreeZ = 55.767;
-                m_fTreeO = 5.78542;
-                break;
-            }
-            case 6523:
-            {
-                m_fTreeX = -320.191;
-                m_fTreeY = -4122.77;
-                m_fTreeZ = 51.3316;
-                m_fTreeO = 1.09659;
-                break;
-            }
-            case 6526:
-            {
-                m_fTreeX = -375.545;
-                m_fTreeY = -4018.32;
-                m_fTreeZ = 50.3465;
-                m_fTreeO = 2.86373;
-                break;
-            }
-            default:
-            {
-                //if there's no tree point, set it to current so the peon doesn't wander off
-                m_fTreeX = m_creature->GetPositionX();
-                m_fTreeY = m_creature->GetPositionY();
-                m_fTreeZ = m_creature->GetPositionZ();
-                m_fTreeO = m_creature->GetOrientation();
-            }
-        }
     }
 
     uint32 m_uiSleepTimer; //Time, until the npc goes to sleep
@@ -184,10 +59,6 @@ struct npc_lazy_peonAI : public ScriptedAI
     uint32 m_uiKneelTimer; //Time, delay kneeling to allow sound to finish playing
     uint32 m_uiGetUpTimer; //Time, npc stays at the pile so kneel animation can finish
     bool m_bIsAtPile;
-    float m_fTreeX;
-    float m_fTreeY;
-    float m_fTreeZ;
-    float m_fTreeO;
     float m_fSpawnO;
 
     void RestartWakeTimer() {
@@ -242,33 +113,36 @@ struct npc_lazy_peonAI : public ScriptedAI
 
     void MovementInform(uint32 uiMotionType, uint32 uiPointId) override
     {
-        if (uiMotionType != POINT_MOTION_TYPE || !uiPointId)
-            return;
+        /*if (!uiPointId)
+            return;*/
 
-        switch (uiPointId)
+        if (uiMotionType == POINT_MOTION_TYPE && uiPointId)
         {
-            case 1: //lumber pile
+            switch (uiPointId)
             {
-                m_creature->SetWalk(true);
-                m_bIsAtPile = true;
-                break;
+                case 1: //lumber pile
+                {
+                    m_creature->SetWalk(true);
+                    m_bIsAtPile = true;
+                    break;
+                }
+                case 2: //spawn
+                {
+                    m_creature->SetFacingTo(m_fSpawnO);
+                    m_creature->GetMotionMaster()->MoveTargetedHome(); //hacky way of getting the peon to sleep
+                    m_creature->HandleEmote(EMOTE_STATE_NONE);
+                    RestartWakeTimer();
+                    break;
+                }
             }
-            case 2: //tree
-            {
-                m_creature->SetFacingTo(m_fTreeO);
-                m_creature->PlayDistanceSound(SOUND_PEON_WORK);
-                m_creature->HandleEmote(EMOTE_STATE_WORK_CHOPWOOD);
-                m_uiChopTimer = urand(25000, 30000); //timer duration guessed, true retail time unknown
-                break;
-            }
-            case 3: //spawn
-            {
-                m_creature->SetFacingTo(m_fSpawnO);
-                m_creature->GetMotionMaster()->MoveTargetedHome(); //hacky way of getting the peon to sleep
-                m_creature->HandleEmote(EMOTE_STATE_NONE);
-                RestartWakeTimer();
-                break;
-            }
+        }
+        else if (uiMotionType == WAYPOINT_MOTION_TYPE)
+        {
+            //tree
+            m_creature->GetMotionMaster()->MoveIdle();
+            m_creature->PlayDistanceSound(SOUND_PEON_WORK);
+            m_creature->HandleEmote(EMOTE_STATE_WORK_CHOPWOOD);
+            m_uiChopTimer = urand(25000, 30000); //timer duration guessed, true retail time unknown
         }
     }
 
@@ -279,7 +153,7 @@ struct npc_lazy_peonAI : public ScriptedAI
             if (m_uiGetUpTimer <= uiDiff)
             {
                 m_uiGetUpTimer = 0;
-                m_creature->GetMotionMaster()->MovePoint(2, m_fTreeX, m_fTreeY, m_fTreeZ);
+                m_creature->GetMotionMaster()->MoveWaypoint();
                 m_bIsAtPile = false;
             }
             else
@@ -321,7 +195,7 @@ struct npc_lazy_peonAI : public ScriptedAI
                 float fX, fY, fZ;
                 m_creature->GetRespawnCoord(fX, fY, fZ, &m_fSpawnO);
                 m_uiSleepTimer = 0;
-                m_creature->GetMotionMaster()->MovePoint(3, fX, fY, fZ);
+                m_creature->GetMotionMaster()->MovePoint(2, fX, fY, fZ);
                 m_bIsAtPile = false;
                 m_uiGetUpTimer = 0;
                 m_uiKneelTimer = 0;
--- a/src/scriptdev2/scripts/kalimdor/durotar.cpp
+++ b/src/scriptdev2/scripts/kalimdor/durotar.cpp
@@ -33,14 +33,17 @@ EndContentData */
 
 enum
 {
-    SAY_PEON_AWAKE_1        = -1000795,
-    SAY_PEON_AWAKE_2        = -1000796,
+    SAY_PEON_AWOKEN = -1000795,
 
-    SPELL_PEON_SLEEP        = 17743,
-    SPELL_AWAKEN_PEON       = 19938,
+    SOUND_PEON_WORK = 6197,
+    SOUND_PEON_AWAKE_1 = 6292,
+    SOUND_PEON_AWAKE_2 = 6294,
 
-    NPC_SLEEPING_PEON       = 10556,
-    GO_LUMBERPILE           = 175784,
+    SPELL_PEON_SLEEP = 17743,
+    SPELL_AWAKEN_PEON = 19938,
+
+    NPC_SLEEPING_PEON = 10556,
+    GO_LUMBERPILE = 175784,
 };
 
 struct npc_lazy_peonAI : public ScriptedAI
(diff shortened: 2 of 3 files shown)


Phatcat:
I'll test and review this soon. Thank you for your work.

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,53 +33,208 @@ EndContentData */
...
 struct npc_lazy_peonAI : public ScriptedAI
 {
     npc_lazy_peonAI(Creature* pCreature) : ScriptedAI(pCreature)
     {
         Reset();
-        m_uiStopSleepingTimer = urand(30000, 120000);       // Set on spawn to a potential small timer, to get nice results for initial case
+        //tree waypoint
+        //TODO do this via invisible creatures once they're added in the db (or actual waypoints)
+        switch (m_creature->GetGUIDLow())
+        {
```
xfurry: **OMG!!! Using GUIDLow in SD2?!? This is clearly a NO NO.**

We must to find another way. For example creature_movement, which can support GUID movement and positions.
GUIDs are totally controlled by DB, so we cannot force a DB provider to set these GUIDs for the NPCs just because we want to.

BareRose: Yea I think I'm just going to start ignoring you entirely from now on. If you had read the TODO that's literally *in* the excerpt you posted you would know that this is a temporary solution that I used because I couldn't get creature_movement to work.

If you know how to get creature_movement to work for the tree positions you could use that knowledge to make an actual contribution instead of uselessly yelling.

Also, IIRC the GUID's are sniffed so they shouldn't change. And I honestly don't see how the DB provider would suddenly change these. It's not like it's a third party, it's part of the same larger project.

Rushor: calm down ladies, it's just a game :)

xfurry:
> Also, IIRC the GUID's are sniffed so they shouldn't change. And I honestly don't see how the DB provider would suddenly change these. It's not like it's a third party, it's part of the same larger project.

We don't use sniffed GUIDs. GUIDs are generated by DB providers and they need to be free to choose whatever they want. We have many DB providers, not only the https://github.com/cmangos/wotlk-db or the others that you see here. So we need to keep this compatible.


> If you know how to get creature_movement to work for the tree positions you could use that knowledge to make an actual contribution instead of uselessly yelling.

Using creature_movement is easy.
you just add your guid and movement pattern to each creature, then you trigger the movement by issues the MoveWaypoint() command in SD2 or DBscript.

Laizerox:
@BareRose GUIDs are not sniffed because on retail they're dynamic. Ie. whenever creature respawns it has a new GUID. Therefore you should avoid using static GUIDs in scripts especially when you consider the fact theres more than just one DB for cmangos.

BareRose:
> you just add your guid and movement pattern to each creature, then you trigger the movement by issues the MoveWaypoint() command in SD2 or DBscript.

As I said in the peon issue I've done exactly that and mangos wasn't finding the waypoints somehow. Help would be much appreciated.

cala:
You can find an example of how SD2 can interact with a creature using waypoint movement here:
https://github.com/cmangos/mangos-classic/blob/master/src/scriptdev2/scripts/kalimdor/desolace.cpp#L506
This is the script for the Gizelton Caravan in Desolace.

There is one difference however, the NPC is using waypoints from `creature_movement_template` I can't tell for sure that this will also work for waypoints from `creature_movement` like the one you are using.

BareRose:
@cala Maybe I'm blind or something, but I don't see any call to waypoint movement there. Clearly there's something I fundamentally misunderstand about how waypoints work in cmangos.

BareRose:
> especially when you consider the fact theres more than just one DB for cmangos.

This is specifically for classic though, it's a pull to mangos-classic, or am I misunderstanding something?

Laizerox:
@BareRose So you want to make it work with classic and let @xfurry and @cala to bash their heads backporting to master and tbc repo?

xfurry:
@BareRose 
For move waypoint you can use command 20 described here: https://github.com/cmangos/issues/wiki/DBScripts
Or for SD2 you can see an example here: https://github.com/cmangos/mangos-wotlk/blob/master/src/scriptdev2/scripts/northrend/utgarde_keep/utgarde_pinnacle/boss_skadi.cpp#L199

It is trued that for classic we only have one DB. But, as mentioned before, we are totally against hardcoding GUID in a script. There are many reasons for that.
Also, hardcoding xyz coords in SD2 or Core should be avoided as much as possible. That's why we need to use the DB for that.

I didn't have time to read the entire thread that was going on between you and @Phatcat this week. 
However, from what I understand, the way the Peon is behaving is wrong because of bad movement and bad coords. This can be easily fixed by DB by giving the exact coords, or in SD2, by generating the points dynamically, like shown in this function: https://github.com/cmangos/mangos-wotlk/blob/master/src/scriptdev2/scripts/northrend/azjol-nerub/ahnkahet/boss_jedoga.cpp#L188

BareRose:
@Laizerox Woah calm down. I was just asking a question. And yea, I guess ease of backporting is a good enough reason. Still they'll have to backport the waypoints anyways, which involves finding the right GUIDs and transferring the waypoints from classic-db to whichever db they're backporting to.

And like I said, I get using GUIDs like that is hacky, already talked that out with @Phatcat before I even submitted this PR. Will see if I can figure out creature_movement, otherwise @Phatcat will probably do it as he said he'd help me.

@xfurry Yea since I'm doing this in SD2 I'll use the SD2 way of doing it. And I think my mistake was calling pSummon->GetMotionMaster()->MoveWaypoint() with an argument to move to a specific waypoint, though since there's only 1 waypoint it should be fine without an argument. Will try that when I'm back home.

> I didn't have time to read the entire thread that was going on between you and @Phatcat this week. 

That's fair enough. Problem is you commented in that thread without really reading what you were commenting about multiple times, which (combined with the fact that I had just woken up) was why I had a hard time taking your review seriously. Eitherways, either me or @Phatcat should have the GUID issue sorted out soon enough.

cala:
> This is specifically for classic though, it's a pull to mangos-classic, or am I misunderstanding something?

Yes, this is a PR to the classic branch. But the quest is unchanged in TBC and WotLK and in such case, we will port the code to the two other cores, meaning it is preferable to have it as cross-core as possible. And we will point towards the original DB code as an example for other DB providers.

> @cala Maybe I'm blind or something, but I don't see any call to waypoint movement there. Clearly there's something I fundamentally misunderstand about how waypoints work in cmangos.

Or maybe this is I that misunderstand what you want to achieve by accessing DB waypoints from SD2. =)

From what I read in your script, for every NPC, waypoint 1 = lumber pile, waypoint 2 = tree, waypoint 3 = spawn point. This means that the waypoint path logic is the same for each NPC, from a scripting point of view, if you know at which WP the NPC is, you know exactly what it is doing so you know what action to script.

The various emote at each waypoints (kneeling, emotes, chopping) would be better handled by dbscripts_on_creature_movement in my opinion ([documentation](https://github.com/cmangos/issues/wiki/DBscripts)). The sleeping/awaking effect can then be handled by timer in SD2 code (like you are basically doing), along with making the NPC run when awaken (you just have to force it to walk again by DBScripts when reaching the tree or lumber pile waypoints).

If you want to stop the WP movement, you need to switch between `pSummon->GetMotionMaster()->MoveWaypoint()` and `pSummon->GetMotionMaster()->MoveIdle()` in SD2. But not that the NPC will stop moving when reaching the *next* waypoint. So if you want it to visually stop in the middle of its way toward the lumber pile from the tree, you need to set intermediate waypoints for this purpose.

If I misunderstood what you want the NPCs to do, please let me know.

BareRose:
@cala You misunderstood slightly: There is only 1 waypoint for the tree. The lumber pile is done dynamically through SD2 (and needs to be for maximum blizzlikeness) and the spawn point is already in the DB so doing it dynamically avoids duplicating it in the DB.

I've already looked at dbscripting/EventAI for this quest, and I've found that (at least with the current EventAI and dbscript systems, which I frankly feel could need some work, that may be my next project actually) it's just a tad too complex for dbscripting/EventAI to handle. I could partially handle some of it through dbscripts/EventAI as you say, but it would only make things more complicated for little to no benefit. For this reason I decided to do it through SD2 until it becomes possible to port the entire quest to dbscript/EventAI.

Anyway, I'm back home now so I'll try to make the waypoints work and update my PRs accordingly if I make any progress.

Phatcat:
@BareRose Hey mate, could you come on IRC possibly? It's gonna be much easier going back and forth there rather than starting up yet another 'endless' thread ^^

BareRose:
Alright, got the waypoint to work, except it doesn't seem to apply the orientation I specified in creature_movement. (Cmangos really doesn't handle orientations too well). Is there a way to get the orientation of a waypoint entry through SD2 so I can set it manually?

BareRose:
@Phatcat I'm on IRC now, name's BareRose.

Phatcat:
please connect to the rizon server and come on #cmangos =)

xfurry:
> Alright, got the waypoint to work, except it doesn't seem to apply the orientation I specified in creature_movement.

You need to fill in the "orientation" field in creature_template with the correct orientation. Trust me, it will work.

BareRose:
@xfurry There is no "orientation" field in creature_template... Did you mean a different table?

xfurry:
> @xfurry There is no "orientation" field in creature_template... Did you mean a different table?

Sorry, I meant creature_movement: https://github.com/cmangos/issues/wiki/creature_movement
https://github.com/cmangos/issues/wiki/creature_movement_template

BareRose:
@xfurry already did that, and like I said it's not applying it properly.
unbenannt

xfurry:
@BareRose    Maybe you are not giving the right value.
in the latest script that I developed it worked just fine: https://github.com/cmangos/wotlk-db/blob/6ac1654b1054bdecb089a8174ff9f7ba544e22a4/Updates/598_quests_7631_7629.sql#L79

Use your character, and then call the .gps command. That will give you the orientation of the character.  This will help you to find the right value.

BareRose:
That's creature_movement_template. I have to use creature_movement as each peon has a different tree waypoint. Maybe orientation isn't working right for creature_movement?

BareRose:
Commited the waypoint-using version of the script. Orientations of the waypoints still aren't working, but everything else should be (unless there's some weirdness with waypoints I'm unaware of).

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,89 +33,184 @@ EndContentData */
...
-            if (pInvoker->GetTypeId() == TYPEID_PLAYER)
-            {
-                DoScriptText(SAY_PEON_AWAKE_1, m_creature);
-                ((Player*)pInvoker)->KilledMonsterCredit(m_creature->GetEntry(), m_creature->GetObjectGuid());
-                m_creature->GetMotionMaster()->MovePoint(1, fX, fY, fZ);
-            }
-            else
-                m_creature->GetMotionMaster()->MovePoint(2, fX, fY, fZ);
+            m_creature->HandleEmote(EMOTE_STATE_NONE); //cancel chopping
+            m_creature->GetMotionMaster()->MovePoint(1, fX, fY, fZ);
```
xfurry: My question is:

I see that you use MovePoint(x,y,z) based on the dynamic coords that you get from the Gameobject.
Also I see that you use the MoveWaypoint() command below.

Why is that? My understanding is that these two functions should give the exact same result.

BareRose: From what I understand, MovePoint lets you send the npc to the specified x/y/z coordinates, and once it arrives MovementInform is triggered with POINT_MOTION_TYPE. That's all it does.

MoveWaypoint causes the npc to switch into "waypoint mode" so it'll follow its waypoints as specified in creature_movement. If there's multiple waypoints it'll keep moving between them until it's told to stop with MoveIdle. Waypoint-related MovementInform also has WAYPOINT_MOTION_TYPE instead of POINT_MOTION_TYPE.

xfurry: Your understanding is right.
My question was: why do you need both in the script? Can't we just use the one of the options?

BareRose: Nah, I need MovePoint so I can use dynamic coordinates (MoveWaypoint only uses static ones from DB), and I need MoveWaypoint to avoid having to hardcode the GUIDs. It's the most reasonable way of doing what needs to be done for the quest.

Phatcat: @xfurry
We are missing serverside tree GO waypoints; https://github.com/cmangos/issues/issues/1143#issuecomment-260153214 - if we had these we wouldn't need to use the waypoint system at all and could do it all through the point movement system;
> ```code
ServerToClient: SMSG_ON_MONSTER_MOVE (0x6E17)
GUID: Full: 0xF130293C000065D4 Type: Creature Entry: 10556 Low: 26068
Spline Type: FacingTarget (3)
Facing GUID: Full: 0xF112AEA700002B09 Type: GameObject Entry: 175783 Low: 11017
```
Object missing in WoTLKDB... and probably Classic/TBC also... (i'll add soon*)
```code
175783: TreeWayPoint (DND)
Type: 5
displayID: 327
Size: 1
Data1: 1
Data2: 1
Data6: -1
```
Unfortunately - spawnpoints for these objects are hidden

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -130,11 +225,12 @@ bool EffectDummyCreature_lazy_peon_awake(Unit* pCaster, uint32 uiSpellId, SpellE
     // always check spellid and effectindex
     if (uiSpellId == SPELL_AWAKEN_PEON && uiEffIndex == EFFECT_INDEX_0)
     {
+        //TODO some of these checks are now redundant, should be removed by someone who knows how to do so without breaking things
```
Phatcat: Don't doubt yourself.

ask yourself if the code below makes sense; basically what's happening is the spell is cast on the peon; if the spell failed in checkcast, it won't get to hit the peon, so all of this won't execute at all. but what you are really checking, is if the target is not sleeping, or it's not a player casting it, or the target isn't a peon, the.... return true? - that's all wrong. It should be return false in that case.

the reason why it was used in a weird way before is because the spell was used by the peon on himself in the old script. It isn't any longer. the check is completely redundant, especially when it returns the absolute opposite of what it ought to.

src/game/Spell.cpp
```diff
@@ -4508,6 +4508,14 @@ SpellCastResult Spell::CheckCast(bool strict)
...
             case SPELL_EFFECT_DUMMY:
             {
+                //By Spell ID
+                if (m_spellInfo->Id == 19938)               // Awaken Lazy Peon
+                {
+                    Unit* target = m_targets.getUnitTarget();
+                    if (!target || !target->HasAura(17743 /*SPELL_PEON_SLEEP*/) || target->GetEntry() != 10556 /*NPC_SLEEPING_PEON*/)
+                        return SPELL_FAILED_BAD_TARGETS;
+                }
+                //By SpellIconID
```
Phatcat: should make the line below 'else if' instead of 'if' - no reason to execute the checks if we just ran the if above.

src/game/Spell.cpp
```diff
@@ -4508,6 +4508,14 @@ SpellCastResult Spell::CheckCast(bool strict)
         {
             case SPELL_EFFECT_DUMMY:
             {
+                //By Spell ID
+                if (m_spellInfo->Id == 19938)               // Awaken Lazy Peon
+                {
+                    Unit* target = m_targets.getUnitTarget();
+                    if (!target || !target->HasAura(17743 /*SPELL_PEON_SLEEP*/) || target->GetEntry() != 10556 /*NPC_SLEEPING_PEON*/)
```
Phatcat: Should consider moving the comments out of the calls, normally comments are only done directly in code in declarations, in calls we comment above usually, the comment could looks something like
`// 17743 = Lazy Peon Sleep | 10556 = Lazy Peon`

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,89 +33,184 @@ EndContentData */
...
-    uint32 m_uiStopSleepingTimer;                           // Time, until the npcs (re)starts working on its own
+    uint32 m_uiSleepTimer; //Time, until the npc goes to sleep
+    uint32 m_uiWakeTimer;  //Time, until the npc wakes up on its own
+    uint32 m_uiChopTimer; //Time, until the npc stops chopping at the tree
+    uint32 m_uiKneelTimer; //Time, delay kneeling to allow sound to finish playing
+    uint32 m_uiGetUpTimer; //Time, npc stays at the pile so kneel animation can finish
+    bool m_bIsAtPile;
+    float m_fSpawnO;
+
+    void RestartWakeTimer() {
```
Phatcat: The `{` should have it's own line. I should have shown you this; https://github.com/cmangos/issues/wiki/Coding-Standards

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,89 +33,184 @@ EndContentData */
...
         }
         else
             script_error_log("No GameObject of entry %u was found in range or something really bad happened.", GO_LUMBERPILE);
     }
 
     void MovementInform(uint32 uiMotionType, uint32 uiPointId) override
     {
-        if (uiMotionType != POINT_MOTION_TYPE || !uiPointId)
-            return;
+        /*if (!uiPointId)
```
Phatcat: should maybe add a comment stating why it's temporarily commented out, that we are currently missing game object waypoints in the trees.

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: I'm not doubting myself on the things you brought up, I'm doubting myself on the true/false as I don't really understand what true/false do in that context exactly. If you could clear that up I could easily clean up that check thing.

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: Nah I've already read that page myself, hence I've applied it everywhere. I guess I just forgot in that case. Will fix.

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: Actually it isn't meant to be commented out at all, it's meant to be removed. Will fix.

src/game/Spell.cpp
BareRose: I figured someone might want to do something for a spell both based on its ID and its icon, but I guess if that's ever needed they can still change it back to an if instead of if else.

src/scriptdev2/scripts/kalimdor/durotar.cpp
Phatcat: Let me ask you in another way then, look at line 238. we return false. That's being done when the peon is hit by a wrong spell. the only reason to return true is if the contents of this dummy (line 231 + 232) have been executed, otherwise we should return false.

`pCreatureTarget->GetEntry() != NPC_SLEEPING_PEON` in line 228 is as redundant as can be; the only npc that can run this script is this npc, with that entry, and there is nothing I can think of which would allow that creature to change entry during runtime yet still continue keep it's sd2 script. Even more, if this npc wasn't the target to begin with, this function would have never been executed, no?

Now, for the caster check, since the spell is no longer cast by the npc itself on itself, but only through an item, which is now being checked properly in checkcast, then I don't see it possible that any creature could ever use that spell, the spell is now exclusively related to that item.

and for the aura check, basically it's the same, since it's now being done it checkCast it's just a redundant check since there's no real chance of the mob waking up on it's own in between the time it takes for spell to be cast by the item and when this dummy is triggered.

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,89 +33,183 @@ EndContentData */
...
+                case 1: //lumber pile
+                {
+                    m_creature->SetWalk(true);
+                    m_bIsAtPile = true;
+                    break;
+                }
+                case 2: //spawn
+                {
+                    m_creature->SetFacingTo(m_fSpawnO);
+                    m_creature->GetMotionMaster()->MoveTargetedHome(); //hacky way of getting the peon to sleep
```
Phatcat: To get the peon to sleep, try having it cast spell 17743 on itself.

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: Doesn't exactly answer what true/false do, i.e. what is whatever is calling the dummy check thing doing with the return value. But it looks like it basically should just return true in the middle part (and awake the peon obviously). Not sure if it's even possible for uiSpellId to be wrong, so maybe that check is redundant too.

Phatcat: It is very much possible for uispellId to be wrong, to my understanding, this function;
https://github.com/cmangos/mangos-classic/pull/220/files#diff-e1032aae835a2b320c86b6ff4b63a3a8L128
is executed whenever the peon is hit by **any** spellEffect, what-so-ever, (it's simply a EffectDummyNPC with a poor name; https://github.com/cmangos/mangos-classic/pull/220/files#diff-e1032aae835a2b320c86b6ff4b63a3a8R248) <- as you can see, it's a function defined in the script exclusively. Not used by the core. If you don't use it, reduce it =)

I'd say rather turn it into a void function instead of an int function since the result is not used anywhere in the script anyways, and then perhaps also give it a slightly better name.

to me, I see no reason it couldn't just be;

void EffectDummyCreature_lazy_peon(Unit* pCaster, uint32 uiSpellId, SpellEffectIndex uiEffIndex, Creature* pCreatureTarget, ObjectGuid /*originalCasterGuid*/)
{
// Only handle peon awaken, ignore any other spell effects.
if (uiSpellId == SPELL_AWAKEN_PEON && uiEffIndex == EFFECT_INDEX_0)
if (npc_lazy_peonAI* pPeonAI = dynamic_cast<npc_lazy_peonAI*>(pCreatureTarget->AI()))
pPeonAI->Awaken(pCaster);
}

but if it for some reason has to be an int function because of sd2 internals then it could be;

void EffectDummyCreature_lazy_peon(Unit* pCaster, uint32 uiSpellId, SpellEffectIndex uiEffIndex, Creature* pCreatureTarget, ObjectGuid /*originalCasterGuid*/)
{
// Only handle peon awaken, ignore any other spell effects.
if (uiSpellId == SPELL_AWAKEN_PEON && uiEffIndex == EFFECT_INDEX_0)
if (npc_lazy_peonAI* pPeonAI = dynamic_cast<npc_lazy_peonAI*>(pCreatureTarget->AI()))
{
pPeonAI->Awaken(pCaster);
return true;
}

// failed to cast to AI or wrong spell
return false;
}

(sorry for poor formatting)

BareRose: > the result is not used anywhere in the script anyways

Now that's useful. Also explains why I couldn't find the place where SD2 uses the return. I'll turn it into a void that only does peon awaking stuff and see if SD2 complains.

BareRose:
Hope I put in the rebase correctly, sorry if I made a mess by accident.

Phatcat:
Don't worry about makin a mess, it can all be cleaned up easy before merging, we'll take care of that later.

I can tell you that you did not put them in the right updates folder though, I meant the updates folder in the database repo for the database updates, we only use the updates folder in the core for updates to the base sql files (structural changes)

sorry if I didn't make that clear before.

BareRose:
Fucking hell now I have to move it back and rebase classic-db aswell. Removed that commit, gonna put one into DB.

Phatcat:
I see no reason not to squeeze at least the first 3 commits together. Probably also along with the fourth.

also on the top of the durotar script it states a completion percentage. that probably shouldn't be 100% considering that some of the orientations don't seem to work currently, along with the missing tree waypoint GOs, etc.

none of that kept me from testing it, though, and functionality wise it seems to be working really well.

Couldn't find a single issue in-game :+1: , sure maybe a tiny bit of polish here and there with timers and stuff if we get some more data on the matter.

Phatcat:
Okay, there is one small issue (which I think is unrelated to your script) but the sound of the chopping seems to be directly related to the camera of the client looking at it, if the player is standing right next to the peon but you position him out of view of the camera there is no sound from the choppin at all. It should work just like the other sounds, as distance sounds. May be a more general issue in the core rather than with this script, can't seem to find where the sound of the lumbering is handled.

Warlockbugs:
@Phatcat 
Chopping sound originates from unit M2 model animation, so it is bound to unit position.
Client can use either player position for 3D sound picture or camera position, these modes are being switched by CVar `Sound_ListenerAtCharacter` for newer clients or `SoundListenerAtCharacter` for vanilla.

Phatcat:
Well, I guess this is ready to merge then if there's nothing to be done about the sound and that's just how it behaves.

still should be squashed, however, there's no real reason for it to be split up into 4 commits, and completion stated at the top shouldn't be at 100% for reasons I've already mentioned.

@cala @xfurry @killerwife  what do you think?

BareRose:
Yea obviously it would need to be squashed. I assumed that was part of the merging process you guys use, but if you need me to do it myself I can definitely do that.

Also I'm not sure it's actually ready. I've seen some weird glitches while playing with the fix. Did you do an actual in-game test? If so and you didn't have any issues maybe it's the fact I upped my visibility settings in the conf, but that shouldn't be causing AI glitches so it's still a problem. Haven't yet found time to try to figure out how to reproduce the glitches and what is causing them, but once I do find time I'll try to fix them to the best of my abilities.

Phatcat:
> Did you do an actual in-game test?

Seriously?

you think I'd say this;

> none of that kept me from testing it, though, and functionality wise it seems to be working really well.
> Couldn't find a single issue in-game 👍 , sure maybe a tiny bit of polish here and there with timers and stuff if we get some more data on the matter.

if not?

this whole comment i made after continuing to test it for a bit; https://github.com/cmangos/mangos-classic/pull/220#issuecomment-262005047 would make no sense if I hadn't tried it out in game, now would it?

you want details on how I didn't find any issues really worth of mention? fine.

I tested the blackjack on the peons in different states as well as different targets and the blackjack seemed to be working as expected.
I did NOT test the blackjack on all 14 peons to see how they behaved when getting hit, I tested 4 or 5 different (some of them several times) - the AI is generic for all of them, if the issue was with the AI it would have shown on any one of them.
I think the 'work work' chant that they do shouldn't be happening staticly whenever they get to the tree, it should happen WHILE they are working, probably somewhat at random. I considered this neglectable so I thought I wouldn't mention it as it would require a rewrite just for such a small thing we don't even have sniff data on to get correct in the first place.
I went around and looked at the different sleeping and working points of the different peons, I saw no issue there.

I don't know what kind of people you are used to but I don't like getting questioned on the things I explicitly say I've done. I'm not a liar. Frankly I am getting a bit tired of spending all this time helping someone who clearly doesn't appreciate it.

BareRose:
@Phatcat I meant no offense. I didn't see the message where you explicitly stated you had tested in-game, so for all I could tell you mainly reviewed code or only did a brief test (the issues I'm talking about don't happen in a brief test). Definitely not calling you a liar.

Phatcat:
That's okay, I snapped and I shouldn't have. I apologize for that.
You are helping us and spending a lot of time on this, please know that I do appreciate that and will always be willing to help out anyone interested in helping out cmangos.

Well, it would help some if you were to actually elaborate on those issues, in explicit terms - maybe I'll be able to help figure out what is going on.

BareRose:
Well like I said I haven't had much time to really figure them out, and I'l fairly certain there's several more I haven't yet encountered. But anyway, here's what I know thus far:

One of the peons (the one to the left of the hidden path) seems to sometimes miss the pile when woken up, meaning that he goes to the pile but doesn't stop and just goes straight to the tree, often playing the kneeling animation while moving. Another thing is that peons sometimes glitch out and just stop at the pile indefinitely.

Eitherways, I'm 99% sure these are in some way related to the pile and having to use WP movement for the tree. Possible that the engine is having issues with three different ways of moving so close together.

I'll be testing this later today, so we'll see what exactly is going on,

xfurry:
This looks better now. So if you guys tested the script, it should be fine to push.

My only note is regarding the ```if (!target || !target->HasAura(17743) || target->GetEntry() != 10556)``` check in the core.
I know that this is correct and makes sense, but it would be nice to have a better design.
For example, the entry you can check in the ```item_require_target``` table. There is no possibility to check for the aura yet, but, eventually, we could add this in the table as well (same table), or link this to the condition tables.

Phatcat:
@xfurry
Yeah, if we add the peon in the item_require_target, we'd still need to check for the aura in checkcast to be able to supress the actual item use, so we'd still be pestering stuff in checkcast, now it would just be fractured as well... Until we are able to check for more stuff in the item_require_target such as states and auras of the target, wouldn't be better to keep it collected in one place?

also, as BareRose says, there are a couple of minor hickups with the AI where it'll misbehave or interrupt completely. I didn't experience that while testing, but I probably should test it more.

BareRose:
Welp, turns out the major issue I was experiencing (peons getting stuck at their piles) was caused by the fact that I replaced the number before my sql file with 'xxxx', changing it to '1xxx' fixes that as apparently the install script only applies updates starting with a number. Fixed now.

There's still the other thing (peons missing the pile) that I'll test a bit more.

BareRose:
Turns out peons walking past their pile were doing this only if they were on their way to the tree as their sleep timer ran out and they went to sleep, so waypoint movement never got cancelled as they never hit the tree. Fixed this by adding some extra idle movement commands and also added some EMOTE_STATE_NONE in a few places to fix other more minor animation issues.

Also, I'm absolutely sure that peons saying "work work" whenever they reach their tree is correct. If you watch Joana's video you'll notice that the sound is only heard if a nearby peon just reached his tree.

Phatcat:
Well that is just great, found and fixed all the issues yourself. :+1:

I'm heading to bed now, but I'll get some feedback for you sometime tomorrow. I'll try to spend some more time with each peon individually this time around.

> Also, I'm absolutely sure that peons saying "work work" whenever they reach their tree is correct.

You sound way more confident about it than I am in doubting the implementation. I do trust you but I'm gonna skim through joana's first video like you suggest, if nothing else then simply just to prove you right.

Phatcat:
I am testing this out in game currently and I can say that if the peons wake up on their own (at least the first time they do), they will no longer do anything at all other than standing up. This seems to be consistent and with all the peons

I found not a single issue with the player waking them up, though, so if those issues you experienced earlier are gone, then I think at least that part is good and solid now.

I also skimmed through joana's video and found evidence for what you're saying.
https://youtu.be/SwLjJH8LXfc?t=14m45s
https://youtu.be/SwLjJH8LXfc?t=22m14s

Phatcat:
> if the peons wake up on their own (at least the first time they do), they will no longer do anything at all other than standing up. This seems to be consistent and with all the peons

Yeah, it happened again, I am not sure what makes it happen, but it just happened again on two other peons. They will pick up eventually and start behaving, and then they will not misbehave as long as they are kept active (within distance of some player) but for the first cycle they become active (get in range of a player) they seem to not do anything at all.
Never happens when the player wakes them up, only happens if they wake up on their own.

Phatcat:
Yeah, now I can reproduce consistently. Has to do with their active state. Try moving around between all of the peons and you'll notice that a lot of those you get to, if they are not sleeping then they are just standing around idle.

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -127,9 +127,10 @@ struct npc_lazy_peonAI : public ScriptedAI
                 }
                 case 2: //spawn
                 {
-                    m_creature->SetFacingTo(m_fSpawnO);
-                    m_creature->GetMotionMaster()->MoveTargetedHome(); //hacky way of getting the peon to sleep
+                    m_creature->GetMotionMaster()->MoveIdle();
                     m_creature->HandleEmote(EMOTE_STATE_NONE);
+                    m_creature->SetFacingTo(m_fSpawnO);
+                    m_creature->CastSpell((Unit*)m_creature, SPELL_PEON_SLEEP, false);
```
xfurry: in SD2 we don't use this format, unless the spell is missing from DBC and added by DB.
Please replace with ```DoCastSpellIfCan(m_creature, SPELL_PEON_SLEEP);```
Also the cast from m_creature to Unit* is not required, as m_creature is already a Unit.

BareRose: That's what I did at first, but it didn't want to accept the arguments so I added more and did a cast. Tried again this way now and it works. Weird, maybe my VS was being dumb last time.

src/scriptdev2/scripts/kalimdor/durotar.cpp
```diff
@@ -33,89 +33,185 @@ EndContentData */
...
+                    m_creature->GetMotionMaster()->MoveIdle();
+                    m_creature->HandleEmote(EMOTE_STATE_NONE);
+                    m_creature->SetFacingTo(m_fSpawnO);
+                    m_creature->CastSpell((Unit*)m_creature, SPELL_PEON_SLEEP, false);
+                    RestartWakeTimer();
+                    break;
+                }
+            }
+        }
+        else if (uiMotionType == WAYPOINT_MOTION_TYPE)
```
BareRose: Don't see anything wrong with my code there.

BareRose:
Welp, looks like the command indeed needs 3 arguments. @xfurry

BareRose:
@Phatcat Yea noticed that too way back, but it never happened after that (probably because I play on a very high visibility range). Not sure it's anything within my control, maybe active states just break things and the problem is with how they deal with SD2 rather than my code. Will see if I can find a way to prevent it.

src/scriptdev2/scripts/kalimdor/durotar.cpp
xfurry: I didn't say it's wrong. I was asking you if that's your intended purpose.
Usually for WAYPOINT_MOTION_TYPE we provide a uiPointId.

src/scriptdev2/scripts/kalimdor/durotar.cpp
xfurry: > Welp, looks like the command indeed needs 3 arguments. @xfurry

Which command?

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: There's only one waypoint, so no need to check for uiPointId.

src/scriptdev2/scripts/kalimdor/durotar.cpp
BareRose: Nvm I had the wrong command. Thx for pointint that out.

BareRose:
It seems I've found what's causing the problems with peons just standing around: Apparently the engine doesn't call UpdateAI when the peon is outside the loaded area, but the aura's duration is still ticking down so the aura will fall off because it only has 2min duration while the timer is paused. Could probably hackfix just for this quest, but I feel like there may be a deeper problem here.

BareRose:
Hello? Need to know if I'm supposed to just hackfix this (and how), or if this is some engine-side problem that needs to be addressed there, or if there's some "cleaner" way of doing my timers I'm unaware of.

@Phatcat

Rushor:
on tc there is a function called setActive() which solve such grid loading issues 
https://github.com/TrinityCore/TrinityCore/blob/7fbc4374f33ac296432e6e042c9ddd7e01a0bdfc/src/server/game/Entities/Object/Object.cpp#L1495

xfurry:
That function also exists in cmangos. Alternatively you can enable the config option to keep the grid loaded.
But, it's not really recommended to use this function on world maps. It might increase the memory and cpu usage too much.

BareRose:
Yea forcing it to always be active is probably bad. Are there any functions to get the server's system time? If  so I could tie my timers to that so when a peon in un- and then re-loaded and the timer's been exceeded it will automatically wake up and resume its rythm. Seems like the best solution for this.

BareRose:
Any news on this? Will probably have time to work on this again this weekend. If nobody has a better idea by then I'll just hackfix by re-applying the sleep aura when a peon gets re-loaded while he's in sleep state. Most direct way of fixing the issue IMO.

BareRose:
Alright, added a check that will re-apply the sleep aura when needed. Hacky but better than forcing the peons to always be loaded.

Also, it turns out the supposedly free VStudio Community Edition isn't so free after its 30 day period runs out, so I haven't been able to compile & test any further. Would be great if @Phatcat or someone else could do another quick test and this could finally be merged if there are no further issues.

BareRose:
Anyone else still testing this to confirm it works and can be merged in? So far haven't found any further issues myself.

@Phatcat @cala @xfurry

cala:
I'm in holidays at the moment. I will review this when I'm back.

cala:
The code seems OK to me. I'll test this week and merge it if it's ok. 👍

Rushor:
btw, in the original issue, someone said that there should be gameobjects inside the trees that could be used as movement trigger. I know the author scripted it without it and it works fine, but is there any chance we could get more information about this gameobject? (template and spawns?) :)

Rushor:
ah nvm got it 175783: TreeWayPoint (DND)

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/272157e47f03545093c66edea447f734ea914f53.
Thanks to @BareRose and everyone involved in reviewing, testing and researching this PR. :+1:

cala:
(and sorry it took so long)

PR #210, opened 2016-10-26
General charm/pet fixes

cyberium:
- Now creatures can be charmed.
- Now players can be charmed.
- We can now charm/possess even if we have already a summoned pet.
- Charmed pet have their own threatlist. But their original one before they were charmed is preserved.
- Charmer have now attack spell in pet bar even if he charmed a player.
- Its now possible to charm a player while dueling.
- Charmed players/creature will not be teleported if areatrigger is triggered.
- Uncharm before telport.
- We can now stable unsumoned pet.
- If we are in flight while loading the pet will be set as temporary unsumoned and will be correctly resummoned at landing.
- Priest pet is now not saved to db.
- Its now not possible to summon/stable a dead pet. If its not summoned we are now able to revive it from DB.
- Client error message are improved in different case like trying to summon a dead pet.
- Now in combat state is applied only after spell hit. (before it was impossible to cast a spell with not in combat attribute)

PR #221, opened 2016-11-20
Fix memory leak in Unit

Jose123456:
Found with Valgrind


cyberium:
Thank you very much.
However this is already fixed in upcoming PR -> https://github.com/cmangos/mangos-classic/pull/210/files#diff-b1318d9ecce2e0c12047a2d2c24da46cR408 so i can close this.

PR #222, opened 2016-11-20
Improved Mend Pet fix

HanFox:
This is my first time both working with cmangos and C++, so I apologise for any mess or bits that are just plain silly. I made the PR then realised it had issues, and panicked and made a mess of the PR. Now I just hope someone more knowledgeable can give me some feedback about it and how best to implement it.

This PR attempts to fix the fact Improved Mend Pet does not roll to dispel every tick (https://github.com/cmangos/issues/issues/306) and it'd be desirable if it removed the "Nothing to dispel" message.

The way I have set the chance is currently hard-coded as I couldn't figure out how to pull the information from the spell or the talent information.

It currently also ignores the "Nothing to dispel" check/error and just casts a dispel regardless -- this particular behaviour is undesirable, but I'm not sure how to get around not needing to return a spell result.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4251,11 +4251,12 @@ SpellCastResult Spell::CheckCast(bool strict)
             bool dispelTarget = false;
             uint32 mechanic = m_spellInfo->EffectMiscValue[0];
             SpellEntry const* spell = nullptr;
-
+            
             Unit::SpellAuraHolderMap& Auras = target->GetSpellAuraHolderMap();
             for (Unit::SpellAuraHolderMap::iterator iter = Auras.begin(); iter != Auras.end(); ++iter)
             {
                 spell = iter->second->GetSpellProto();
+                
                 if (spell->Dispel == mechanic)
                 {
                     dispelTarget = true;
@@ -4263,8 +4264,12 @@ SpellCastResult Spell::CheckCast(bool strict)
                 }
             }
 
+            // Need to see if target is a pet in order to avoid showing dispel fail message
+            Creature* target = (Creature*)m_targets.getUnitTarget();
+
             if (!dispelTarget)
-                return SPELL_FAILED_NOTHING_TO_DISPEL;
+                if (!target->IsPet())
+                    return SPELL_FAILED_NOTHING_TO_DISPEL;
         }
     }
 
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3185,6 +3185,11 @@ void Spell::SendCastResult(Player* caster, SpellEntry const* spellInfo, SpellCas
 {
     WorldPacket data(SMSG_CAST_FAILED, (4 + 1 + 1));
     data << uint32(spellInfo->Id);
+    
+    if (spellInfo->Id == 24406 && result == SPELL_FAILED_NOTHING_TO_DISPEL) // Improved mend pet - don't send anything to player if nothing to dispel
+    {
+        return;
+    }
 
     if (result != SPELL_CAST_OK)
     {
@@ -4246,7 +4251,7 @@ SpellCastResult Spell::CheckCast(bool strict)
 
         // Classic only (seems to be handled client side tbc and later (not verified)) - Only check effect 0
         // Spells such as Abolish Poison have dispel on effect 1 but should be castable even if target not poisoned
-        if (m_spellInfo->Effect[0] == SPELL_EFFECT_DISPEL && m_spellInfo->Id != 24406) // Ignore Hunter's Improved Mend Pet
+        if (m_spellInfo->Effect[0] == SPELL_EFFECT_DISPEL)
         {
             bool dispelTarget = false;
             uint32 mechanic = m_spellInfo->EffectMiscValue[0];
@@ -4256,7 +4261,7 @@ SpellCastResult Spell::CheckCast(bool strict)
             for (Unit::SpellAuraHolderMap::iterator iter = Auras.begin(); iter != Auras.end(); ++iter)
             {
                 spell = iter->second->GetSpellProto();
-                if (spell->Dispel == mechanic)
+                if (spell->Dispel == mechanic || (m_spellInfo->Id == 24406 && (spell->Dispel == DISPEL_MAGIC || spell->Dispel == DISPEL_CURSE || spell->Dispel == DISPEL_DISEASE || spell->Dispel == DISPEL_POISON)))
                 {
                     dispelTarget = true;
                     break;
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4263,11 +4263,10 @@ SpellCastResult Spell::CheckCast(bool strict)
                 }
             }
 
-            // Need to see if target is a pet in order to avoid showing dispel fail message
-            Creature* target = (Creature*)m_targets.getUnitTarget();
-
             if (!dispelTarget)
-                if (!target->IsPet())
+                if (m_spellInfo->Id == 24406) // Hunter's Improved Mend Pet
+                    return SPELL_CAST_OK;
+                else
                     return SPELL_FAILED_NOTHING_TO_DISPEL;
         }
     }
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4246,7 +4246,7 @@ SpellCastResult Spell::CheckCast(bool strict)
 
         // Classic only (seems to be handled client side tbc and later (not verified)) - Only check effect 0
         // Spells such as Abolish Poison have dispel on effect 1 but should be castable even if target not poisoned
-        if (m_spellInfo->Effect[0] == SPELL_EFFECT_DISPEL)
+        if (m_spellInfo->Effect[0] == SPELL_EFFECT_DISPEL && m_spellInfo->Id != 24406) // Ignore Hunter's Improved Mend Pet
         {
             bool dispelTarget = false;
             uint32 mechanic = m_spellInfo->EffectMiscValue[0];
@@ -4264,10 +4264,7 @@ SpellCastResult Spell::CheckCast(bool strict)
             }
 
             if (!dispelTarget)
-                if (m_spellInfo->Id == 24406) // Hunter's Improved Mend Pet
-                    return SPELL_CAST_OK;
-                else
-                    return SPELL_FAILED_NOTHING_TO_DISPEL;
+                return SPELL_FAILED_NOTHING_TO_DISPEL;
         }
     }
 
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3186,11 +3186,6 @@ void Spell::SendCastResult(Player* caster, SpellEntry const* spellInfo, SpellCas
     WorldPacket data(SMSG_CAST_FAILED, (4 + 1 + 1));
     data << uint32(spellInfo->Id);
     
-    if (spellInfo->Id == 24406 && result == SPELL_FAILED_NOTHING_TO_DISPEL) // Improved mend pet - don't send anything to player if nothing to dispel
-    {
-        return;
-    }
-
     if (result != SPELL_CAST_OK)
     {
         data << uint8(2); // status = fail
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -4251,12 +4251,11 @@ SpellCastResult Spell::CheckCast(bool strict)
             bool dispelTarget = false;
             uint32 mechanic = m_spellInfo->EffectMiscValue[0];
             SpellEntry const* spell = nullptr;
-            
+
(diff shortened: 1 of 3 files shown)


src/game/Spell.cpp
```diff
@@ -4263,11 +4263,10 @@ SpellCastResult Spell::CheckCast(bool strict)
                 }
             }
 
-            // Need to see if target is a pet in order to avoid showing dispel fail message
-            Creature* target = (Creature*)m_targets.getUnitTarget();
-
             if (!dispelTarget)
-                if (!target->IsPet())
+                if (m_spellInfo->Id == 24406) // Hunter's Improved Mend Pet
+                    return SPELL_CAST_OK;
```
HanFox: Not sure how else to return. If no return is passed now it'll randomly try to dispel nothing.

src/game/SpellAuras.cpp
```diff
@@ -4389,6 +4389,33 @@ void Aura::PeriodicTick()
...
+                    {
+                        int32 triggerAmount = 0;
+                        
+                        // Set threshold for dispel check
+                        if (pCaster->HasSpell(19572))
+                            triggerAmount = 15; //rank 1
+                            
+                        if (pCaster->HasSpell(19573))
+                            triggerAmount = 50; //rank 2
+
```
HanFox: Is there a better way to get the triggerAmount? I tried to do it via Talents, or spellProto modifiers, but could never return a value.

src/game/SpellAuras.cpp
```diff
@@ -4389,6 +4389,33 @@ void Aura::PeriodicTick()
                     }
                 }
             }
+
+            // Hunter's Improved Mend Pet dispel check
+            if (spellProto->SpellFamilyName == SPELLFAMILY_HUNTER && spellProto->SpellFamilyFlags && spellProto->SpellIconID == 267)
+            {
+                // Check caster has one of the ranks of Improved Mend Pet
+                if (pCaster->HasSpell(19572) || pCaster->HasSpell(19573)) {
```
Phatcat: { should be on the line below.

src/game/SpellAuras.cpp
```diff
@@ -4389,6 +4389,33 @@ void Aura::PeriodicTick()
...
+                        // Set threshold for dispel check
+                        if (pCaster->HasSpell(19572))
+                            triggerAmount = 15; //rank 1
+
+                        if (pCaster->HasSpell(19573))
+                            triggerAmount = 50; //rank 2
+
+                        if (roll_chance_i(triggerAmount))
+                        {
+                            pCaster->CastSpell(target, 24406, true, nullptr, this);
```
Phatcat: no need for brackets on this single line contents

src/game/Spell.cpp
```diff
@@ -4264,7 +4264,10 @@ SpellCastResult Spell::CheckCast(bool strict)
             }
 
             if (!dispelTarget)
-                return SPELL_FAILED_NOTHING_TO_DISPEL;
+                if (m_spellInfo->Id == 24406) // Hunter's Improved Mend Pet
```
Phatcat: basically you want to revert all of this right here. This is the reason why;
> It currently also ignores the "Nothing to dispel" check/error and just casts a dispel regardless -- this particular behaviour is undesirable

is happening. what you wanna try instead is change line https://github.com/cmangos/mangos-classic/pull/222/files#diff-b4db11cbf2813cd24b3c165885fc08a7R4249 to
if (m_spellInfo->Effect[0] == SPELL_EFFECT_DISPEL && m_spellInfo->Id != 24406) // Ignore Hunter's Improved Mend Pet)

by doing that you also avoid having to execute the whole shabang of dispel check, and you stop returning spell cast ok prematurely. possibly the basepoints or the effectmiscvalues for this effect, particularly for classic, could be used in a way we don't yet understand.

The whole dispel check was just recently implemented, it seems like the classic branch is a bit different to the other branches. Perhaps it could be interesting to see the behaviour of the spell/talent on the other cores, if it wasn't removed in tbc that is.

HanFox: Ah, I see. I'll test that in a bit.

The [Mend Pet](http://wow.gamepedia.com/Mend_Pet) spell in Cata up to WoD is no longer a channel (it is a channel again in Legions). [Improved Mend Pet](http://wow.gamepedia.com/Improved_Mend_Pet) itself was removed in MoP, but was altered in TBC to a 25%/50% chance.

I would guess a channel spell works differently compared to a HoT. I haven't looked.

HanFox:
Ah, I see https://github.com/cmangos/mangos-classic/pull/222/commits/1a74adb07b0d14bba0f7ab48958b44cddd0c25a9 just makes sure true is returned at the right time, but doesn't fix the visual issues for the client which is what I meant by undesirable.

Basically what needs to happen is that it checks for an ailment on the pet to dispel. If there's something to dispel it dispels it and the client is told so to do what it needs to do, but if there's nothing to dispel the client does nothing at all (i.e. no error sound/message).

Currently [SpellCastResult Spell::CheckCast](https://github.com/HanFox/mangos-classic/blob/bd145ef0f498f67f7ff7a47c18fc0c0769e851f3/src/game/Spell.cpp#L3926) requires a return and all of these returns seem to have a noticeable bit of feedback. It appears that in the client any false return will make an error noise/message and a true return visually shows the spell being cast.

I'm not sure if any other spells have this sort of passive chance to them on ticks, so I'm not sure if there's already an implementation for circumventing the "erroneous" behaviour.

HanFox:
https://github.com/cmangos/mangos-classic/pull/222/commits/f67dfb53a0b8f63830727458af9b505eb8112a7a is a bit convoluted so it may need tidying somehow, but in the current state it works as expected.

Phatcat:
> but if there's nothing to dispel the client does nothing at all (i.e. no error sound/message).

You sure about that? because that's opposite behaviour of a spell such as Abolish Poison;
https://youtu.be/hHqSBcJRwTk?t=1m46s
(pay attention to the error message ticking in)

src/game/SpellAuras.cpp
```diff
@@ -4396,20 +4396,17 @@ void Aura::PeriodicTick()
...
+                        triggerAmount = 15; //rank 1
 
-                        if (pCaster->HasSpell(19573))
-                            triggerAmount = 50; //rank 2
+                    if (pCaster->HasSpell(19573))
+                        triggerAmount = 50; //rank 2
 
-                        if (roll_chance_i(triggerAmount))
-                            pCaster->CastSpell(target, 24406, true, nullptr, this);
+                    if (roll_chance_i(triggerAmount)) {
```
Phatcat: What's { doing at the end of that line? Also, why is the brackets needed at all on that single liner?

src/game/Spell.cpp
```diff
@@ -3185,6 +3185,11 @@ void Spell::SendCastResult(Player* caster, SpellEntry const* spellInfo, SpellCas
 {
     WorldPacket data(SMSG_CAST_FAILED, (4 + 1 + 1));
     data << uint32(spellInfo->Id);
+    
+    if (spellInfo->Id == 24406 && result == SPELL_FAILED_NOTHING_TO_DISPEL) // Improved mend pet - don't send anything to player if nothing to dispel
```
Phatcat: This check, being done here, that's not gonna fly. Move it back to checkcast somewhere.

HanFox:
Yes, but [Abolish Poison](http://www.wowhead.com/spell=14253/abolish-poison) is an instant cast which is purposefully trying to clear a debuff.

Improved Mend Pet is not actively trying to clear a debuff, it's passively trying to do it on the heal ticks.

If you have 5 ticks, and each one succeeds to try to debuff then you'll get 5 (1 per second) "Nothing to dispel" messages which is tantamount to spam.

> Gives the Mend Pet spell a 50% chance of cleansing 1 Curse, Disease, Magic or Poison effect from the pet each tick.

src/game/Spell.cpp
```diff
@@ -4256,7 +4261,7 @@ SpellCastResult Spell::CheckCast(bool strict)
             for (Unit::SpellAuraHolderMap::iterator iter = Auras.begin(); iter != Auras.end(); ++iter)
             {
                 spell = iter->second->GetSpellProto();
-                if (spell->Dispel == mechanic)
+                if (spell->Dispel == mechanic || (m_spellInfo->Id == 24406 && (spell->Dispel == DISPEL_MAGIC || spell->Dispel == DISPEL_CURSE || spell->Dispel == DISPEL_DISEASE || spell->Dispel == DISPEL_POISON)))
```
Phatcat: Yeah, this right here does not look well... It looks a bit like a hardcoded hackfix, wouldn't you say? =/ - Not saying there's any prettier way for us to check it by what we know currently, but still..

Phatcat:
> Yes, but Abolish Poison is an instant cast which is purposefully trying to clear a debuff.

Not true, it's a spell which will apply a buff on the user, that buff will attempt to remove a poison every.. 2 or 3 seconds, don't remember.

the point is that neither of these spells ought to fail as a cast even if the target isn't affected by anything, it should however fail the periodic casts if there's nothing to cure, this they seem to have in common these 2 spells (imp mend pet, abolish poison)
Is there any proof that when the periodic casts fail for imp mend pet that the error message should actually be silenced, or that the cure effect shouldn't just be cast even if there's nothing to cure? (like with the initial cast of abolish poison)

HanFox:
Oh, I didn't look at the vanilla spell, oops. Keep doing that, sorry.

I've failed to find a source to verify how it works. Most the videos from then are basically potato quality so I can't read messages or just can't discern what's really happening and none provide builds anyway.

I suppose the consensus would likely be it'd be the same for all spells.

So how should it be in your opinion on a spell that can have a 50% chance of doing something every second (over 5 seconds)?

The way I have it now is just what I'd personally expect from it (that you only get feedback when it actually dispels something). If I had to change it I'd rather it just did the dispel effect w/o any failed error messages.

Phatcat:
actually, it shouldn't give any error message at all, I just noticed now that each time it gives an error that's him right clicking the macro...

gonna fix that in a moment, then you don't have to worry about it, it will only cast if there's something to dispel, otherwise it will silently fail, like other triggered spells.

Phatcat:
> it'd be desirable if it removed the "Nothing to dispel" message.

Should be done in https://github.com/cmangos/mangos-classic/commit/e6b34b21764440a0ce316b766e4dfcc9790056dd

Phatcat:
Wait... that was a bit fast on the trigger. I have no idea if that commit fixes issue 306, I doubt it very much in fact. I was only refering to the 'nothing to dispel' message. That part should be taken care of.

HanFox:
Accidentally destroyed my branch. >.>

Didn't mean to.

Phatcat:
Effects
Effect 0:
Id: 6 (SPELL_EFFECT_APPLY_AURA)
BasePoints = 15
Targets (1, 0) (TARGET_SELF, NO_TARGET)
Aura Id 112 (SPELL_AURA_OVERRIDE_CLASS_SCRIPTS), value = 15, misc = 4086 (4086), periodic = 0
EffectItemType = 0x00800000
+ 136 - Mend Pet (Rank 1)
+ 3111 - Mend Pet (Rank 2)
+ 3661 - Mend Pet (Rank 3)
+ 3662 - Mend Pet (Rank 4)
+ 13542 - Mend Pet (Rank 5)
+ 13543 - Mend Pet (Rank 6)
+ 13544 - Mend Pet (Rank 7)
Effect 1:
NO EFFECT
Effect 2:
NO EFFECT

> Aura Id 112 (SPELL_AURA_OVERRIDE_CLASS_SCRIPTS), value = 15, misc = 4086 (4086), periodic = 0

Looks like the effect of the aura has to be scripted, but look at that, you get both the spell it's supposed to cast (4086) and how much chance (15% - since this is rank 1 I am looking at the info from)

Phatcat:
24406 is the spell to cast of course, as you do in your current code.

@Warlockbugs you know anything about scripted class auras..?

HanFox:
@Phatcat is this clean enough? It only casts on Mend Pet. It replaces "does the caster have this spell" entirely (I'm not sure about the for loop really, but like this it currently works):

`Unit::AuraList const& aurasOverrideClassScripts = pCaster->GetAurasByType(SPELL_AURA_OVERRIDE_CLASS_SCRIPTS);
				for (Unit::AuraList::const_iterator iter = aurasOverrideClassScripts.begin(); iter != aurasOverrideClassScripts.end(); ++iter)
				{
					triggerAmount = (*iter)->GetModifier()->m_amount;
				}
                if (triggerAmount > 0 && roll_chance_i(triggerAmount))
					pCaster->CastSpell(target, 24406, true, nullptr, this);`

HanFox:
I also think the `spellProto->SpellFamilyFlags` still needs to have a bitwise check on a some sort of uint64? I'm a bit lost with that though.

robinsch:
All of this can be done by spell_proc_entry.

HanFox:
Then do it in there.

PR #187, opened 2016-08-20
Hunter Casting Times

dagochen:
Current situation (issues which are fixed with this PR):
Multishot casting time is not improved by rapid fire, quick shots or anything else.
Aimed Shot casting time is improved by quiver.
Aimed Shot casting time is decreased by 0,5 seconds for no reason.

What does this PR do?
Aimed shot must not profit from quiver (any hunter guide, any youtube video...) and the casting time is set to 3 seconds. It improves correctly by rapid fire, same as the multi shot.

Proof:
https://youtu.be/ACM3yXuz0Hw?t=1m30s
Aimed shot needs 3 seconds with quiver equipped

Spelleffects: 
Quiver: 14829 - Apply Aura: Haste - Ranged (Ammo Only) 
Rapid Fire: 3045 - Apply Aura: Haste - Ranged
Quick Shots: 6150 - Apply Aura: Haste - Ranged

Any hunter guide who says that Ashjre'tul is best, because of the slow weapon speed. If Aimed shot would get bonus from quiver, all ranged weapons with higher attack rate would be better.

PR #223, opened 2016-11-22
Feature casttimefix

dagochen:
redone the fix from last PR, with little improvements

since everybody ignored the first PR, this might also be something for @Warlockbugs

### DIFF
--- a/src/game/SpellMgr.cpp
+++ b/src/game/SpellMgr.cpp
@@ -26,6 +26,8 @@
 #include "Spell.h"
 #include "Unit.h"
 #include "World.h"
+#include "SpellAuras.h"
+
 
 bool IsPrimaryProfessionSkill(uint32 skill)
 {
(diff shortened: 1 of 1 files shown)


Warlockbugs:
I am afraid, we will have to start from the beginning on the issue then.
A correctly described problem usually contains the solution within description. 😄 

Original PR link: https://github.com/cmangos/mangos-classic/pull/187

> Aimed Shot casting time is decreased by 0,5 seconds for no reason.

There is no such thing as no reason. 😄 Debugging is needed. DBC entry is set to 3s, so this is core shenanigans.

Now lets move on to ammo issue.

You mentioned that Aimed Shot and Multi-shoot should not receive cast time bonus from ammo. But what about other shots, including those which came with newer expansions (Steady, etc)?

src/game/SpellMgr.cpp
```diff
@@ -120,23 +122,36 @@ uint32 GetSpellCastTime(SpellEntry const* spellInfo, Spell const* spell)
 
     int32 castTime = spellCastTimeEntry->CastTime;
 
+    // Multishot only (alternative set CastingTimeIndex = 3)
+    if (spellInfo->HasAttribute(SPELL_ATTR_RANGED) && (!spell || !spell->IsAutoRepeat()) && spellInfo->SpellFamilyFlags & 0x0000000000001000) 
+        castTime = 500;
```
dagochen: yes

dagochen:
Oh did I wrote decreased? Ofcourse its an increase of 0.5 seconds...someone wanted to increase multishot and did that for EVERY ranged spell. 

And about the steady shot thing I am not sure, because its TBC and then all this mechanics work completely different, considering delay of autoshots and haste on items and so on....

Warlockbugs:
I need it to know if all shots (i.e. ranged weapon abilities which use ammo), do not benefit from quiver's speed. In this case there is a fairly straightforward way to handle it.

dagochen:
You propably need to check for the spell effects on haste-gear in tbc/wotlk and other spells like Heroism, which improved casting time for steady shot. If there is still a difference between "Apply Aura: Haste - Ranged (Ammo Only)" (quivers in classic) and "Apply Aura: Haste - Ranged" (non-quiver-effects), then this fix should work fine.

dagochen:
And later on quivers have been removed, which might help too?

cyberium:
Only part of this is merged. The rest need more research/work/proof.

PR #224, opened 2016-11-22
Make all projects compile with MP and add EditAndContinue build config

killerwife:
@namreeb


cyberium:
This in not needed now

PR #226, opened 2016-11-26
Implement TriggerCastFlags

killerwife:


Phatcat:
seriously, what? up for 1 hour, how does that allow for any kind of testing or anything?

src/game/AI/CreatureAI.cpp
```diff
@@ -142,7 +142,9 @@ CanCastResult CreatureAI::DoCastSpellIfCan(Unit* pTarget, uint32 uiSpell, uint32
             // Creature should stop wielding weapon while casting
             pCaster->SetSheath(SHEATH_STATE_UNARMED);
 
-            pCaster->CastSpell(pTarget, pSpell, !!(uiCastFlags & CAST_TRIGGERED), nullptr, nullptr, uiOriginalCasterGUID);
+            uint32 flags = (uiCastFlags & CAST_TRIGGERED ? TRIGGERED_OLD_TRIGGERED : TRIGGERED_NONE) | (uiCastFlags & CAST_IGNORE_UNSELECTABLE_TARGET ? TRIGGERED_IGNORE_UNSELECTABLE_FLAG : TRIGGERED_NONE);
```
Phatcat: This right here goes to show it's never been about readability but rather about hating on me and my code in the past. !(A || B) is hard to grasp but this right here is clear as day,. eh?

Laizerox: That is not a || b its flags = flag1 | flag2

Phatcat: .... My point is not about the logic, my point is about

I know the former is a bool operation and the other is a choise between one value or the other. I am talking about readability, though, not the logical operation behind it.

https://github.com/cmangos/mangos-wotlk/blob/master/src/game/Creature.cpp#L2114 <- that is clear and readable.

This here looks like a select nested inside another select, but then when you look closer it's not, it's still however two selections nested inside something else that basically works a selection anyways.

cyberium: @Phatcat is right WE(nicluding me) should avoid line like that. Iam sure its not easly readable even for you @killerwife :)

However its only a style issue so instead of writing those comment better would be to fix it ;)

PR #229, opened 2016-12-04
Don't show gossip menu when learning new flight paths

robinsch:
First click on flightmaster should only show up "New flight path learned" (or whatever) and not open the gossip window.

### DIFF
--- a/src/game/GossipDef.cpp
+++ b/src/game/GossipDef.cpp
@@ -28,6 +28,7 @@ GossipMenu::GossipMenu(WorldSession* session) : m_session(session)
 {
     m_gItems.reserve(16);                                   // can be set for max from most often sizes to speedup push_back and less memory use
     m_gMenuId = 0;
+    m_gIgnoreGossip = false;
 }
 
 GossipMenu::~GossipMenu()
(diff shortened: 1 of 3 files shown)


killerwife:
@robinsch One specific question. There is a specific flight master at which you have both gossip and fly path. Stair of Destiny, in hellfire peninsula. Question is, do you do "fly path learned, and then on next click gossip/fly path" or do you do "fly path learned and show available gossip, and on next click fly path as well"

This applies to any flight master which has any "deliver XYZ to city xyz", before you use him to acquire his flight path.

robinsch:
This is classic - I can't think of a example in classic.

killerwife:
I can, what if NE/Dwarf level in westfall and dont have flight path in westfall prior to picking up the "travel to Stormwind" quest.

robinsch:
Then the client should remove the flightmaster flags from NPC which is another issue handled by https://github.com/cmangos/mangos-classic/pull/228

robinsch:
wops wrong button

hlarsen:
works for me; not sure if it's a bug or not but if you have a quest to turn in to the flight master the ? over Thor's head disappears after you get the flight path.

repro:
```
.go taxinode 4
.quest add 6181
see ? disappear
```

PR #230, opened 2016-12-04
Implement TRIGGERED_IGNORE_LINE_OF_SIGHT

robinsch:
This is used for NPCs that reward a spell cast but are inside of gameobjects. For example NPCs in a cage.

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -11957,7 +11957,7 @@ void Player::RewardQuest(Quest const* pQuest, uint32 reward, Object* questGiver,
                 }
             }
 
-            caster->CastSpell(this, spellProto, TRIGGERED_OLD_TRIGGERED);
+            caster->CastSpell(this, spellProto, TRIGGERED_OLD_TRIGGERED | TRIGGERED_IGNORE_LINE_OF_SIGHT);
         }
     }
 
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -315,6 +315,7 @@ Spell::Spell(Unit* caster, SpellEntry const* info, uint32 triggeredFlags, Object
 
     m_ignoreHitResult = false;
     m_ignoreUnselectableTarget = m_IsTriggeredSpell;
+    m_ignoreLineOfSight = false;
 
     // determine reflection
     m_canReflect = false;
@@ -323,6 +324,9 @@ Spell::Spell(Unit* caster, SpellEntry const* info, uint32 triggeredFlags, Object
 
     if (triggeredFlags & TRIGGERED_IGNORE_UNSELECTABLE_FLAG)
         m_ignoreUnselectableTarget = true;
+
+    if (triggeredFlags & TRIGGERED_IGNORE_LINE_OF_SIGHT)
+        m_ignoreLineOfSight = true;
 
     CleanupTargetList();
 }
@@ -6180,7 +6184,7 @@ bool Spell::CheckTarget(Unit* target, SpellEffectIndex eff) const
             // Get GO cast coordinates if original caster -> GO
             if (target != m_caster)
                 if (WorldObject* caster = GetCastingObject())
-                    if (!target->IsWithinLOSInMap(caster))
+                    if (!m_ignoreLineOfSight && !target->IsWithinLOSInMap(caster))
                         return false;
             break;
     }
--- a/src/game/Spell.h
+++ b/src/game/Spell.h
@@ -382,6 +382,7 @@ class Spell
         // Trigger flag system
         bool m_ignoreHitResult;
         bool m_ignoreUnselectableTarget;
+        bool m_ignoreLineOfSight;
 
         int32 GetCastTime() const { return m_casttime; }
         uint32 GetCastedTime() const { return m_timer; }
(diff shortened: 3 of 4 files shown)


cyberium:
@killerwife but look ok to me

robinsch:
I would rather just save triggeredFlags in Spell instead of creating new members for the check itself.

killerwife:
@robinsch which spell is it needed for? Because I found out that there are two special relationships for LOS, 50k yard spells ignore LOS by default, and then there is an attribute for it.

robinsch:
None of both, Questgiver: 3430 a example spell would be 16612. it's failing because the NPC is inside of a GO.

killerwife:
Good, that is the same example I came into contact with. I think that spell should globally ignore LOS instead of needing triggered cast flag. That being said, I do not oppose this PR, I just want to avoid using local settings instead of global settings where applicable.

robinsch:
I think the check is needed for AOE spells, other spells checking LOS on cast etc.

killerwife:
Check for what?

robinsch:
Spell::CheckTarget for line of sight.

killerwife:
Yes well, I know of a certain individual who can send simulated cast spell packets despite mob being in LOS. So its there to prevent cheating.

robinsch:
> I think that spell should globally ignore LOS instead of needing triggered cast flag

But adding a attr won't solve a thing. This is targeted for all quest giver spells. Or do I get you wrong there?

killerwife:
I was questioning whethere we are supposed to do it this way, or researching each spell it fails at and adding a rule for it (attribute, or override). Anyway, I will think about it, but first I will export some LOS related code, because 2 days ago I found out one of my PRs was not merged correctly, resulting in me not being able to PR my LOS rule I explained above.

robinsch:
What comes into my mind is to disable the entire check for TYPEID_UNIT caster (exclude pets). All LOS related checking should be done in this case before this.

killerwife:
Yes but it seems they did use attributes correctly on creature only spells, but they tend to be lazy with exporting it to client on that front, meaning that it is likely that they just intentionally didnt give us those. So still, your fix is probably good for our framework, but it might diverge from official logic. Maybe in the future I will take a gander at researching all the cases :)

robinsch:
Those attributes should be checked somewhere else. I feel like this check is just there for your case (the cheating one) and should be removed for non player related units at all.

killerwife:
Oh, you missed one LOS check in CheckCast, there are two :)

robinsch:
No, as described here https://github.com/cmangos/mangos-classic/pull/230/files#diff-e4e3db61fb7500a6206a995b963d79faR318 it's only for the CheckTarget

But my point is that I'm pretty sure this entire check should be removed for non player case. Anyone giving there thought on this?

killerwife:
https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spell.cpp#L4366

robinsch:
I don't really get what you want.

killerwife:
That you did not add the condition to all LOS checks in checkcast.

robinsch:
It not needed there because it won't check in any case because of if (!m_IsTriggeredSpell

killerwife:
That is really really bad design. I am trying to move cmangos away from misusing triggered in some cases. That was the whole reason for introducing TriggeredFlags in the first place. To enable researching each case visibly, and in the meantime give the scripting team more capabilities.

robinsch:
Why should I add a check to something that will never be reached. How is that bad design. This is not something that should be adressed in this PR. This PR was only created to give something back to the community. (The original fix was done for a complete diffrent codebase)

killerwife:
It is bad design to begin with. I.e. we do not ignore LOS for some spells but we ignore it for others using m_isTriggered. I suggest you look how I implemented TRIGGERED_IGNORE_UNSELECTABLE_FLAG. As far as my LOS research goes, there are no double rules for some spells over others, the way we are using m_isTriggered.

That means instead of putting your "ignore los" check only in one place, instead populate the boolean using TRIGGERED_OLD_TRIGGERED first, and then remove m_isTriggered check and replace it with m_ignoreLineOfSight. In addition to that, gotta add ignore los to EAI interface cast flags so the opportunities are equivalent for scripting DB vs C++.

Please keep in mind I am very well aware that you do not primarily work on cmangos. I am only properly reviewing it, since @cyberium tagged me. I am just pointing out logical flaws specifically relative to CMaNGOS code behaviour. There is some logic we are trying to move towards, that is why I am telling you in the first place.

robinsch:
Yeah no worries. I just have a diffrent mindset or this might be really related to code behaviours. 

But this might be obsolete because the entire check seems wrong to me.

liwanfu:
December BUG：
1、Rogue stealth pickup BUG-Remove stealth
2、mind soothe and Soothe Animal BUG-not reduce the warning range
3、 Hunter and Warlock pet BUG-level HP wrong
4、event Arena Master and Fishing competition error
5、Lupos BUG NO Shadow Strike
6、 Snarler BUG NO 100 resistance
7、Rak'shiri BUG NO 3 Claw
8、 Mind vision BUG error

cyberium:
Implemented another way (if iam not wrong)

PR #232, opened 2016-12-04
Minimum speed rate is 0.01f

robinsch:
Confirmed by sniffs (but 2.3) and still valid in 6.x

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -7155,8 +7155,7 @@ struct SetSpeedRateHelper
 
 void Unit::SetSpeedRate(UnitMoveType mtype, float rate, bool forced)
 {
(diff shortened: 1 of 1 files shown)


killerwife:
This means that mobs which do not move at all, for example fel cannons, still have at least 0.01f?

robinsch:
Those should be rooted?

killerwife:
We do it by turning off combat movement. I dont particularly know, just asking.

robinsch:
I think in your case it's handled by some root flag. I don't think those are sending movementpackets (except for turning). The lowest speedOpcodes << speed can be is 0.01f

killerwife:
Fair enough, thx for explaining.

PR #234, opened 2016-12-06
[Spell] Wand shoot should not proc anything, e.g. not proc Arcane Con…

ryancheung:
…centration

The spell prod flags in dbc includes range hit proc, but I think this should be wrong.

### DIFF
--- a/src/game/SpellMgr.cpp
+++ b/src/game/SpellMgr.cpp
@@ -1099,6 +1099,10 @@ bool SpellMgr::IsSpellProcEventCanTriggeredBy(SpellProcEventEntry const* spellPr
     // No extra req need
     uint32 procEvent_procEx = PROC_EX_NONE;
 
+    // wand should not proc anything
+    if (procSpell && procSpell->Id == 5019)
+        return false;
+
     // check prockFlags for condition
     if ((procFlags & EventProcFlag) == 0)
         return false;


killerwife:
This is a hack. Should be done in DB. Though spellmod system requires rework which I have done but first I need to merge my other 13 PRs.

cala:
I don't fully agree. Wand shoot should not proc arcane concentration, yes. But it should still proc some other spells like the Aura of the Blue Dragon from the Darkmoon Fair card and probably other ones.

dagochen:
@cala Darkmoon Fair Card Blue Dragon: This item will no longer trigger from using any ability which does not cost mana.
http://wowwiki.wikia.com/wiki/Patch_1.11.0

cala:
Thanks for pointing that out. I was no longer using the card when Naxxramas was released, but I made an intensive use of it in prior encounters, mostly long ones like Nefarian.

However, I'm still not sure that wand should not proc anything.

killerwife:
Well, I introduced a simple rule - only proc SpellMod if SpellMod was consumed by given spell, and it fixed the issue on our end. Hence I do not think this PR is needed at all, since that one rule eliminated whole issue. I can give you guys a diff for the feature if u want, but I do not see it being merged within the next month, due to too many upcoming changes.

liwanfu:
December BUG：
1、Rogue stealth pickup BUG-Remove stealth
2、mind soothe and Soothe Animal BUG-not reduce the warning range
3、 Hunter and Warlock pet BUG-level HP wrong
4、event Arena Master and Fishing competition error
5、Lupos BUG NO Shadow Strike
6、 Snarler BUG NO 100 resistance
7、Rak'shiri BUG NO 3 Claw
8、 Mind vision BUG error

cyberium:
If this should be fixed it should not this way.

PR #238, opened 2016-12-11
Migrate spell.dbc from DBC to SQL

killerwife:
@cyberium

Disclaimer: This is the only branch I did not test to run, since I do not have classic setup.

### DIFF
--- a/src/game/AI/CreatureAI.cpp
+++ b/src/game/AI/CreatureAI.cpp
@@ -110,7 +110,7 @@ CanCastResult CreatureAI::DoCastSpellIfCan(Unit* pTarget, uint32 uiSpell, uint32
     // Allowed to cast only if not casting (unless we interrupt ourself) or if spell is triggered
     if (!pCaster->IsNonMeleeSpellCasted(false) || (uiCastFlags & (CAST_TRIGGERED | CAST_INTERRUPT_PREVIOUS)))
     {
-        if (const SpellEntry* pSpell = sSpellStore.LookupEntry(uiSpell))
+        if (const SpellEntry* pSpell = sSpellTemplate.LookupEntry<SpellEntry>(uiSpell))
         {
             // If cast flag CAST_AURA_NOT_PRESENT is active, check if target already has aura on them
             if (uiCastFlags & CAST_AURA_NOT_PRESENT)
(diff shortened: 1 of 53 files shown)


cala:
Unless I'm mistaken, this was merged in https://github.com/cmangos/mangos-classic/commit/e72ad52fc983309be6f726cd094f153456dcc8d3 so I just have a few questions before closing this.

I followed this thread from afar, so forgive me if I ask some stupid or already answered questions:
- what does this change means for DB providers: will they now be in charge of the spells integrity or is this still handled by core developers? Stated otherwise: who is in charge of filling the new table and checking its content is correct (we know that some DBCs values are wrong or at least we are using them wrong or both)?
- does this change allow more power to DB providers or SD2 devs to fix/script some weird or dummy spells required in events/encounters/quests? Or is the workflow still the same?

In any case, thanks @killerwife for this work! 👍

killerwife:
Ok @cala let me explain the whole problematic.

Main reason to do this is to fix mistakes that had to have been hacked in C++, whereas in SQL it will be a single change. The client spell.dbc contains a lot of them, some of which are clientside and we cant fix without repacking client (which is a big legal no no). Thats why we will be able to fix issues like dummy, that needs to be periodic, not having timer, or having wrong amounts of MaxAffectedTargets, this is for example case for Magtheridon or Prince Malchezaar in TBC (sorry i dont have vanilla examples)

Then there are missing spells removed over the expansions. Namely darrowshire event, or Plagued Lands has a few or Racial Stoneform, which I even found in 2.4.3 sniff to be sent to client, even though its not in client DBC. This enables us to add them in their full glory.

Problem with all of this is that some columns must not be changed at any cost, because it might cause client to crash. Effect1-3 or EffectApplyAuraName1-3 are a big big 'dont go there'.

Also the idea is to keep it core-side after a very long discussion I had with @cyberium 2 months ago. The point was that spell_template contains data needed to run an empty world, hence should be in core. Also, i dont mean to be disrespectful to DB developers, but as of now very few of you actually have a clue how spells work in c++, hence it should be restricted for them. Since you have core push rights, its not an issue for you.

And lastly, some of these changes will be expansion specific, so we need to be careful when porting them up or down, because there might be a new replacement for such spell (naxxramas comes to mind).

I am working on documenting what each column does on our wiki, but im kinda too busy to finish that. Sometime when my PRs are merged, I will make a "core cleanup" PR where I will remove most/if not all core hacks that were done because we could not change the DBC. Also, please keep all changes to the separate query file with comments, so in retrospect we know what and why it was changed.

cala:
Thank you for taking the time to write such a complete answer, I appreciate it. 👍 
This answers my questions and I agree with the workflow and roles you described. 😄 
Nice work!

I'm closing the PR as it was merged, @cyberium can still reopen it and punish me if I missed something.

PR #240, opened 2016-12-16
Improve Spell Effect SPELL_EFFECT_TELEPORT_UNITS_FACE_CASTER

cala:
When radius is zero, players should be ported at the caster location
without using bounding radius (neither for caster nor target).
Exemple: Azuregos teleport spell 21150 triggered by spell 21147: all
players are to be ported _under_ Azuregos, not close to him

Also clean some whitespaces in SpellEffects.cpp

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -946,7 +946,7 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                         SpellModifier* mod = new SpellModifier(SPELLMOD_RESIST_MISS_CHANCE, SPELLMOD_FLAT, damage, m_spellInfo->Id, uint64(0x0000000000000100));
                         ((Player*)unitTarget)->AddSpellMod(mod, true);
                     }
-                    
+
                     break;
                 }
                 case 12472:                                 // Cold Snap
@@ -1208,7 +1208,7 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                     int32 bonusDamage = m_caster->SpellBaseDamageBonusDone(GetSpellSchoolMask(m_spellInfo))
                         + unitTarget->SpellBaseDamageBonusTaken(GetSpellSchoolMask(m_spellInfo));
                         // Does Amplify Magic/Dampen Magic influence flametongue? If not, the above addition must be removed.
-                    float weaponSpeed = float(m_CastItem->GetProto()->Delay) / IN_MILLISECONDS;     
+                    float weaponSpeed = float(m_CastItem->GetProto()->Delay) / IN_MILLISECONDS;
                     bonusDamage = m_caster->SpellBonusWithCoeffs(m_spellInfo, bonusDamage, 0, 0, SPELL_DIRECT_DAMAGE, false); // apply spell coeff
                     int32 totalDamage = (damage * 0.01 * weaponSpeed) + bonusDamage;
 
@@ -1486,7 +1486,7 @@ void Spell::EffectTeleportUnits(SpellEffectIndex eff_idx)   // TODO - Use target
     // post effects for TARGET_TABLE_X_Y_Z_COORDINATES
     switch (m_spellInfo->Id)
     {
-        case 23441:                                 // Ultrasafe Transporter: Gadgetzan 
+        case 23441:                                 // Ultrasafe Transporter: Gadgetzan
         {
             // Wrong destination already rolled for, only handle minor malfunction on sucess
             m_caster->CastSpell(m_caster, 23450, TRIGGERED_OLD_TRIGGERED); // Transporter Arrival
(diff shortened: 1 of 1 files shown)


Rushor:
nice!

GitHub Logo

cala:
I only test this on Classic, but you are right it should also fix the similar spell added to the Emerald Dragons in patch 2.0

Rushor:
don't worry, i only test it with a dragon and ".cast back" command :D

cala:
Kind poke @killerwife to see if he can think of some spell/NPC that could be broken by this change.
If not, I should merge this soon.

liwanfu:
You foreign friends of God, I have two BUG now. Please help me deal with, I come from china. 
第一个BUG是牧师安抚心灵和安抚动物，添加了使用安抚心灵不造成仇恨以后，但是根本无法减少怪物警戒范围，核心是12月份的。
The first BUG is the priest soothe the mind and soothe the animal, adding the use of soothe the soul does not cause hatred later, but simply can not reduce the scope of the monster alert, the core is December. 
第二个BUG是盗贼的潜行已经修复了强化闷棍和消失，但是在潜行状态下，拾取物品会破除潜行。
Second BUG is a rogue stealth has been repaired and improved SAP disappeared, but in stealth, pick up items will get rid of stealth. 
第三个BUG是12月份的核心发现宠物升级以后血量的都和原来的不对应的，比如600血的宠物，升级以后只有300血，需要解散再召唤出来才能正常。
The third BUG is the core of December found that pet upgrade output and the original non corresponding, such as 600 blood pet upgrade only 300 blood needs to be dissolved again summoned to normal.

liwanfu:
请给我直接修复的补丁，因为我是新手，不会C+。另外希望求一份高修复的技能DBC，目前12月份的核心宕机问题很多，官网上面提供的补丁实在是很少。

liwanfu:
ase give me direct fixes because I am a novice and will not C+. In addition to seek a high repair skill DBC, many of the core problem of downtime in December the official website provided above patch is really few.

liwanfu:
Thank you very much, I also need to repair the other 2 tasks, one is the core task of the masters and the 1.12 classic arena Stranglethorn fishing contest, source code and data repair also want to find 1 in the battle, begged the patch. The core problem of feedback on the 1 December, if the use of biological rejuvenation will cause downtime, but also hopes to help repair. I also have 1 tasks is to hang black head, can not display the skull model Nefarian, should be the database inside what is missing. There are no hunters ancient leaf task specific event trigger, the official should be the 1 event, the hunter received the task of standing on the hill above will show 3 trees NPC phantom, at present I have only added to the 3 NPC.

liwanfu:
Arena Master/e battle of Ahn'qiraj /Fishing competition/Ancient stone leaf event triggered
Patches that need to be repaired

liwanfu:
(BUGS)Rogue stealth pickup/mind soothe Will not reduce the warning range /Warlock and Hunter for pet  After the upgrade of the volume is not correct -HP/creature The use of rejuvenation  Downtime;Crash /The priest mind is not valid/Soothe Animal  bug

liwanfu:
BUG boss_jeklik.cpp Summon ZG leopard leopard marker BOSS game player after the game player will not take the initiative to attack.

liwanfu:
Hunter pet  Lupos No shadow attack after taming 
Hunter pet  Snarler  No 100 resistance after taming 
Hunter pet  Rak'shiri  No three Combo 
Please help repair the

liwanfu:
Hunter pet  Lupos No shadow attack after taming 
Hunter pet  Snarler  No 100 resistance after taming 
Hunter pet  Rak'shiri  No three Combo 
Please help repair the

Rushor:
what is this guy talking about?

liwanfu:
Please help me.

liwanfu:
December  BUG：
1、Rogue stealth pickup    BUG-Remove stealth 
2、mind soothe  and  Soothe Animal         BUG-not reduce the warning range     
3、 Hunter and Warlock  pet                     BUG-level HP wrong
4、event    Arena Master  and  Fishing competition    error
5、Lupos  BUG NO Shadow Strike 
6、 Snarler BUG NO  100 resistance
7、Rak'shiri   BUG NO 3 Claw
8、 Mind vision  BUG error

killerwife:
@cala due to this spam i did not notice your question. The answer is that it influences two spells, Gravity lapse on Kaelthas and Teleport Maiev on Ilidan, but I havent even looked at those encounters yet. From me its good to go, I have seen many instances where we port things into vicinity instead of right under.

xfurry:
I think that the problem is with the model boundaries of the caster. Probably the spell effect doesn't consider that, and that's why the players are teleported directly under the caster unit.

Rushor:
but the fix was intended to directly port the victim unter the unit and NOT in front of him @xfurry

liwanfu:
Cmangos My friends, please help repair the next BUG, thank you.

cala:
@liwanfu will you stop spamming the commit/PR messages please? We have [an issue tracker](https://github.com/cmangos/issues/issues). Please use it and open your issue there and only one per issue. Thanks.
I'm cleaning the out of place messages.

cala:
@killerwife thank you for the feedback. I'll have a look at those two encounters when I can get the chance. If they are working as they should, I'll merge this.

@xfurry this PR is in fact fixing the opposite. In its current state, the spell effect uses the caster boundary, the target boundary and the effect radius to calculate the coordinates. This is correct when the target is expected to be ported in front/near the caster. But there are some spells where the targets must be ported at the caster location without using boundaries. This behaviour is not handled by the current spell effect, so this PR adds this case when the effect radius is zero.

liwanfu:
BUG:
 1、Insert the totem should be destroyed after no player experience
 2、The mage's extreme cold extension caused a blizzard to slow down。
3、The hunter's beast precepts have no practical effect
 4、Master of the Ice Shards talent no actual effect
 5、No shadow attack effect of biological Perth tamed wolf Lu
 6、The growl has no resistance to 100 after taming

PR #242, opened 2017-01-09
WIP: adds a somewhat fair WorldQueue

MrMarvin:
**This MR introduces a new feature:** Letting player who disconnect while in world queue get back to their last position, instead of being put at the end of the queue.

Although this seems somewhat cosmetic for most realms, sitting in a 11k player / 6 hours+ queue is not fun...

If a player disconnects during her client waiting in the queue
to enter the world we now keep the connection alive for some time.
If she reconnects within the logOut time, the old session is replaced
and the client enters the queue at the position it was before.
The default time to reconnect is 0 seconds, which mimics the current behavior.

**There is a weird (at least for me, first time c++ ...) bug left which leads to nasty segfaults when updating queue positions. This happens in around 1/2 of my manual tests. Will have a look into it later, so marking this MR as WIP for now. Any help or comments greatly appreciated!**


SoulSharer:
Concerning segfault, possible race condition?

dagochen:
Is that just for the login queue or does it also affect players, who are in the game and running around?

cyberium:
I'll work on that when i'll introduce grace period for player reconnect.

cyberium:
Should be implemented

PR #244, opened 2017-01-15
Children's Week Orphan Quests Support

Zingzah:
This PR fixes the issue of players being unable to complete orphan
discovery quests for Children's Week.  The current code uses spells from
Wrath to award quest credit which obviously aren't available in classic
or TBC.  This code removes the use of these spells and adds support for
quest completion.  This change in conjuction with database additions will
allow these quests to be completed in classic (and TBC with subsequent
additons for outland)


cala:
Thanks! I'll review this but looks good at first glance.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/963616fa3a87486973f52b72808d6db15c33247b
Thank you !

PR #246, opened 2017-01-15
TARGET_FLAG_DEST_LOCATION should check for LOS

robinsch:
For example TARGET_ALL_ENEMY_IN_AREA_CHANNELED spells can be placed through obstacles.


cyberium:
merged thanks

PR #249, opened 2017-01-18
Fix player can reset Instances when group members offline.

wowmui:
Now offline will not make players avoiding be teleported to the entrance of dungeon and unlimited kill boss.


wowmui:
@cyberium  You can check this . This isn't a hack.

Muehe:
No idea if the code is correct, but for indentation you should use spaces, not tab-characters.

[Coding Standards - Tab size](https://github.com/cmangos/issues/wiki/Coding-Standards#tab-size)

> First of all, we use space. Tabs are four-character width. That is, no 8-space tabs, no 2-space tabs. Four. Unfortunately there’s no such thing as ‘standard tab width’, and 4-space indenting looks best from our point of view, besides MSVC’ editor has this setting by default.

I agree the fix is correct logic wise though, players should not be able to reset an instance as long as anyone is inside. Getting a new instance would require recreating the group in such a situation (offline player in instance). However, if memory serves, instances could be reset while offline players logged off outside of them. And if I understand your code correctly it doesn't check that?

Warlockbugs:
Kicking offliners and then attempting reset was enough.

wowmui:
For example:
1:Leader out of the Dungeon.
2:Team members offine in dungeon.
3:Leader /run ResetInstances()
4: Leader into dungeons.
5: Team members get online.
6: Team members avoided be teleported to the entrance of dungeon.

This

>  SendResetInstanceFailed(**1**, state->GetMapId());
0 means have player in dungeons error code.
1 means offline error code.
This is a client side msg.

cala:
@cyberium : I remember you did a few commits/fixes about instance map reset and such. If you don't mind, I'm assigning this to you.

PR #250, opened 2017-01-18
fix hunter can't revive dead pet when pet corpse disappeared.

wowmui:


cala:
Confirmed and seems to work. Poking @cyberium for his opinion before I merge this.

PR #251, opened 2017-01-19
Cross faction raid

Hyperorx:


cyberium:
Big thanks for this code unfortunatly we are not for adding more "cross faction" feature. It should need big amount of time to check all the case and even if we want to please all of you with feature request we cant make everyone happy.
So for now i'll close this.

PR #254, opened 2017-01-23
working witch doctor in senjin village

Gacnt:
This enables the Sen'Jin witch doctor "Bom'Bay" to properly do his thing when interacting with him and his gossip options

Gossip options should only appear if the player has completed the 2 required quests

Interacting with any of the gossip options will cause him to cast a randomise spell on you (spells listed here http://www.wowhead.com/npc=10578/bombay#abilities)

This needs to be updated in the database:
```
-- Insert Bom'Bays gossip options
INSERT INTO `gossip_texts` VALUES (-3501000, 'Low spirits', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501001, 'Bad hang nail', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501002, 'Feeling underpowered', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501003, 'Jungle fever', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501004, 'Uni-brow', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501005, 'Whiplash', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
INSERT INTO `gossip_texts` VALUES (-3501006, 'I don\'t want to go back to work', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);

-- Remove gossip_menu from Bom'Bay in creature_template and add script
UPDATE `creature_template` SET GossipMenuId = 0 AND ScriptName = "npc_bombay_witchdoctor" WHERE Entry = 10578;
```


cala:
Thank you for your pull request. It appears it contains more changes than the witch doctor part and unrelated to it. Do you have an issue with merging at some time ?

Regarding the content, there is already a dummy spell that handle the random swith of all the sub-spells: 
https://github.com/cmangos/mangos-classic/blob/master/src/game/SpellEffects.cpp#L615

Thus, I think everything could be handled DB side with a mix of gossip, conditions and DBScripts.

PR #256, opened 2017-02-01
Implement quest 6148

MantisLord:
I would have preferred to handle this quest in database only, but I ran into a snag described more here:
https://github.com/cmangos/issues/issues/1276

The only reason I had to go this route and move Nathanos Blightcaller's AI from ACID to SD2 is to prevent multiple spawns of the creatures over a long distance.
Some SQL from the linked issue is required if this solution is used (waypoints, linking, creature_template changes).

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -270,6 +270,7 @@ UPDATE creature_template SET ScriptName='npc_stinky_ignatz' WHERE entry=4880;
 
 /* EASTERN PLAGUELANDS */
 UPDATE creature_template SET ScriptName='npc_eris_havenfire' WHERE entry=14494;
+UPDATE creature_template SET ScriptName='npc_nathanos_blightcaller' WHERE entry=11878;
 
 /* ELWYNN FOREST */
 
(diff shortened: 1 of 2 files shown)


wowmui:
you are amazing!

killerwife:
@MantisLord If you dont mind, put there a TODO: remove this when generic map wide IsSpawned is implemented.

cala:
I'm not satisfied of using SD2 for something that is *almost* doable with DBScripts, but your argument is perfectly valid. As long as we move this back to DB once the IsSpawned mentioned by @killerwife is implement, I'm fine with this.

I add this on my to do list of the things to review/merge next week.

👍

killerwife:
@cala the mechanic is needed all over the world, for example on Stitches in Duskwood. Problem is its easy to hack this in one way or another, but the ideal way is to make a condition type for this, so that it supports our current DB workflow. And thats a tricky prospect cos it needs to be thread safe and what not. I will try to prototype the feature whenever I have time.

MantisLord:
Closing this...
I believe the decision was made to implement this quest in DB and fix the terminate condition once we have core support for checking if creature is spawned map-wide.

PR #258, opened 2017-02-16
"Gnomish Shrink Ray" fix with roll

catterpiler74:
The author assumed that the roll variable would be assigned with a random value, and then this value would be compared with 3. However, the priority of the comparison operation is higher than of the assignment operation (see the table "Operation priorities in C/C++", so the random number will be compared with 3 first, and then the result of the comparison (0 or 1) will be written to the roll variable.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -469,8 +469,9 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                 {
                     if (unitTarget && m_CastItem)
                     {
+                        uint32 roll = urand(0, 99)
                         // These rates are hella random; someone feel free to correct them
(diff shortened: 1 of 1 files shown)


src/game/SpellEffects.cpp
```diff
@@ -469,8 +469,9 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                 {
                     if (unitTarget && m_CastItem)
                     {
+                        uint32 roll = urand(0, 99)
```
catterpiler74: sorry forget about ";"

Warlockbugs:
If possible, do ammend the second commit into first one and force-push, so it will end up as a single commit. With the recently Github Pull Request merging functionality it will look cleaner than us creating a new commit and then having to close the pull request by hand. :smile:

PR #260, opened 2017-02-19
please check this instance_blackrock_depths

catterpiler74:
The author may have forgotten to put curly braces after the if statement, which caused the call pCreature->SetStandState(UNIT_STAND_STATE_STAND) to be executed regardless of the if condition.

If this behavior was meant intentionally, then the formatting should be corrected:

if (m_auiEncounter[11] == DONE)
  pCreature->SetFactionTemporary(....);
pCreature->SetStandState(UNIT_STAND_STATE_STAND);


cala:
You are right, this is my mistake: bad Python habit (equal indentation = block).
Merged in https://github.com/cmangos/mangos-classic/commit/707d8e7a6a1d34d0dac2b04bb12dce36e19d6673 thank you !

PR #261, opened 2017-02-19
Map bulider fix with Incorrect order of arguments

catterpiler74:
PVS-Studio warning: V764 Possible incorrect order of arguments passed to 'loadVMap' function: 'tileY' and 'tileX'. MapBuilder.cpp 279
The analyzer detected suspicious passing of arguments to the function - the arguments tileX and tileY swapped places.

If we have a look at the prototype of the function loadVMap(), then we can clearly see that this is an error.
bool loadVMap(uint32 mapID, uint32 tileX, uint32 tileY, MeshData& meshData);


cyberium:
Thank you, but its not the solution unfortunately there is another inversion somewere else that we need to catch. Doing this will broke vmap by not loading correct data.

PR #263, opened 2017-02-19
variable is assigned values twice

catterpiler74:
PVS-Studio warning: V519 The 'stat' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 1776, 1781. DetourNavMeshQuery.cpp 1781
The analyzer detected a suspicious fragment, where the stat variable is assigned with different values twice. This code is definitely worth reviewing.


cala:
Tagging @cyberium here because he knows better that part of the code.

hlarsen:
looks like this is in recast (and is worked on/fixed in https://github.com/recastnavigation/recastnavigation/commit/2f382134d4de33d14ddab2eec2fe477b05305c05), shouldn't we just pull in a new version?

cyberium:
Not critical change it will be fixed in next recastnavigation update or with new mmap API.

PR #264, opened 2017-02-19
Redundant condition player.cpp

catterpiler74:
Redundant condition
PVS-Studio warning: V728 An excessive check can be simplified. The '||' operator is surrounded by opposite expressions '!realtimeonly' and 'realtimeonly'. Player.cpp 10536
The check (a && b) || !a can be simplified to !a || b, which can be seen in the truth table:
http://www.viva64.com/media/images/content/b/0476_CMaNGOS_ru/image2.png
Thus, the original expression can be simplified to this code:


cala:
Good catch! 👍 In https://github.com/cmangos/mangos-classic/commit/506f62267aada6a5b24a1e69497c87e60139a748

PR #265, opened 2017-02-19
Unconditional exit from the for loop Pet.cpp

catterpiler74:
PVS-Studio warning: V612 An unconditional 'break' within a loop. Pet.cpp 1956
V612 An unconditional 'break' within a loop. Pet.cpp 895
It was not clear what was meant here, but an unconditional break statement in the body of the for loop looks very suspicious.
 Even if there is no error here, it's better to refactor the code, and get rid of the unnecessary loop, because the iterator _spell_idx takes a single value.

### DIFF
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -892,7 +892,7 @@ int32 Pet::GetTPForSpell(uint32 spellid) const
             return 0;
 
         basetrainp = _spell_idx->second->reqtrainpoints;
-        break;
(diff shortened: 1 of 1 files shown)


cala:
Agreed: something is strange here and should be checked.

Patman64:
It looks like the loops only care about the first spell with training points, so the old code seems fine.

If it was wrong then **do not comment out the code**, just remove it.

PR #266, opened 2017-02-19
item_orb_of_draconic_energy fix

catterpiler74:
Integer division
PVS-Studio warning: V674 The '0.1f' literal of the 'float' type is compared to a value of the 'unsigned int' type. item_scripts.cpp 44
The method Unit::GetHealth() returns the value of the uint32_t type, and the method Unit::GetMaxHealth() also returns the value of the uint32_t type, so the result of the division is an integer, and it's pointless comparing it with 0.1f.

To correctly identify 10% of the health, this code can be rewritten like this.

### DIFF
--- a/src/scriptdev2/scripts/world/item_scripts.cpp
+++ b/src/scriptdev2/scripts/world/item_scripts.cpp
@@ -41,7 +41,9 @@ bool ItemUse_item_orb_of_draconic_energy(Player* pPlayer, Item* pItem, const Spe
 {
     Creature* pEmberstrife = GetClosestCreatureWithEntry(pPlayer, NPC_EMBERSTRIFE, 20.0f);
     // If Emberstrife is already mind controled or above 10% HP: force spell cast failure
-    if (pEmberstrife && pEmberstrife->HasAura(SPELL_DOMINION_SOUL) || pEmberstrife->GetHealth() / pEmberstrife->GetMaxHealth() > 0.1f)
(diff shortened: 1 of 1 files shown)


cala:
Again, another mistake of mine with implicit type cast habit from most script languages (Python, Lua).
This does no harm in the current C++ code, but you are right we should clean this with proper code.
If you don't mind, I would rather change the math logic here to remove the two cast and change a division to a multiplication:
```diff
-    if (pEmberstrife && pEmberstrife->HasAura(SPELL_DOMINION_SOUL) || pEmberstrife->GetHealth() / pEmberstrife->GetMaxHealth() > 0.1f)
+    if (pEmberstrife && pEmberstrife->HasAura(SPELL_DOMINION_SOUL) 
+        || 10 * pEmberstrife->GetHealth() > pEmberstrife->GetMaxHealth())
```
If you agree, I'll merge with this change.

catterpiler74:
Of course, do so. the multiplication is faster than division ащк CPU?

ulfgebhardt:
http://stackoverflow.com/questions/4125033/floating-point-division-vs-floating-point-multiplication

in general it should not be the programmers concern what is faster, but the task of the compiler to chose the best solution - programmers should focus on readability. < opinion

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/b5c625cfee5b406737ba3e280c7123e4a3332395

@ulfgebhardt I agree that code readability should be the first concern since most modern compilers optimise the code by themselves. In my opinion, the last change does not alter readability, especially as this part of the code is commented.

Thanks both of you.

PR #267, opened 2017-02-19
Copy-Paste fix BattleGround

catterpiler74:
PVS-Studio warning: V760 Two identical blocks of text were found. The second block begins from line 213. BattleGround.cpp 210
The same actions are performed twice in this fragment. This code is most likely as a result of using Copy-Paste.


cala:
Merged in https://github.com/cmangos/mangos-classic/commit/5e68e7001e0cc5439e1b3e4484f0b7eb56847d73 thank you !

PR #268, opened 2017-02-19
Bit wise AND with a null constant

catterpiler74:
Bit wise AND with a null constant
PVS-Studio warning: V616 The 'SPELL_DAMAGE_CLASS_NONE' named constant with the value of 0 is used in the bitwise operation. Spell.cpp 674
The constant SPELL_DAMAGE_CLASS_NONE has a null value, and the bitwise AND of any number and null is a null, therefore the condition will always be false, and the next block will never be executed.

A similar error:
V616 The 'SPELL_DAMAGE_CLASS_NONE' named constant with the value of 0 is used in the bitwise operation. Spell.cpp 692

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -671,7 +671,7 @@ void Spell::prepareDataForTriggerSystem()
         default:
             if (IsPositiveSpell(m_spellInfo->Id))           // Check for positive spell
             {
-                if (m_spellInfo->DmgClass & SPELL_DAMAGE_CLASS_NONE) // if dmg class none <- the next block will never be executed
+                if (m_spellInfo->DmgClass == SPELL_DAMAGE_CLASS_NONE) // if dmg class none <- the next block will never be executed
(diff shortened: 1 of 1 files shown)


Warlockbugs:
Tagging: @Cyberium @killerwife 

In our case, SPELL_DAMAGE_CLASS_NONE is a special, little explored kind of attack table. In some aspects, it follows the ruleset of spell casts (such as using spell crit chance), while ignoring others.
Since we have a special proc event attached to it, and damage class is not a bitmask, i suggest the correct fix for this warning to be:
`if (m_spellInfo->DmgClass == SPELL_DAMAGE_CLASS_NONE)`

killerwife:
I agree.

cyberium:
👍

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/1a296847f5eb077de79eddf3591e6c457176a637 thanks everyone.

PR #269, opened 2017-02-21
BELNISTRASZ agro fix in razorfen_downs script

catterpiler74:
The second and third operands of the ternary operator are identical; this is most likely an error. Judging by the code of the project, we can assume that one of the operands should have the value SAY_BELNISTRASZ_AGGRO_2.


cala:
Merged in https://github.com/cmangos/mangos-classic/commit/dc695441327da04083f6f07d3163a8e91d032c71 thank you ! 👍

PR #275, opened 2017-04-06
Removed unnecessary if-statement

henkethebenke:
Small shit from [https://www.viva64.com/en/b/0476/#ID0EZTBK](https://www.viva64.com/en/b/0476/#ID0EZTBK)


Fabi:
Vote for rejecting it . Reason: Same than your previous pull request...

Warlockbugs:
This is actually a dependency file. If we are aiming to move away from g3d in the future, it wouldn't matter in the long run.

PR #276, opened 2017-04-16
Full integration of SD2 to the core.

cyberium:
No more need to build another separate lib for it.

Well where iam not so happy to introduce script side by side with the code, i must admit it make things easier.
This not change how to make script by any way. So all existings script should work.

Folder structure is changed to put all AI in same folder.
ScriptMgr now manage only dbscripts.

@killerwife @xfurry -> http://i.imgur.com/YBnupnv.png


xfurry:
@cyberium from the structure and organization point of view it looks fine.
However I didn't understand why did you say:

> iam not so happy to introduce script side by side with the code

Also, how will it work from the developer point of view now? Will we drop the 2 VS project files that contain the core and SD2 separately?
If yes, the only downside I can see is the extra time consumed during the core build.

cyberium:
What i meant is script should not be part of the core in perfect world.
Core is core and script should be separate at least in a separate project.
With this patch script is now full part of the core and difference with "script" and "code" will be hard to see.

Yes, we'll drop support for non inculded sd2 script.
Build time is slower but faster than before when sd2 need a rebuild.

Its also to make all of this easier for upcoming unified cmake patch.

killerwife:
@xfurry if we had two projects, we would have something called "cyclic dependency". And thats very bad for design. The problem is with the monolithical structure of wow-emu cores. The issue is that SD2 scripts create functions/objects which in turn the core uses. If we wanted to have a "pure" structure we would need  to extract all non-framework code from game lib. This includes spell scripts, aura scripts, but as well as the update sequences of objects.

xfurry:
@killerwife I was just asking, so please don't kill me for this. 😄 

From my point of view it's fine anyway. I think that now we have good enough computers, so that the build performance doesn't matter that much.

cala:
My opinion was not requested on this but I give it anyway because I'm a punk:
- as @xfurry said compile time should not be an issue because we have powerful enough computers for this kind of projects. When I have to build several times in a row, this is because I'm working on a script and only a handful of files need to be rebuilt. This is not an issue.
- this is better from a dependency point of view, @killerwife is right on this
- I see this as a move towards a more unified script system

However : 
- this will make script/core issues harder to distinguish for newcommers on the project and we need to keep those issues separated because the people working on core are not the same than the ones working on SD2 (though there are cross-contributions)
- I think we should still keep server part separated from the content part, from a design point a view and from a legal point of view (I know this last argument looks outdated but the recent nolstalrius thing should remind us that we are still in gray zone and that this gray may be darker than expected)

So, yes from a developer/design point of view, but this will probably raise some "Public Relation" issues like the last major changes.

cyberium:
Thank @cala for your comment, you know that you are always welcome to do so ;)
I invite everyone to react.
I know its a big change but thats only the start ;)

For what i see all current reaction are positive wich is already a bit surprising on my side ;) I expected more against reaction.
So i let that PR still a couple of days to let more chance to react then after fixing linux build i'll push it.

Inifield:
Honestly I see no harm in it, great job

cala:
> For what i see all current reaction are positive wich is already a bit surprising on my side ;) I expected more against reaction.

For my part, I'm OK with this kind of changes in the structure, though I would prefer development to focus on the many parts of the core that still need updates and that are related to the playable content. But I think that everyone is free to work on what is interesting him/her as long as it does not break the work from others.

killerwife:
@cala Actually this solves me exporting GameObject AI to cmangos, because I had a nasty circular dependency (objects were only created in .so and linker was refusing to export symbols). I had to put GO AI where Cyber put whole SD2 for it to work nicely.

cala:
@killerwife I'm eager to see that GameObject AI. 👍

xfurry:
Gameobject AI? :question: What is that supposed to do? How can a static gameobject have an AI component?

killerwife:
Because GOs are not as static as they are in our core. They cast spells, and apparently, launch music during certain events, as already mentioned bells in major cities, darkmoon faire music, pirates day music, midsummer festival music, brewfest music or black temple illidan fight. While we still need to implement go spell casting support (for which I have sniffed proof), music is a nice and easy start. Though it brings me to a funny limitation of having to implement zone/area/distance based music sending, as well as emote distance based sending, for which we have sniffed/video proof as well. I will export it next month when I get around to merging this sd2 move to voa core.

PR #277, opened 2017-04-18
Molten Core - Garr

Tobschinski:
+ update existing timers
+ add Massive Eruption

requires https://github.com/cmangos/classic-db/pull/51

### DIFF
--- a/src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
@@ -131,12 +131,7 @@ struct boss_garrAI : public ScriptedAI
 
 struct mob_fireswornAI : public ScriptedAI
 {
-    mob_fireswornAI(Creature* pCreature) : ScriptedAI(pCreature)
-    {
-        m_pInstance = (ScriptedInstance*)pCreature->GetInstanceData();
-    }
-
-    ScriptedInstance* m_pInstance;
+    mob_fireswornAI(Creature* pCreature) : ScriptedAI(pCreature) {}
 
     void Reset() override {}
 
--- a/src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
+++ b/src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
@@ -16,8 +16,8 @@
 
 /* ScriptData
 SDName: Boss_Garr
-SD%Complete: 80
-SDComment: Firesworn erruption needs to be revisited
+SD%Complete: 95
+SDComment: 'Summon Player' missing
 SDCategory: Molten Core
 EndScriptData */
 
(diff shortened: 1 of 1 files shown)


cala:
As I understand the massive eruption from your code and the spells DBC:
- After 6 min Garr will cast a trigger spell and that will make its adds explode (a more powerful version) and he will then cast it again every 20 seconds. Am I right, m'kay?

src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
```diff
@@ -110,43 +134,31 @@ struct mob_fireswornAI : public ScriptedAI
     mob_fireswornAI(Creature* pCreature) : ScriptedAI(pCreature)
     {
         m_pInstance = (ScriptedInstance*)pCreature->GetInstanceData();
-        Reset();
```
cala: Why remove the `Reset()` call?

src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
```diff
@@ -16,8 +16,8 @@
 
 /* ScriptData
 SDName: Boss_Garr
-SD%Complete: 80
-SDComment: Firesworn erruption needs to be revisited
+SD%Complete: 95
+SDComment: 'Summon Player' missing
```
cala: Was there a summon player back in Classic for Garr? I know it was only introduced in early TBC for Emerald Dragons and Onyxia. (and in fact, I did not even know he had a summon ability)

Tobschinski: Most bosses in MC have this ability, but no clue when it got added. ID 20477 is certainly a vanilla spell.

Tobschinski:
Yes, that's how it should be done, afaik.

It will make the encounter significantly easier after 6 minutes, because a simple 9% enrage on Garr will be no match for a Firesworn's damage output. But who am I to question this.

src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
Tobschinski: I don't see any combat values that need to be initialized or reset here. The instance data is also no longer required and can be removed.

src/scriptdev2/scripts/eastern_kingdoms/molten_core/boss_garr.cpp
```diff
@@ -26,17 +26,20 @@ EndScriptData */
...
     SPELL_ANTIMAGICPULSE        = 19492,
     SPELL_MAGMASHACKLES         = 19496,
     SPELL_ENRAGE                = 19516,
+    SPELL_SUMMON_PLAYER         = 20477,
+    SPELL_ERUPTION_TRIGGER      = 20482,
     SPELL_SEPARATION_ANXIETY    = 23487,                    // Aura cast on himself by Garr, if adds move out of range, they will cast spell 23492 on themselves
 
-    // Add spells
-    SPELL_THRASH                = 8876,
-    SPELL_IMMOLATE              = 15733,
```
cala: wowhead and wowpedia say that Firesworn should use Immolate, though I don't remember them using that ability (and can't tell for sure).

cala: Oh. creature_template_addon. Right. You have my blessing to punch me in the face. With a trout.

Tobschinski: I wouldn't mind to assign these spells here, but with the need of all these re-casts and re-checks if it's really up it's kinda stupid (imho). Why not just use creature_addon_template instead.

cala: Agree, I just missed that you've moved these spells to creature_template_addon when writing my first comment.

Tobschinski: https://youtu.be/kxQbullCk5M?t=38

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/01f7b81a3de9ccc8cff071d15de4adb9edb1ab96
Thank you! 👍

PR #278, opened 2017-04-19
Knock Away - Script Effect

Tobschinski:
+ some versions of 'Knock Away' lower the target's threat by 50% (used by Molten Giants, Broodlord Razorgore and Vem)
+ assign correct spells to Broodlord Razorgore and Vem

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -3424,6 +3424,10 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
 
                     return;
                 }
+                case 18670:
+                case 18945:
+                    m_caster->getThreatManager().modifyThreatPercent(unitTarget, -50);
+                    return;
                 case 22539:                                 // Shadow Flame (All script effects, not just end ones to
                 case 22972:                                 // prevent player from dodging the last triggered spell)
                 case 22975:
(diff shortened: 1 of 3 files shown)


Tobschinski:
Actually ... I'm no longer convinced that the Script Effect does the threat reduce.

Warlockbugs:
It does threat reduction indeed, but exact values are unknown at this point. Empirical data points to 75% for some spells and 50% for others. This was done via serverside script effect in very old spells before new shiny dedicated threat effects were introduced into the game.

killerwife:
There are spells in TBC that do it using Script Effect as well, so I support what @Warlockbugs said.

dagochen:
Maybe take old KTM versions as a source? I think there were some values in it which worked very good for the players.

Tobschinski:
Well, I'm not so insecure about the threat reduction amount. It should be known since Wrath, (?) when the server started to give accurate threat updates.

I just have issues with the script effect. There's a bunch of spells added roughly at the same time as the spells in this commit, without a script effect, but with a threat reduction. Why are they different, just a matter of preference?

There's more Knock Away spells, all with a script effect, but not all of them with a threat reduction. I assume all of them are just copies of the spell added first, but who knows ... It is still possible that this effect does something completely different.

Warlockbugs:
I can take care of all knock away variants later if you are in doubt, i think i have enough data.

cala:
Merging it then and pocking @Warlockbugs for taking care of the variants later as he proposed.

PR #279, opened 2017-04-21
AQ20 - Ossirian the Unscarred

Tobschinski:
Update Ossirian's runspeed after 10 seconds into the fight, making the pull less challenging.

### DIFF
--- a/src/scriptdev2/scripts/kalimdor/ruins_of_ahnqiraj/boss_ossirian.cpp
+++ b/src/scriptdev2/scripts/kalimdor/ruins_of_ahnqiraj/boss_ossirian.cpp
@@ -82,9 +82,11 @@ struct boss_ossirianAI : public ScriptedAI
     uint32 m_uiCycloneTimer;
     uint32 m_uiStompTimer;
     uint32 m_uiSilenceTimer;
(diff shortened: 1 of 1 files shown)


src/scriptdev2/scripts/kalimdor/ruins_of_ahnqiraj/boss_ossirian.cpp
```diff
@@ -82,9 +82,11 @@ struct boss_ossirianAI : public ScriptedAI
     uint32 m_uiCycloneTimer;
     uint32 m_uiStompTimer;
     uint32 m_uiSilenceTimer;
+    uint32 m_uiSpeedUpTimer;
     uint8 m_uiCrystalPosition;
 
     bool m_bSaidIntro;
+    bool m_bSpeedUp;
```
cala: I think you can skip this boolean variable by setting `m_uiSpeedUpTimer` to 0 once the timer expired and checking if it is zero or not, instead.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/840d015f7132ef4ce906923e946e069178387417

PR #281, opened 2017-04-27
Hate to Zero

Tobschinski:
set hate towards target to zero (what a surprise...), Lava Annihilator will use this to swap targets after each successful melee hit


cyberium:
Please confirm if that need to be merged to other core too

Tobschinski:
Don't see a reason not to ...

Warlockbugs:
Already merged.

PR #282, opened 2017-04-28
Cross faction battleground

kuramajssken001:


cyberium:
Thank same answer than https://github.com/cmangos/mangos-classic/pull/251

PR #283, opened 2017-04-28
New folder structure

cyberium:
This PR introduce new folders structure for game project.
I think it will make easier to dev with. Some may think its not a critical change and its true. But we will always have more important things to do than that and if we dont do it now we loose oportunity to make dev work easier.

It also have a big advantage to make it easier to generate Visual Studio sln with CMake. With this change we are now able to remove sln file. We can generate our personal project file.

I added ability to generate each important projects separatly.
ex : you can now generate a project with only server game (mangosd) with or without scriptdev!
You can then avoid realmd build for nothing and generate a working core fastly without having to wait all scripts.
If you prefer you can generate a project file that contain all projects. So you'll have to build it one time and you'll get mangosd, realmd, and all extractor.
Ofc i tryed to make it working same way for linux so it should work as before plus bonus of the new way to select projects.

I changed some variable name so now you have to use to generate everything:
> cmake -DCMAKE_INSTALL_PREFIX=../_install -DBUILD_EXTRACTOR=ON -DBUILD_VMAP_EXTRACTOR=ON -DBUILD_MMAP_EXTRACTOR=ON ..

Feel free to comment, like, dislike or everything you would like to do. But do something! ;)


cyberium:
i think i will simplify extractors switch and make only available BUILD_EXTRACTORS. Its not hurting to generate 3 ones at same time.

PR #284, opened 2017-05-02
always display Obsidian creature's maximum power correctly

Tobschinski:
... such as Moam, Obsidian Eradicator, etc


cyberium:
Thank you its merged.

PR #290, opened 2017-05-05
Update to MySQL 5.7.18

some2:
This updates the MySQL client libs (libmysql) from version 5.1.49 to 5.7.18 (current)


cyberium:
Thank you very well for all those lib update i'll process them next week.

cyberium:
Thank you its merged.

PR #293, opened 2017-06-18
Update Creature.cpp

bdebaere:
Changed order of bytes1 being applied after auras to allow more consistent control through the database of the StandState of the creature

Setting the StandState flags last to allow a change in StandState in the database to be unaffected by the spell listed in auras
This fixes the StandState issue with spell 16093 for example (see Gordok Hyena)

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -1936,20 +1936,6 @@ bool Creature::LoadCreatureAddon(bool reload)
     if (cainfo->mount != 0)
         Mount(cainfo->mount);
 
-    if (cainfo->bytes1 != 0)
-    {
-        // 0 StandState
-        // 1 LoyaltyLevel  Pet only, so always 0 for default creature
-        // 2 ShapeshiftForm     Must be determined/set by shapeshift spell/aura
-        // 3 StandMiscFlags
-
-        SetByteValue(UNIT_FIELD_BYTES_1, 0, uint8(cainfo->bytes1 & 0xFF));
-        // SetByteValue(UNIT_FIELD_BYTES_1, 1, uint8((cainfo->bytes1 >> 8) & 0xFF));
-        // SetByteValue(UNIT_FIELD_BYTES_1, 1, 0);
-        // SetByteValue(UNIT_FIELD_BYTES_2, 2, 0);
-        SetByteValue(UNIT_FIELD_BYTES_1, 3, uint8((cainfo->bytes1 >> 24) & 0xFF));
-    }
-
     // UNIT_FIELD_BYTES_2
     // 0 SheathState
     // 1 Bytes2Flags, in 3.x used UnitPVPStateFlags, that have different meaning
@@ -1981,6 +1967,23 @@ bool Creature::LoadCreatureAddon(bool reload)
             CastSpell(this, *cAura, TRIGGERED_OLD_TRIGGERED);
         }
     }
+
+	// Setting the StandState flags last to allow a change in StandState in the database to be unaffected by the spell listed in auras
+	// This fixes the StandState issue with spell 16093 for example
+	if (cainfo->bytes1 != 0)
+	{
+		// 0 StandState
+		// 1 LoyaltyLevel  Pet only, so always 0 for default creature
+		// 2 ShapeshiftForm     Must be determined/set by shapeshift spell/aura
+		// 3 StandMiscFlags
+
+		SetByteValue(UNIT_FIELD_BYTES_1, 0, uint8(cainfo->bytes1 & 0xFF));
+		// SetByteValue(UNIT_FIELD_BYTES_1, 1, uint8((cainfo->bytes1 >> 8) & 0xFF));
+		// SetByteValue(UNIT_FIELD_BYTES_1, 1, 0);
+		// SetByteValue(UNIT_FIELD_BYTES_2, 2, 0);
+		SetByteValue(UNIT_FIELD_BYTES_1, 3, uint8((cainfo->bytes1 >> 24) & 0xFF));
+	}
+
     return true;
 }
 


killerwife:
How does it fix the issue? Afaik some "sleeping" states are not StandState but aura visual.

bdebaere:
The issue is that with for example 16093 the StandState is adjusted to SIT. While in the database it is DEAD. Gordok Hyena does not have an animation for SIT, which means it looks stoned with a Zzz's animation. This fix makes sure the DEAD StandState is applied which is what it should be.

killerwife:
Question is, why is it adjusted to SIT?

bdebaere:
Because of the spell being cast.
Creature::LoadCreatureAddon() looks up the spell to be cast in auras. This spell is cast on creature create.

When this spell has an AuraInterruptFlags containing AURA_INTERRUPT_FLAG_NOT_SEATED the target of the spell is forced to sit, which it should do for spells like drinking and eating.
See SpellAuras.cpp:
if (m_spellProto->AuraInterruptFlags & AURA_INTERRUPT_FLAG_NOT_SEATED && !m_target->IsSitState())
        m_target->SetStandState(UNIT_STAND_STATE_SIT);

killerwife:
It could very well be that that piece of code is completely wrong, and it should work for more stand states instead :)

bdebaere:
How do you see it could be wrong? It's logical that a spell which is interrupted upon not standing still (AURA_INTERRUPT_FLAG_NOT_SEATED) forces you to sit on cast.

It could be that it needs to work for other stand states I cannot comment on that. But that piece of code as it is now in SpellAuras.cpp is logical.

killerwife:
Because its a guessed Aura Interrupt Flag. Sitting down is just the most common one, hence the asumption was made.

bdebaere:
Alright, I get that. Even so, the code I changed makes the creature bytes1 flag in the database to be consistent with it's result in game. We could add a TODO to SpellAuras.cpp saying if it's reworked one should take a look at the LoadCreatureAddon function if that is something you would want.

killerwife:
The issue is more deep than you think. If a spell has that flag, and you change its StandState, the ZZZ aura should be effectively removed. We know this should not happen, hence we can make an asumption, that this stand state interrupt flag could be instead smth like "AURA_INTERRUPT_FLAG_STAND_STATE_CHANGED"
and that we need a handler for applying stand states during spells, not just sitting.

bdebaere:
Are you saying a creature with StandState DEAD with Zzz animation going to StandState SIT should not have it's Zzz removed?

killerwife:
AURA_INTERRUPT_FLAG_NOT_SEATED means that if an aura has it and you /stand, it removes flag. So if you /stand during the ZZZ aura, it will remove it. Thats okay, but /sleep will interrupt this as well, which is quite nonsensical. It is more tricky than it seems.

killerwife:
As of now, the spell works like "Must remain seated while ZZZ"

bdebaere:
"Thats okay, but /sleep will interrupt this as well, which is quite nonsensical."

It does not: https://ibb.co/bYsdF5

killerwife:
Because its incorrectly implemented. Is food interrupted by /sleep?

bdebaere:
No, it is not. Which is indeed odd looking.

Spell 430 (Refreshing Spring Water) has AURA_INTERRUPT_FLAG_NOT_ABOVEWATER and AURA_INTERRUPT_FLAG_NOT_SEATED.

Spell 16093 has AURA_INTERRUPT_FLAG_UNK0, AURA_INTERRUPT_FLAG_DAMAGE, AURA_INTERRUPT_FLAG_UNK2, AURA_INTERRUPT_FLAG_MOVE, AURA_INTERRUPT_FLAG_TURNING, AURA_INTERRUPT_FLAG_TALK, AURA_INTERRUPT_FLAG_USE, AURA_INTERRUPT_FLAG_MELEE_ATTACK, AURA_INTERRUPT_FLAG_UNK13, AURA_INTERRUPT_FLAG_NOT_SEATED.

The only one they share is AURA_INTERRUPT_FLAG_NOT_SEATED. In fact every drink/food in the database has that flag in common. I believe this guess is therefor quite educated.

If we would change this to AURA_INTERRUPT_FLAG_STAND_STATE_CHANGED, how would you know which StandStates are allowed for each spell?

I think instead /sleep does not trigger a StandState change to UNIT_STAND_STATE_SLEEP and this is what we have to add!

What you say is true concerning the spell 16093 in that THEN it would be cancelled by /sleep which would be illogical **for players**. But players will not have this visual, it is intended for creatures. Creatures however will use the DEAD animation for sleeping as they do not often have a SLEEP animation, it is also unnecessary to emote a creature /sleep when we have this DEAD animation combined with the Zzz spell animation.

So we can either chose to apply the bytes1 after the auras or we can look for a Sleep visual which does not have the flag AURA_INTERRUPT_FLAG_NOT_SEATED and hence does not force the target to use it's SIT animation. There exists such a spell: 25148. However, this does mean we have to cancel this aura on any movement, either in code or through creature_ai_scripts.

Your thoughts?

killerwife:
I tested on retail and food/drink is removed by /sleep. Which means this part is not working properly. In addition to that, only 4 copy pasted Sleep spells have the aura interrupt flag. The flag is removed on combat, so thats not a problem. Easiest course of action is to remove the AIF from these 4 identical spells and script the StandState in the Aura Dummy implementation.

Also why Stand State Dead for the hyena, and not Stand State Sleep?

bdebaere:
Because StandState SLEEP has no animation (for a lot of creatures, including Gordok Hyena):
https://ibb.co/jCDM2k
https://ibb.co/mywxa5
https://ibb.co/egCENk

DEAD is often the "SLEEP" or resting animation. You see it a lot, I haven't seen a SLEEP animation used yet in game.

killerwife:
Did you try Unit_stand_state_custom?

bdebaere:
What integer value is that?

killerwife:
9

bdebaere:
No, 9 is the same as standard animation: idle.
I've created a pull request which solves the drinking while sleeping problem.
https://github.com/cmangos/mangos-classic/pull/294

killerwife:
That one does look good. In regards to the rest of the problem, I will think about it.

cyberium:
Ok i think its not reasonable to change that order for a couple of spell problem.
Its a big risk for all others.
So it seem we are going to edit those spell to make them work as all others same kind of spells.

Thank you very much for your contribution

PR #297, opened 2017-06-22
[RFC] ZG Trash Cleanup

Phatcat:
Clean up the trash mobs in ZG, details at; https://github.com/cmangos/issues/issues/1406

Edit*: @cala @Grz3s @xfurry - It's ready for review. You are probably gonna have to rerun cmake. I have to going from and to this PR (think it's because of the new zg file)

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -584,6 +584,7 @@ UPDATE creature_template SET ScriptName='npc_gurubashi_bat_rider' WHERE entry=14
 UPDATE creature_template SET ScriptName='npc_hakkari_blood_priest' WHERE entry=11340;
 UPDATE creature_template SET ScriptName='npc_zulian_prowler' WHERE entry=15101;
 UPDATE creature_template SET ScriptName='npc_zulian_panther' WHERE entry=11365;
+UPDATE creature_template SET ScriptName='npc_soulflayer' WHERE entry=11359;
 
 --
 -- Below contains data for table `script_texts` mainly used in C++ parts.
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -581,6 +581,7 @@ UPDATE creature_template SET ScriptName='mob_zealot_lorkhan' WHERE entry=11347;
 UPDATE creature_template SET ScriptName='mob_zealot_zath' WHERE entry=11348;
 UPDATE creature_template SET ScriptName='mob_healing_ward' WHERE entry=14987;
 UPDATE creature_template SET ScriptName='npc_gurubashi_bat_rider' WHERE entry=14750;
+UPDATE creature_template SET ScriptName='npc_hakkari_blood_priest' WHERE entry=11340;
 
 --
 -- Below contains data for table `script_texts` mainly used in C++ parts.
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -585,6 +585,7 @@ UPDATE creature_template SET ScriptName='npc_hakkari_blood_priest' WHERE entry=1
 UPDATE creature_template SET ScriptName='npc_zulian_prowler' WHERE entry=15101;
 UPDATE creature_template SET ScriptName='npc_zulian_panther' WHERE entry=11365;
 UPDATE creature_template SET ScriptName='npc_soulflayer' WHERE entry=11359;
+UPDATE creature_template SET ScriptName='npc_razzashi_adder' WHERE entry=11372;
 
 --
 -- Below contains data for table `script_texts` mainly used in C++ parts.
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -1982,7 +1982,9 @@ INSERT INTO script_texts (entry,content_default,sound,type,language,emote,commen
 (-1309025,'The brood shall not fall!',0,1,0,0,'marli SAY_TRANSFORM_BACK'),
 
 (-1309026,'%s emits a deafening shriek!',0,2,0,0,'jeklik SAY_SHRIEK'),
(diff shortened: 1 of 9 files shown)


cala:
I assume this is also related to the PR in Classic-db. :)

cala:
Assigning myself.

oppahansi:
@cala Btw 

> Add Soulflayer AI - 87a7d12

Has been made by me, @Phatcat has found that mistake in the commit history, so just in case you have questions or need anything.

cala:
Back from holidays, I review this PR and this is a nice work with a lot of research (about old things, targeting specific details). :+1: 

I think all the commits can be merged as they are but three of them:
- The Hakkari Bloodpriest (see review comment)
- The Zulian Panther and Razzashi Adder commits. I think they belong to ACID.

For these last two, unless I missed something, they were moved to SD2 to ensure that Expose Weakness spells are only used when the NPC is behind the target (as stated in the spell tooltip). I saw in the spell DBC that they are missing the attributes to ensure this position (I think only player spells have those attributes). Basically, this is this issue: https://github.com/cmangos/issues/issues/239
So either: 
- we update the code to hardcode these spells (I strongly dislike this)
- we add the attributes in `spell_template` table (but if so many spells are missing them, I think this is not a Blizz mistake but intended)
- we add some EVENT_T_IN_BACK_TARGET to eventAI to handle this case (cleanest solution in my opinion). Poking @cyberium and @killerwife for their opinion on that

What do you think @Phatcat @oppahansi ?

oppahansi:
I prefer the clean solution adding an event to eventAI

Phatcat:
> The Hakkari Bloodpriest (see review comment)

I can't find any review nor any comment... What's the metter with it? If it's the dispel logic there's also one of the other commits that contain that same kind of logic... the... zealot one I think it is... Is it the targeting of furthest away?

Well if I could see the review comment I'd know, so I'm just gonna wait for that. Thank you for getting around to this =)

cala:
Merged in: https://github.com/cmangos/mangos-classic/commit/51282e2db96c8edb5a8c56fc45b105cfb047b95a, https://github.com/cmangos/mangos-classic/commit/2515b137d0709540f734b4733b331e96b042e162, https://github.com/cmangos/mangos-classic/commit/8c2c589a9c5abf8dc744321839029861c259e850, https://github.com/cmangos/mangos-classic/commit/1a2048f24ba13783229dd6919e751cbb6c7f589a, https://github.com/cmangos/mangos-classic/commit/631004102ba323328653588cdd356ea1b1297364, https://github.com/cmangos/mangos-classic/commit/e6beb492cbd3f5a069acaa4fbcbd83db1b5238cc and https://github.com/cmangos/mangos-classic/commit/1ce9c03ced61c07530855bf0104ee796030cfebc
Thanks to both of you. :+1: 


As discussed, the missing commits will be ported to ACID (tomorrow, I can). The Hakkari Bloodpriest will also be in ACID, using the recent EVENT_T_SELECT_ATTACKING_TARGET but it could be moved back to SD2 if that option is not satisfying enough.

Before up-porting this to TBC and WotLK, could you point out if some parts are not suitable to TBC/WotLK, please? Thanks again!

Phatcat:
> The Hakkari Bloodpriest will also be in ACID, using the recent EVENT_T_SELECT_ATTACKING_TARGET but it could be moved back to SD2 if that option is not satisfying enough.

How is it supposed to target the player farthest away through ACID? - That was the sole reason I moved it to sd2....

Phatcat:
To my knowledge nothing in these commits are specific to classic, it should all apply to the other cores as well.

cala:
> How is it supposed to target the player farthest away through ACID? - That was the sole reason I moved it to sd2....

I understand that you did not get the comment that led to that decision. Here it is:
```c++
 if (m_creature->GetHealthPercent() < 40.0f)
+                if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_FARTHEST_AWAY, 0, SPELL_DRAIN_LIFE, SELECT_FLAG_NOT_IN_MELEE_RANGE))
+                    if (DoCastSpellIfCan(pTarget, SPELL_DRAIN_LIFE) == CAST_OK)
```
> I'm afraid I have to disagree with you here: Drain Life (24618) has a 20 yards range (http://www.wowhead.com/spell=24618), meaning that you could nullify that ability by making everyone in range (but the tank) and clearly, all strats about this NPC say the opposite (http://www.wowhead.com/npc=11340/hakkari-blood-priest#comments:id=1294).
Based on this, I assume the life drain mentioned in the Dungeon Companion is in fact the Blood Funnel which has a 100 yards range (http://www.wowhead.com/spell=24617).

> In my humble opinion, when the authors of the Dungeon Companion wrote : 
>> Rather than having a close range health drain, the Bloodpriest drains life from those farthest from them. Make sure everyone is at close range. Stun lock at 40%.

> I don't think they meant that the NPC always pick specifically the target that was the farthest (as the SD2 script here does), but *one* (or some) of the target that is not in melee to interrupt the channel. Based on this assumption (that is open for discussion, of course), I think that this NPC is doable in ACID with EVENT_T_SELECT_ATTACKING_TARGET (see: https://github.com/cmangos/mangos-wotlk/blob/master/doc/EventAI.txt#L97)

As stated above, that is open for discussion. You can see the ACID implementation here: https://github.com/cmangos/classic-db/commit/ea727a398b927c7d222bddc7a8e3f0a0c227f19a

Phatcat:
I was sure I saw one of the spells having a 100 yard range, and surely as you say, that is however the funnel and not not the drain, but you're right, that description from DC could very well be anyone outside of melee range, which, yeah, I see with the EVENT_T_SELECT_ATTACKING_TARGET and then a set min range which is outside of melee range then yeah, that seems totally acceptable.

Thanks for clarifying, I am 100% down with your changes :+1:

cala:
@Phatcat and @oppahansi 
And here is the PR allowing the NPCs to only cast Exploit Weakness when they are behind the target:
https://github.com/cmangos/mangos-classic/pull/314

Once merge, I'll update the NPCs' scripts and we should be good. 😸

oppahansi:
GJ 👍

PR #299, opened 2017-06-24
update the bot AI

yushicsm:
the bots will heal,dispel and Resurrect the players who in the same group first

### DIFF
--- a/src/game/PlayerBot/Base/PlayerbotClassAI.cpp
+++ b/src/game/PlayerBot/Base/PlayerbotClassAI.cpp
@@ -222,12 +222,27 @@ Player* PlayerbotClassAI::GetHealTarget(JOB_TYPE type)
         for (Group::member_citerator itr = groupSlot.begin(); itr != groupSlot.end(); itr++)
         {
             Player *groupMember = sObjectMgr.GetPlayer(itr->guid);
-            if (!groupMember || !groupMember->isAlive() || groupMember->IsInDuel())
+            if (!groupMember || !groupMember->isAlive() || groupMember->IsInDuel() || !groupMember->IsInSameGroupWith(m_bot))
                 continue;
+			if (groupMember->IsInSameGroupWith(m_bot))
+			{
             JOB_TYPE job = GetTargetJob(groupMember);
             if (job & type)
                 targets.push_back( heal_priority(groupMember, (groupMember->GetHealth() * 100 / groupMember->GetMaxHealth()), job) );
+			}
         }
+		for (Group::member_citerator itr = groupSlot.begin(); itr != groupSlot.end(); itr++)
+        {
+            Player *groupMember = sObjectMgr.GetPlayer(itr->guid);
+            if (!groupMember || !groupMember->isAlive() || groupMember->IsInDuel())
+                continue;
+			if (!groupMember->IsInSameGroupWith(m_bot) && m_ai->GetManaPercent() >= 80)
+			{
+            JOB_TYPE job = GetTargetJob(groupMember);
+            if (job & type)
+                targets.push_back( heal_priority(groupMember, (groupMember->GetHealth() * 100 / groupMember->GetMaxHealth()), job) );
+			}
+		}
     }
     else
     {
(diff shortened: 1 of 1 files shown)


cala:
Hi,

Thank you for this pull request.

I'm currently very busy with the merging of TBC and WotLK version of Playerbot into their respective core. TBC branch requires especially a lot of work as it is 4 years behind Classic branch and I have hundred of commits to backport first.

Thus, I lack of time to review your PR in details. My first comment to make it more readable is to use our code style: indentation is four spaces, not tab, please.

I like your idea to introduce group priority in finding which target to heal. But in the current state of your PR, I'm afraid that main tanks in raid settings will have their survivability decreased. In Classic, most of the MT healers were in other groups due to the slots in MT group being occupied by other classes. With this change, the MT healers would priorize their barely injured group mates over the heavy hit MT.

My suggestion to work around that would be to add a secondary combat orders for healers to focus on the MT whatever their group is.

yushicsm:
The tima that I study C ++  is not very long, so it is not very standardized，sorry.
I have a server for 30 players who love to use the bots
Our real situation is that in a 40 raid which have 14-15 real players，
MT healers were all real players，about 2-3.
And before a battle of a boss，4-5 real players will switch their characters to heal
This healers is enough for MT

yushicsm:
Before I changed it，the heal bots will go out of mana quickly in a battle of BOSS
Because they always get the same target and made lost of  overheal
Now they work much better.

cala:
> Our real situation is that in a 40 raid which have 14-15 real players，
MT healers were all real players，about 2-3.
And before a battle of a boss，4-5 real players will switch their characters to heal
This healers is enough for MT

In the current state of playerbot, a full 40 man raid only needs 4 real players, a 20 man raid only 2 and a 10 man raid only one. This means that your code also needs to work for these dire cases.

Don't mind me: I like the idea of priorizing heal by group, it would definitely have the positive effects you described and is closer of what was done on retail. I'm just saying that your code is not robust enough and may only work for *your* raid set-up (ration players/bots).

I would suggest your code to be enhanced by adding a specific job like HEAL_MT and force healers with that job to priorize MT > their group members > other raid members.

cala:
Hi,

I have added new orders for bots so they can be main tank or main healer. Therefore, I am currently looking at your PR and I have a few remarks:
- basically, your changes to `GetResurrectionTarget()` and `GetDispelTarget()` seem to be ineffective. Though you are adding targets from the bot's group and from outside the group, one after the others, the targets are all sorted out later based on their individual job, resulting in the same behaviour than the initial function
- in `GetHealTarget()` you hardcode a mana threshold to exclude targets outside the bot's group. I think that such thresholds should be handled in each bot class AI (all classes don't have the same constraints). Additionally,  a bot with 79% mana will ignore a target from another group that is about to die, even if it has nobody requiring healing in its own group.

Hence, is my proposal:
```patch
- Player* PlayerbotClassAI::GetHealTarget(JOB_TYPE type)
+ Player* PlayerbotClassAI::GetHealTarget(JOB_TYPE type, bool onlyPickFromSameGroup)
 {
     if (!m_ai)  return nullptr;
     if (!m_bot) return nullptr;
@ -225,8 +225,12 @@ Player* PlayerbotClassAI::GetHealTarget(JOB_TYPE type)
         for (Group::member_citerator itr = groupSlot.begin(); itr != groupSlot.end(); itr++)
         {
             Player* groupMember = sObjectMgr.GetPlayer(itr->guid);
-            if (!groupMember || !groupMember->isAlive() || groupMember->IsInDuel())
+            if (!groupMember || !groupMember->isAlive() || groupMember->IsInDuel()
+                             || (!groupMember->IsInSameGroupWith(m_bot) && onlyPickFromSameGroup))
                 continue;

             JOB_TYPE job = GetTargetJob(groupMember);
             if (job & type)
             {
```
Thus, we keep a generic method that can include or exclude targets from outside the bot's group on demand. We move to each bot class AI the decision to exclude or not some targets and with which threshold based on bot orders.

cala:
The global idea of your change is merged, thank you. I'm currently updating individual classes AI to make use of it.

PR #302, opened 2017-07-01
Patch 3

bdebaere:
Clearcasting should not be consumed by Faerie Fire (Feral), Faerie Fire or Prowl.


killerwife:
Wrong? Faerie fire costs mana, and it is an offensive ability. If I am missing something, correct me please.

bdebaere:
You are correct, Faerie Fire costs mana indeed. It should be consumed. Then we have a problem because there is no way to differentiate between the two Faerie Fires based on their SpellFamilyFlags. Let me make some fixes.

killerwife:
The problem is that it should only be consumed by spells that actually used it up, which means messaging proc system in some way or another. I have a prototype for this for quite some time but its time consuming to copypaste it into all three cores. I will get around to it eventually.

bdebaere:
Prowl should definitely not be in the mask. It needs to be removed!

wowmui:
@bdebaere I'm curious how these values are calculated. Would you like to teach me?

bdebaere:
@wowmui These are hexadecimal masks. The spell_affect table decides which spells affect other spells by using masks. Since spell 16870 affects Prowl and it shouldn't, we need to subtract this value from the existing mask 0x001007F100E3FEFF. Prowl has a SpellFamilyFlags decimal value of 16384 so 0x001007F100E3FEFF to decimal minus 16384 = 4512331310808831. 4512331310808831 back to hex = 0x001007F100E3BEFF.

Of course, when doing this, you must be careful and check if any other spells have SpellFamilyFlags 16384 for the Druid. I checked and there are no other spells for Druid, only Prowl.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/c9491a097ce7a56199087e57030a679baeae3f0d thank you!

PR #304, opened 2017-07-11
[z2703] Remove duplicate info from creature_template

Phatcat:
creature_template contains a lot of duplicate info.

This removes Most things to do with ranged and streamlines faction for both ally and horde.

@Cyberium @cala @Tobschinski @killerwife @Grz3s

### DIFF
--- a/sql/base/mangos.sql
+++ b/sql/base/mangos.sql
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `db_version`;
 CREATE TABLE `db_version` (
   `version` varchar(120) DEFAULT NULL,
   `creature_ai_version` varchar(120) DEFAULT NULL,
-  `required_z2702_01_mangos_spell_chain_totems_typos` bit(1) DEFAULT NULL
+  `required_z2703_01_mangos_creature_template_fixup` bit(1) DEFAULT NULL
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC COMMENT='Used DB version notes';
 
 --
@@ -1162,8 +1162,7 @@ CREATE TABLE `creature_template` (
   `ModelId2` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ModelId3` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ModelId4` mediumint(8) unsigned NOT NULL DEFAULT '0',
-  `FactionAlliance` smallint(5) unsigned NOT NULL DEFAULT '0',
-  `FactionHorde` smallint(5) unsigned NOT NULL DEFAULT '0',
+  `Faction` smallint(5) unsigned NOT NULL DEFAULT '0',
   `Scale` float NOT NULL DEFAULT '1',
   `Family` tinyint(4) NOT NULL DEFAULT '0',
   `CreatureType` tinyint(3) unsigned NOT NULL DEFAULT '0',
@@ -1189,15 +1188,11 @@ CREATE TABLE `creature_template` (
   `MaxLevelHealth` int(10) unsigned NOT NULL DEFAULT '0',
   `MinLevelMana` int(10) unsigned NOT NULL DEFAULT '0',
   `MaxLevelMana` int(10) unsigned NOT NULL DEFAULT '0',
-  `MinMeleeDmg` float NOT NULL DEFAULT '0',
-  `MaxMeleeDmg` float NOT NULL DEFAULT '0',
-  `MinRangedDmg` float NOT NULL DEFAULT '0',
-  `MaxRangedDmg` float NOT NULL DEFAULT '0',
+  `MinDmg` float NOT NULL DEFAULT '0',
+  `MaxDmg` float NOT NULL DEFAULT '0',
   `Armor` mediumint(8) unsigned NOT NULL DEFAULT '0',
-  `MeleeAttackPower` int(10) unsigned NOT NULL DEFAULT '0',
-  `RangedAttackPower` smallint(5) unsigned NOT NULL DEFAULT '0',
-  `MeleeBaseAttackTime` int(10) unsigned NOT NULL DEFAULT '2000',
-  `RangedBaseAttackTime` int(10) unsigned NOT NULL DEFAULT '2000',
+  `AttackPower` int(10) unsigned NOT NULL DEFAULT '0',
+  `BaseAttackTime` int(10) unsigned NOT NULL DEFAULT '2000',
   `DamageSchool` tinyint(4) NOT NULL DEFAULT '0',
   `MinLootGold` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `MaxLootGold` mediumint(8) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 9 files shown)


killerwife:
The data is not duplicate.
    UNIT_FIELD_BASEATTACKTIME                 = OBJECT_END + 0x008D, // Size: 2, Type: INT, Flags: PUBLIC
    UNIT_FIELD_RANGEDATTACKTIME               = OBJECT_END + 0x008F, // Size: 1, Type: INT, Flags: PRIVATE

    UNIT_FIELD_ATTACK_POWER                   = OBJECT_END + 0x00CC, // Size: 1, Type: INT, Flags: PRIVATE, OWNER_ONLY
    UNIT_FIELD_RANGED_ATTACK_POWER            = OBJECT_END + 0x00CF, // Size: 1, Type: INT, Flags: PRIVATE, OWNER_ONLY
    UNIT_FIELD_MINRANGEDDAMAGE                = OBJECT_END + 0x00D2, // Size: 1, Type: FLOAT, Flags: PRIVATE, OWNER_ONLY
    UNIT_FIELD_MAXRANGEDDAMAGE                = OBJECT_END + 0x00D3, // Size: 1, Type: FLOAT, Flags: PRIVATE, OWNER_ONLY

The only duplicate thing is:


-  `FactionAlliance` smallint(5) unsigned NOT NULL DEFAULT '0',
--
  |   | -  `FactionHorde` smallint(5) unsigned NOT NULL DEFAULT '0',
  |   | +  `Faction` smallint(5) unsigned NOT NULL DEFAULT '0',

But gotta first figure out why it was designed that way in the first place.

Phatcat:
https://1drv.ms/i/s!Agpa05rA5GSClWUWxZpwY3kaNbI2

nothing with rangedattackspeed
nothing with rangedattackpower
nothing with rangeddamage

Phatcat:
is the value kept in rangedmin and rangedmax ever going to be different than what is kept in meleemin and meleemax? from a template perspective? When we look at a ranged mob like the gurubashi axe thrower or the hakkari shadow hunter, both of them have 0 in min and max ranged damage (https://github.com/cmangos/classic-db/commit/971f0aa5eeb107908de7f2f30ed76aca96a4af4c). This is because their dedicated spells takes care of their ranged damage. Is there any mob that does not handle it's ranged damage through a dedicated spell?

is there any particular reason we are storing attack power in both creature_template and creature_template_classlevelstats? is rangedattackpower ever going to matter if min and max fields will always be 0 for mobs?

killerwife:
The answer is very bad and very uninformative, but essentially these values come from sniffs and sniffs only. CLS is incomplete and needs to be extended to damage fully as well, but that is an unfinished concept. The person to ask about that is @cyberium . The removal of those fields cant be justified until all of that research and development is finished.

That being said, ask around for FactionAlliance and FactionHorde, ever since I came aboard I havent seen it being utilized.

Phatcat:
Is there any creature which does NOT have 0 in UNIT_FIELD_MINRANGEDDAMAGE and UNIT_FIELD_MAXRANGEDDAMAGE (which is sniffed as you say) ?

> CLS is incomplete and needs to be extended to damage fully as well, but that is an unfinished concept.

Once again, that is outside the scope of what I'm proposing here. I am not proposing to remove min and max damage altogether. This change should not break or change a single mob functionality wise.

>  The person to ask about that is @cyberium

And I did tag him, didn't I?

> The removal of those fields cant be justified until all of that research and development is finished.

Is there anything I can work on at all or is everything in one big state of 'in development' blocking out any non core developer?

killerwife:
I actually have a really good suggestion. SummonCreature function enables us to create temporary summons in C++ or SD2 alike. Recent evidence has uncovered that a tempsummon can have custom UnitFlags, NPCFlags or Faction. We need to add a table lets say called dbscript_tempsummon_settings which will enable us to make "templates" for tempsummons and set any parameter. It needs to be loaded in C++, and used in EAI and DBScripts and the UnitFlags, NPCFlags or Faction need to be added to SummonCreature function. Most likely need to create a C++ struct for it, now that the parameter count is getting insanely high, and also a wrapper for it. So far I know 6 scripts where this is needed, and I am sure WOTLK will need it even more.

FactionHorde, FactionAlliance seems to be used in Western Plaguelands for a spectral gryphon master who changes factions based on who is controlling the outpost (outdoor pvp). So it seems its used after all. - worth researching further

For the other unit fields, I tried asking around and we havent found enough legacy information to know why they were implemented, and removing something we dont know why it was added in the first place, is a bad idea.

cyberium:
Thank you for this. However you are wrong. Creature can have both melee and range data with differents values. So wee need to have both set somewhere. Actualy @killerwife said what you need to know. We keep all sniffed data for now, they are precious.
For CLS even when it will be fully implemented iam not sure if it will be wise to remove not more needed data.

Phatcat:
> Creature can have both melee and range data with differents values.

[15:42] <Tobschinski> andi havent seen yet that meleeattackspeed is different from rangedattackspeed
[15:42] <Tobschinski> in sniffs
[15:42] <Tobschinski> unless there's auras
[15:42] <Tobschinski> so it makes sense to only store one of them in c_template

> https://1drv.ms/i/s!Agpa05rA5GSClWUWxZpwY3kaNbI2
nothing with rangedattackspeed
nothing with rangedattackpower
nothing with rangeddamage

and

> [23:39] <MantisLord> actually wait a minute this might be SMSG_UPDATE_OBJECT for Player not creature :D
[23:40] <MantisLord> okay nvm actually. The only occurrences I see of UNIT_FIELD_MINRANGEDDAMAGE in my Full SSC ssniff
[23:40] <MantisLord> are updates to players
[23:40] <MantisLord> not creatures
[23:43] <Phatcat> I think you should make a comment about that on the PR so killerwife can see that.
[23:45] <MantisLord> ok 
[23:51] <MantisLord> talking to him now about it. I actually am looking in some other sniffs from TBC
[23:51] <MantisLord> and it seems like same thing
[23:51] <MantisLord> only occurs for player update

Tobschinski:
Faction, I do not see a reason to keep both fields, faction_templates will always affect horde and alliance, friendly and hostile units. There is currently just one creature in classic_db with different values, some ghost in the Eastern Plaguelands that is summoned if you control the towers. That faction can be changed with scripts, and probably should be.

About the other stuff, I have never seen it to exist for creatures (except rangedbaseattacktime, which is always the same as meleebaseattacktime, I think). Just because it's part of SMSG_UPDATE_OBJECT for some units doesn't necessarily have to imply that it should be stored in creature_template. Especially if you can get it from somewhere else (loading baseattacktime twice, or set something as default value.

However, not partially removing these fields and wait for @cala to finally get all multipliers right (hurry the f*** up) to finish the classlevelstats_implementation makes sense. So maybe just focus on faction then ...

cyberium:
For Faction i remember asking @Schmoozerd about it, if my memory is right i was stating something like.
Its for creature where faction depend on player faction at spawn. You have to read "FactionAlliance" as "Faction for Alliance player". Lot of change have been done in the core so iam not sure if its still right. And yes it can be done by script. So for that @Grz3s @killerwife @cala have to be ok.

killerwife:
From my end if anyone wants to review all instances of using different ones in FA and FH columns and make them work by changing it in the script, be my guest.

PR #305, opened 2017-07-11
Update what is considered CC by EventAI

Phatcat:
Not enough things were considered CCs, and considering isCharmed a CC is
useless since we check if the mob is hostile to us first, which it will be if it's charmed, so we would never even get to this check.


cala:
👍

cala:
Unless someone has arguments against this expansion of what is considered as "crowd crontrol" for NPCs, I will merge this next week.

cyberium:
Poke @Warlockbugs . 😈

Tobschinski:
Did someone say @Warlockbugs ?

Warlockbugs:
Depends on what this is for.
What should temporarily pause AI: Stun, Fear, Confusion 
What should disable/replace AI: Charm, Posses

killerwife:
Just merge it. It will not harm anyone.

PR #306, opened 2017-07-20
Update Player.cpp

Berneth:
To fix the race weapon specialization not-leveling up bug.
Match the change on Player.h.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -5110,7 +5110,7 @@ void Player::UpdateCombatSkills(Unit* pVictim, WeaponAttackType attType, bool de
     if (lvldif < 3)
         lvldif = 3;
 
-    int32 skilldif = 5 * plevel - (defence ? GetBaseDefenseSkillValue() : GetBaseWeaponSkillValue(attType));
+    int32 skilldif = 5 * plevel - (defence ? GetPureDefenseSkillValue() : GetPureWeaponSkillValue(attType));
 
     // Max skill reached for level.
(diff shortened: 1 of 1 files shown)


cala:
Hi. Sorry for the delay in handling your PR (summer time). If I understand correctly what this fixes, it should allow, let's say orcs with +5 bonus skill to axes to level up until 305 skill while they are currently stopping at 300 or 301 skill. Am I right?

killerwife:
Isn't this supposed to work like adding 5 skill instead of leveling it up 5 points further? Just asking.

Berneth:
The racial skill modifiers add 5 to the current and the maximum for a skill.
In the mangos zero (I don't know the rest because I'm just using zero) it doesn't work like that, it just adds 5 to the current and it seems to add 5 to the maximum but you can't level it the last 5 points because of this bug.
IE: if you are a human warrior, you have level 1/5 skill of swords which in reality is 6/10. As 6 is higher than 5 you can level the skill up to 10. With this and the matching change on Player.h (https://github.com/cmangos/mangos-classic/pull/307) this bug is fixed.

I'm sorry I don't know how to submit both changes at the same time.

cala:
> I'm sorry I don't know how to submit both changes at the same time.

You can do that by modifying the Player.h file on the current branch, committing the results and pushing it. It will add to the current Pull Request.

cala:
Unless change request is made, I will merge this next week.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/0ac62f15a654f68b5cb61e24eb145f0769fabc8c
Thank you!

PR #307, opened 2017-07-20
Update Player.h

Berneth:
To fix the race weapon specialization not-leveling up bug.
Match the change on Player.cpp.


cala:
Merged in https://github.com/cmangos/mangos-classic/commit/0ac62f15a654f68b5cb61e24eb145f0769fabc8c
Thank you!

PR #308, opened 2017-07-23
Fix random spawn times.

Zooke:
Don't be too critical, I am no programmer.
But from a bit of code comparison I am guessing that this was how it was supposed to be.
Current behaviour is not random at all, this change seems to be random from a few tests.


killerwife:
Ugh @Warlockbugs ? For me both were random. What OS are u using?

Zooke:
Windows 7.
I set the spawntime of Mother Fang (471) to 30secs min, 90secs max.
Then I timed the respawn, over 10 kills it became obvious that the spawn was far from random.

PR #309, opened 2017-07-25
BWL - Elemental Shield / Broodpower

Tobschinski:
allow more creatures than just Chromaggus to use Elemental Shields & Broodpowers
(e.g. Death Talon Wyrmguard, Death Talon Overseer)


cala:
👍

PR #310, opened 2017-07-26
Script of the Darkmoon Faire Cannon

Br2m:
That's what come with https://github.com/cmangos/classic-db/pull/110
Not really sure about how I handled orientation but it works well that way.
I Took this as a source : https://www.youtube.com/watch?v=ywEpt0u0Qa8
Without that orientation change, the player has to look in the wrong direction to be knocked back to the target, so we just change his direction when the knock back is done and set it back right after...

( Sorry I'm a bit new, don't know how everything have to be sent etc.... So If I do a mistake please tell me! )
Not sure if I should send the DB part too here ?


cala:
Hi,

Thanks for your PR, we will look at it. The DB part is fine where it is now. 👍

cala:
Closed by https://github.com/cmangos/mangos-classic/commit/ee327422ee5aa234b020b5e405322e8211af87f9
Thanks everyone.

PR #311, opened 2017-08-10
Make WorldSocket::SendPacket thread-safe for multithreading of sendin…

ryancheung:
…g packets

Currently, WorldSocket::SendPacket is not thread safe, cause there're two `Socket::Write`, each would acquire a lock of `Socket::m_mutex`. These two `Socket::Write` may cause `m_outBuffer` has unordered packet head and body.


namreeb:
Good find.  I wonder if a simpler solution though would be to alter `WorldSocket::SendPacket` to call `Socket::Write` only once?

ryancheung:
@namreeb Yes it should be simpler. But `m_crypt.EncryptSend` is not thread safe too, I think.

PR #312, opened 2017-08-20
New cooldown system

cyberium:
The purpose of this is to unify all entities cooldown. With this change cd's are handled almost same way for all of them.
I also took that opportunity to fix some issues and add some features.

    now items will overwrite category spell at adding and at checking time.
    category cd can live without spell cd and vice versa.
    saving category and category time to keep correct track of the cd.
    removed the workaround to identify permanent cd.
    lockout can now be applied to creatures.
    added a command to show the list of active cd
    added a command to clear active cd only client side (can help for debug)

cd and gcd are now applied to creatures too. However it seem that they rarely use spell that have those cd. This have to be confirmed.

As always we'll all be happy of any comment about this.

### DIFF
--- a/sql/base/characters.sql
+++ b/sql/base/characters.sql
@@ -455,12 +455,14 @@ UNLOCK TABLES;
 
 DROP TABLE IF EXISTS `character_spell_cooldown`;
 CREATE TABLE `character_spell_cooldown` (
-  `guid` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Global Unique Identifier, Low part',
-  `spell` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Spell Identifier',
-  `item` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Item Identifier',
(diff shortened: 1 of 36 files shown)


cala:
Very nice ! I agree that the cooldown system needed a few polish (but was still fine). I won't be able to make lots of comments about the code, as you might have guess but I can perform some tests if you have some things that you would like to see tested or other suggestions.

If I understand what it is, the lockout on creatures is much needed as all spells that should silence creatures by applying cooldown on a specific spell school are currently not working. Does that means that this should now be fixed or at least rather easily implementable, now ?

cyberium:
> Does that means that this should now be fixed or at least rather easily implementable, now ?

Its fixed, a creature is not able to cast any spell under lockout timer. However, AI should be improved, as you know, interrupt a spell will block any same spell school to be able to be cast. But AI should handle that by, for ex move the creature to the enemy or try to use another spell school. Wich is not so well handled currently.

I'll merge anyway this next week because its already a good improvement.

PR #313, opened 2017-08-31
Issue 1418 - One thread spins at 100% CPU

sterlingalexander:
With the addition of commit 1ccdd8e2, the console was by default enabled.  On Linux this caused one mangosd thread to spin at 100% CPU, polling the console constantly for input.  This commit resolved the issue by cleaning up the pre-processor defines and switching from usleep to nanosleep.  All of this is done using linux-specific code wrapped in pre-processor directives.

### DIFF
--- a/src/mangosd/CliRunnable.cpp
+++ b/src/mangosd/CliRunnable.cpp
@@ -569,7 +569,7 @@ bool ChatHandler::HandleServerLogLevelCommand(char* args)
 
 /// @}
 
-#ifdef linux
+#ifdef __linux__
 // Non-blocking keypress detector, when return pressed, return 1, else always return 0
 int kb_hit_return()
 {
@@ -589,6 +589,13 @@ void CliRunnable::run()
 {
     ///- Init new SQL thread for the world database (one connection call enough)
     WorldDatabase.ThreadStart();                            // let thread do safe mySQL requests
+
+    // Get a timespec for the nanosleep
+#ifdef __linux__
+    struct timespec ts;
+    ts.tv_sec = 0;
+    ts.tv_nsec = 100000;
+#endif
 
     char commandbuf[256];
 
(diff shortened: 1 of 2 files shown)


cyberium:
First of all thank you very much for your contribution. It realy help to get real linux user feedback.

Could you please try to only replace ```linux``` by ```__linux__```.
Also after some though about it i think that code should be wrapped by ```__unix__``` instead because its supposed to work with ```__APPLE__``` too.
And why do you think replacing usleep by nanosleep help?

sterlingalexander:
Ok, I can rework these to use __unix__ and see what happens.  

The rational for replacing usleep by nanosleep is two-fold:  usleep is considered 'obsolete' because it has inconsistent behavior when it is interrupted by a signal.  nanosleep can store and 'restart' the sleep, though I didn't take that approach here as it didn't seem warranted.  Also, nanosleep is an actual system call where usleep is a C(++) standard library construct that uses timers.  

In looking at the output, I have no real idea why the OSX builds failed, and the logs seem to point to files I didn't modify.  I don't have OSX hardware to test on unfortunately.  I do know of one openssl-headers issue, could this be the problem with OSX builds?

  https://spiegelmock.com/2016/07/05/mac-os-x-el-capitan-and-openssl-headers/

cyberium:
The osx fail is not related to this don't worry at all about that.
My remark was more general.

Ok about nanosleep. Iam not so convainced about its use benefit though.
As an another alternative could you please try ```std::this_thread::sleep_for(std::chrono::microseconds(usec));```
It should be a portable way to do this. (chrono will be more heavly used in near future in cmangos)

sterlingalexander:
Pushed some changes based on your feedback, let me know what you think.

cyberium:
Please confirm me it work for you then its ready to merge :)

sterlingalexander:
Tested on linux and it works.  No more 100% CPU and the console runs
commands as expected.

cyberium:
Merged thanks

PR #314, opened 2017-09-06
Implement EVENT_T_FACING_TARGET

cala:
Allows NPCs to perform action only when facing the back or the front of their target in EventAI. This is typically useful for rogue-like NPCs.


cala:
This aims to fix issue https://github.com/cmangos/issues/issues/239.

Currently NPCs using rogue-like spells like Backstab can use it regardless of their position (front or back of their target). Position is controlled for players by the combined use of spell attributes SPELL_ATTR_EX_FACING_TARGET and SPELL_ATTR_EX2_FACING_TARGETS_BACK. Spells used by NPCs do not have these attributes though the tooltip still say that they must be in the back/front of their target.
This PR allows to control NPCs position in their respective AI script and thus prevent them to use abilities when they should not.
ACID scripts of all concerned NPCs will need to be updated to use this new event instead of the actual generic timer usually used.


This can also be used to fix the two NPCs mentionned in https://github.com/cmangos/mangos-classic/pull/297

killerwife:
I dont like this pr because it reuses the casting logic that is faulty at the base level but thats a global problem, not EAI problem. This means that this is as good a solution as any until I or someone else deals with the problem for good.There isnt much chance that this better system will come immediately, since its a lot more complex. As a result u have a green from me too if that matters.

cala:
> I dont like this pr because it reuses the casting logic that is faulty at the base level but thats a global problem, not EAI problem

I agree, but as you said, that is a larger problem to handle.

Once Travis gives its green light, I'll merge this PR.

PR #315, opened 2017-09-21
Don't set attack timer on both weapons

Phatcat:
The current situation of setting both weapons to 100 ms when out of
range / don't have target in front can be abused (every time you get a hit in
with your slow main hand, run out of range or turn around and wait for
the error to display, triggered by your fast offhander; now both your
weapons are ready again since attack timers are set to 100ms on both...

Ref; https://github.com/elysium-project/server/issues/2655


cala:
Sounds OK to me, but I'm not a master of the combat system, especially melee combat.

PR #316, opened 2017-09-23
Remove reference to prospecting_loot_template

Patman64:
Table doesn't exist in mangos-classic DB and causes errors when `z2709_01_mangos_comments.sql` is executed.


killerwife:
That's my fault. Thanks for fixing it.

PR #317, opened 2017-10-02
Fix NPC text localization

mityada:
NPC text locale strings were being erased in ObjectMgr::LoadQuestgiverGreeting due to a typo introduced in 8c09757.
This bug was spotted by @M2us.


killerwife:
@VladimirMangos There is a button in the dropdown menu called "rebase and merge", next time use that please.

VladimirMangos:
ok, but posible next time will not be? or will be some years later)

killerwife:
Oh I wasn't aware that the world would stop using GitHub ^^

PR #318, opened 2017-10-05
Fix questgiver greeting locales loading

mityada:
Use GetOrNewIndexForLocale to get correct locale index


killerwife:
@mityada You synchronized locale loading for locales_questgiver_greeting with the other loading procedures?

mityada:
@killerwife, yes. But synchronization was not the main reason. Your code assumed that all locales are always present and are in order. That's not true, so questgiver greeting localization was not working properly.

killerwife:
Did you test this? I do not have a different locale available so I cannot verify the PR.

mityada:
I did. I'm helping @M2us with Russian locale, he was the first to notice this bug.

killerwife:
I will push this to other branches today.

PR #319, opened 2017-10-16
Improve handle of SPELL_EFFECT_OPEN_LOCK for GAMEOBJECT_TYPE_TRAP

cala:
Improve the way traps are handled when targetted by SPELL_EFFECT_OPEN_LOCK (like disarm trap spell). Instead of being desactivated and despawning, they change state and use an alternate visual for the duration of the open lock (which is equal to the GO respawn time). They are then rearmed (back to GO_STATE_READY)

This should fix:
- the disarming of Suppression Device traps in BWL that were despawning instead of being disarmed
- quests 4734 and 4735 in UBRS where egg traps were still triggering instead of temporary changing state as expected

### DIFF
--- a/src/game/Entities/GameObject.cpp
+++ b/src/game/Entities/GameObject.cpp
@@ -66,6 +66,7 @@ GameObject::GameObject() : WorldObject(),
 
     m_isInUse = false;
     m_reStockTimer = 0;
+    m_rearmTimer = 0;
     m_despawnTimer = 0;
 }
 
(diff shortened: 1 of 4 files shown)


cala:
One minor issue: `SetGoState(GO_STATE_ACTIVE_ALTERNATIVE);` instantly changes the visual aspect of the GO while there should be a transition visual between the two states. Any idea on how playing that visual animation would improve this PR.

killerwife:
Animation state is a value from 0 to 255 i think. Maybe its bogus.

cala:
I'll perform some tests to see what I can get from the GO animation state when I can spare the time. As I said: it's a minor issue.

cala:
Unless new feedback or suggestion is made, I will merge this next week.

I'll take care of the animation part separately as it is a more generic issue that does not only concern this specific case.

PR #320, opened 2017-10-19
issues/863 - fix keyring slots

RichardGe:
fix for   https://github.com/cmangos/issues/issues/863

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -10514,6 +10514,24 @@ void Player::RemoveItemFromBuyBackSlot(uint32 slot, bool del)
     }
 }
 
+uint32 Player::GetMaxKeyringSize() const 
+{ 
+	int playerLevel = getLevel();
+	if ( playerLevel <= PLAYER_KEYRING_SIZE_STEP1_LVLMAX )
+	{
+		return PLAYER_KEYRING_SIZE_STEP1_SIZE;
+	}
+	else if ( playerLevel <= PLAYER_KEYRING_SIZE_STEP2_LVLMAX )
+	{
+		return PLAYER_KEYRING_SIZE_STEP2_SIZE;
+	}
+	else if ( playerLevel <= PLAYER_KEYRING_SIZE_STEP3_LVLMAX )
+	{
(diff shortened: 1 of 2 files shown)


src/game/Entities/Player.cpp
```diff
@@ -10514,6 +10514,24 @@ void Player::RemoveItemFromBuyBackSlot(uint32 slot, bool del)
     }
 }
 
+uint32 Player::GetMaxKeyringSize() const 
+{ 
+	int playerLevel = getLevel();
+	if ( playerLevel <= PLAYER_KEYRING_SIZE_STEP1_LVLMAX )
```
killerwife: Why not static array? divide by 10, 6 numbers, single access @cyberium what do you think

RichardGe: why not: something like :  
[0] = 4   // level : 0 -> 9
[1] = 4   // 10 -> 19
[2] = 4   // 20 -> 29
[3] = 4   // 30 -> 39
[4] = 8   // 40 -> 49
[5] = 12  // 50 -> 59
[6] = 12  //  level >= 60

btw, even if this code is only for Classic, I don't know how the slots evolve on TBC, ...etc...

killerwife: Afaik this isnt sent to client, so you can check tbc or wotlk by leveling up char and watching client UI.

RichardGe: I only have Classic on my side, I'll improve this commit with static array as I have described. later this week.

RichardGe:
modified as @killerwife suggested, with a static array.
I have also renamed GetMaxKeyringSize -->  GetMaxKeyringClientSize, in order to avoid confusion between the size seen by server (always 16 slots), and the size seen by Client (4, 8 or 12 slots)

src/game/Entities/Player.cpp
cala: For TBC and Wotlk,, it was changed in patch 2.0.1 to become a dynamic size keyring (so size no longer based on level). At time of patches 2.4.3 and 3.3.5, the max size of the keyring was 32 slots.

cala:
Merged, thank you very much!

PR #321, opened 2017-10-26
Fix civilian checks in Classic

cala:
- Make use of CREATURE_EXTRA_FLAG_NO_AGGRO to match TBC/WotLK cores behaviour with similar flag CREATURE_EXTRA_FLAG_CIVILIAN
- Add separate function for PvP civilian check that is a Classic specific behaviour related to Honor System at the time

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -1847,7 +1847,7 @@ bool Creature::CanAssistTo(const Unit* u, const Unit* enemy, bool checkfaction /
         return false;
 
     // we don't need help from non-combatant ;)
-    if (GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_AGGRO)
+    if (IsCivilian())
         return false;
 
     if (HasFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_NON_ATTACKABLE | UNIT_FLAG_NOT_SELECTABLE | UNIT_FLAG_IMMUNE_TO_NPC))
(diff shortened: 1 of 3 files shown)


cala:
More infos: https://github.com/cmangos/classic-db/pull/73#issuecomment-338945718

Warlockbugs:
I already reverse engineered one from client itself: https://github.com/cmangos/mangos-classic/blob/master/src/game/Entities/Relations.cpp#L688 This one can be used right away at kill time.

cala:
In the current state of the code, `Unit::IsCivilianForTarget()` will indeed work perfectly for dishonorable kills at kill time but it relies on the confusing use of `Creature::IsCivilian()` (at least in Classic context) that this PR intends to fix.

So, I suggest removing redundant `Creature::IsPvPCivilian()` from the PR and using `Unit::IsCivilianForTarget()` instead after updating it accordingly:

```diff
    // PvP-enabled enemy npcs with civilian flag
-    if (IsPvP() && GetTypeId() == TYPEID_UNIT && static_cast<const Creature*>(this)->IsCivilian())
+    if (IsPvP() && GetTypeId() == TYPEID_UNIT && static_cast<const Creature*>(this)->GetCreatureInfo()->civilian)
        return (IsTrivialForTarget(pov) && IsEnemy(pov));

Warlockbugs:
Commeted in https://github.com/cmangos/classic-db/pull/73#issuecomment-339610935

Warlockbugs:
Alternative way to solve the problem: https://github.com/cmangos/mangos-tbc/pull/214 based on separating "Civilian" and "No aggro" terms.

cala:
I'm OK with this clarification across all cores.

Warlockbugs:
Superseded by https://github.com/cmangos/mangos-classic/commit/433cc474e03ae5b32b66723b54af840459c5edf2.

PR #322, opened 2017-11-01
Update FindBoost.cmake to newest version.

0x2aff:
Resolves some issues with the newest boost library v1.65.1.


killerwife:
Isn't it enough to download latest cmake? That resolved the issue for me.

0x2aff:
The problem is when the cmangos project provides a FindBoost.cmake it will use the cmangos one not the cmake one. At least on windows.

killerwife:
For me it uses the cmake one on windows, but I use the GUI, not the inbuilt VS cmake. Ok, I guess its fair to merge this then.

PR #323, opened 2017-11-03
Love is in the air Core Support

Zingzah:
This PR adds spells triggered by city guard and civilian gossips during the Love is in the air event. The heart candy spell fix is also an addition in this PR. Three city guards gossip menus were also scripted to add checks for modelid(gender) to provide a function which isn't currently handled by the db. See also the classic db PR for more information.
Corresponding DB PR #153


killerwife:
Change tabs to spaces. Set it by default in visual studio if you are using that.

killerwife:
Those guards are doable in DB code afaik. The conditions are elementary, even if a bit complicated.

Zingzah:
Doing them through DB is definitely preferential and I tried for some time unsuccessfully.  If there is a DB solution, would love to know about it.

PR #325, opened 2017-11-14
Gandling Winter Veil Check

Zingzah:
Checks for Winter Veil and adjusts Gandling's model accordingly


cala:
Hello and thanks for this PR. However, I think this is doable by DB alone using https://github.com/cmangos/issues/wiki/game_event_creature_data. This way, the creature will switch to the model (or whatever is specified in the table) when specific event is active.

Tobschinski:
Can't tell you if they are used, but there is a spell for each boss that changes their entry during Winters Veil. For example: 26200

Zingzah:
Thanks for the comments guys.  As to the 1st one the game_event_creature_data table uses only guids.  Gandling is unique to the other hat wearing npcs in that he is summoned and doesn't have one untilthat time and the guid is always different each time.  That's why I chose to implement in sd2 for this one.  If there is a way to implement in DB I would choose that option.  As for the two spells, 26199 and 26200 I tried implementing those but upon death Gandling reverted to the old model (aura from the spells was removed on death) so I found setting the model using sd2 was the way that worked to keep the correct model displayed when killed.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/cb1114ce76c53117fc6613438578caaee0cfd3a5
Thank you.

PR #326, opened 2017-11-14
Love is in the air

Zingzah:


killerwife:
Okay, so I know I commented before, but please tell me, which code out of your PR is not doable in DB?

src/game/Spells/SpellEffects.cpp
```diff
@@ -3877,6 +3877,245 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+					return;
+				}
+				case 26678:                                 // Create Heart Candy
+				{
+
+					uint32 spellId = 0;
+
+					switch (urand(0, 7))
+					{
+					case 0:
```
killerwife: I suggest doing smth like case 0: spellId = 26668; break; saves screen space.

src/game/Spells/SpellEffects.cpp
```diff
@@ -3877,6 +3877,245 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
 
                     return;
                 }
+                case 26663:                                 // Valentine - Orgrimmar Grunt
+				{
```
killerwife: tabs

killerwife:
In regards to conditions: https://github.com/cmangos/issues/wiki/Conditions
This is the ruleset possible in the database. You are using CONDITION_ACTIVE_HOLIDAY, CONDITION_AURA and a lot of CONDITION_OR/CONDITION_AND.

Also next time, don't remake the PR, but force push into it, with the new changes.

Zingzah:
Apologies for the deletion and remaking of the PR.  I'm a real ass using Github.  It's not intuitive for me and I don't make PR's too often so I might make so more blunders in the future until I really get the hang of it.I think I end up spending more time working out how to submit than making the fixes in the 1st place. I honestly poured through the conditions to get this to work for some time so if you could give me a working example I could run with that and propagate it through the remaining content.  Also going to look into Appveyor and figure out why it keeps dissing my pr's.

Zingzah:
What I wasn't able to do with the DB was to perform a check for npc gender to know which gossip to show.  NPCs don't have gender afar as the DB goes, so an npc's model is the only attribute that defines gender.  The event needs female npcs to check whether the player has/doesn't have the cologne buff to present the appropriate gossip and male npcs to check for the presence/omission of the perfume buff.  There isn't any appropriate condition I could find in the DB to achieve this.  This problem only applies to guards in Stormwind, Thunderbluff and Org because these cities can have guards of either gender.  It's also the only instance I've come across that needs this type of check performed.  That's why only guards from these cities have been amended in the code.

There are two spells which seem to handle this function 26685 Get Gossip Female and 26684 Get Gossip Male.  I could not figure how to to trigger these (gossip_menu action script?) and code the spells in the core (using PrepareGossipMenu and SendGossipMenu functions?).  So the guards script was my solution to this issue.

killerwife:
Each creature has gender set in     SetByteValue(UNIT_FIELD_BYTES_0, 2, minfo->gender);
As a result, I suggest you add a new condition type for checking gender. That way it will be much easier to do it in the future, in case the gossip has to be added to stuff in other expansions.

Zingzah:
Really appreciate this info. I think it's the key I've been trying to find to solving this issue.  I'm just wondering whether to implement this using the spells mentioned above  (which would seem to be more blizzlike) or just make a new condition and use the DB for the gossip?

killerwife:
We have no way of knowing how that dummy works. Yes you are correct, blizzard uses it to do this, but some stuff is shrouded in so much mystery that we simply cannot know what it should do. I think it would be best to add the DB condition and do it using them. Should one day a sniff surface where this spell is indeed used, then we will add its usage. Until that time, we don't know in what context it is used (other than changing the gossip).

Zingzah:
Just changed some things around as requested.  The scripted guard gossips have been removed and support has been added for DB implementation of this feature.  DB changes will be posted  shortly to provide the complete changes.

cala:
I don't know if you noticed these and then decided to discard them, but there are two spells related to male/female gossips in Love in the Air event. I don't know more, possibly they are useless in our case.
```
Id: 26685
Name: Get Gossip, Female
Description: Used to signal a unit to adopt the female Valentine gossip.
General info
Category = 0, SpellIconId = 1, ActiveIconId = 0, SpellVisual = (0, 0)
SpellFamilyName = SPELLFAMILY_GENERAL, SpellFamilyFlags = 0x0000000000000000
SpellSchool = 0 (SPELL_SCHOOL_NORMAL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
Attributes
Attributes: 0x00000100 (SPELL_ATTR_UNK8)
Advanced info
Spell Level = 0, BaseLevel 0, MaxLevel 0, MaxTargetLevel 0
Category = 0
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange (Id 1) "Self Only"
MinRange = 0
MaxRange = 0
CastingTime (Id 1) = 0.00
Interrupt Flags: 0x00000000, AuraIF 0x00000000, ChannelIF 0x00000000
Chance = 101, charges - 0
Effects
Effect 0:
Id: 3 (SPELL_EFFECT_DUMMY)
BasePoints = 1
Targets (1, 0) (TARGET_SELF, NO_TARGET)
Effect 1:
NO EFFECT
Effect 2:
NO EFFECT

======================

Id: 26684
Name: Get Gossip, Male
Description: Used to signal a unit to adopt the male Valentine gossip.
General info
Category = 0, SpellIconId = 1, ActiveIconId = 0, SpellVisual = (0, 0)
SpellFamilyName = SPELLFAMILY_GENERAL, SpellFamilyFlags = 0x0000000000000000
SpellSchool = 0 (SPELL_SCHOOL_NORMAL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
Attributes
Attributes: 0x00000100 (SPELL_ATTR_UNK8)
Advanced info
Spell Level = 0, BaseLevel 0, MaxLevel 0, MaxTargetLevel 0
Category = 0
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange (Id 1) "Self Only"
MinRange = 0
MaxRange = 0
CastingTime (Id 1) = 0.00
Interrupt Flags: 0x00000000, AuraIF 0x00000000, ChannelIF 0x00000000
Chance = 101, charges - 0
Effects
Effect 0:
Id: 3 (SPELL_EFFECT_DUMMY)
BasePoints = 1
Targets (1, 0) (TARGET_SELF, NO_TARGET)
Effect 1:
NO EFFECT
Effect 2:
NO EFFECT
```

Zingzah:
These spells were discussed earlier in the PR with Killerwife.  The gender check for guard npcs was implemented per that discussion and it does work.  I chose to add a condition check and db entries because that's the way I managed to get it working.  I had several attempts to implement this functionality using those two spells but was unable to pull it off but wouldn't say it's not possible.

cala:
Shame on me, I still did not take the time to handle this PR. I'll try to do it before summer.

cala:
@killerwife First of all, I'd like to say that I'm happy and relieved to see you being active here for the past days. Being inactive that long was not a habit of yours and I was worried for you that you had some deep IRL concerns. I hope that everything is as fine as it can be for you.

Secondly, I'm finally reviewing this PR and the pending one DB-side. I'm pondering moving some of the spells coded in this PR in `SpellEffects.cpp` to SD2 `spell_scripts.cpp` and refactoring them into a small code as they are using the same scheme with only different parameters.

As you are our spell system wizard, I'd like to have your opinion first before leaving them here or moving them.

cala:
All merged. Thank you very much! 👍

PR #327, opened 2017-12-05
Get creature melee attack power from creature_template_classlevelstats

ktmorgan:
* According to the wiki creature melee attack power is supposed to be
calculated from this value rather than the value 'MeleeAttackPower' on
each entry in creature_template.
* Source: https://github.com/cmangos/issues/wiki/Creature_template#meleeattackpower


killerwife:
The missing stuff is bigger than I have anticipated. I will deal with it soon.

ktmorgan:
@killerwife Looks like this commit https://github.com/cmangos/mangos-classic/commit/eb9f385e180a90c3185d23a0fc3d375fa5ac5ba9, takes care of this. Thanks!

killerwife:
Ahh sorry, I didnt deal with your PR.

PR #328, opened 2017-12-16
Fix damage mod from config after CLS fields change

cala:
Damage modifier from mangosd.conf was only applied when reading old
damage fields from the DB and not used in the new CLS system


cala:
@killerwife it seems that your recent and much welcome update of the CLS system did not use the damage modifiers from config file when using the new system (only for the old).
This fixes the issue, but I'd like that you review it first.

killerwife:
Well I thought it was bcs the config was a means to alleviate the missing CLS calculations.

cyberium:
You can merge it now worries about that. It can permit some to make the world easier or harder.
@cala could you please contact me on discord or irc when you can?

cala:
Thanks, merged in https://github.com/cmangos/mangos-classic/commit/5ca36007a322795cbf54abfff7f7af0ed2be84f2

PR #333, opened 2018-02-24
Issue 1513 consume shadows

RichardGe2:
Here is how I fixed
https://github.com/cmangos/issues/issues/1513
on my server

Maybe this is not the best way to fix it... and also note that if you want to apply it on BC & WOTLK, there are certainly new ranks to add (I think : spell id  27272  , 47987  and 47988 ) ...

But it works on my Classic server, so just sharing


killerwife:
@RichardGe2 Please recheck the issue. I think this was fixed in a different way recently.

cala:
Tested and still not fixed (see https://github.com/cmangos/issues/issues/1513#issuecomment-447571887). This PR does fix the issue but I wonder if it could not be done in a more generic way.

killerwife:
It can be fixed in a more generic way, I just need to doodle a bit more with spell targeting. Which I will inevitably do later this month.

cyberium:
@killerwife If its fixed somewhere in your code is there a chance to add the better code and close this?

killerwife:
Yes, it will be fixed in an export. I fiddles with it but it needs my new targeting code.

cala:
Closed by https://github.com/cmangos/mangos-classic/commit/6ef10bbe3567ec4fd7f59e272039c53ca7bcd35a

PR #334, opened 2018-03-04
Remove faction duplicate

Phatcat:
Remove faction duplicate from creature_template

I ran this on cmangos classic database;

`select Entry, FactionAlliance as FacA, FactionHorde as FacH from creature_template Where FactionAlliance-FactionHorde!='0';`

and got this result;

billede

This guy;

http://wowwiki.wikia.com/wiki/William_Kielar

Then I ran it against cmangos tbc-db and got the same result. Still just that one guy... 

Why he needs to be a different faction to ally and to horde is beyond me but must be able to be handled in his eventAI script, and if not, then move him to SD2 where it should be manageable.

@Tobschinski @cala

### DIFF
--- a/sql/base/mangos.sql
+++ b/sql/base/mangos.sql
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `db_version`;
 CREATE TABLE `db_version` (
   `version` varchar(120) DEFAULT NULL,
   `creature_ai_version` varchar(120) DEFAULT NULL,
-  `required_z2718_01_mangos_spell_affect` bit(1) DEFAULT NULL
+  `required_z2719_01_mangos_creature_template_fixup` bit(1) DEFAULT NULL
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC COMMENT='Used DB version notes';
 
 --
@@ -1149,8 +1149,7 @@ CREATE TABLE `creature_template` (
   `ModelId2` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ModelId3` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `ModelId4` mediumint(8) unsigned NOT NULL DEFAULT '0',
-  `FactionAlliance` smallint(5) unsigned NOT NULL DEFAULT '0',
-  `FactionHorde` smallint(5) unsigned NOT NULL DEFAULT '0',
+  `Faction` smallint(5) unsigned NOT NULL DEFAULT '0',
   `Scale` float NOT NULL DEFAULT '1',
   `Family` tinyint(4) NOT NULL DEFAULT '0',
   `CreatureType` tinyint(3) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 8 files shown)


Phatcat:
Giving him different factions, according to horde and alliance also seems wrong. I get that he can change faction depending on who is in control of the towers, but that is just it, his faction should change according to who controls the towers, he should not have 2 different factions at the same time depending on who interacts with him, that is so illogical and I would love to see anyone defend this or even make any sense of it tbh.

killerwife:
FactionAlliance/FactionHorde needs to be nuked. There are plenty of places where single entry has different faction, but it should be handled by script. Since towers are a script, I don't see why the same mechanic I have been seeing all through tbc should differ here.

My approval, but for political reasons I am staying away from further discussion.

xfurry:
I think we are already handling that npc in script. It would be easy to handle his faction also in script.
Actually this is what I am doing for wotlk pvp scripts.

Tobschinski:
There's also a bunch of event npcs with the same entry but different factions. Or Nefarian in BWL ...

maybe add a field to `creature`, wouldn't be used less than model or weapon ids at least.

Phatcat:
Next thing that needs to be removed is ranged damage and ranged speed.

Tell me, how does a creature use a ranged attack? go ahead, say 'through a spell' and then go ahead and show me a ranged creature spell that isn't assigned damage, right there in the spell. 

Go ahead and tell me how a value known as 'ranged damage' is then supposed to be added on to the dedicated ranged spell damage as a custom spell cast. Is that how it's supposed to be done? 

Also it's funny how no-one wanted to address any of this the last time around;

from my chat with db-dev1:

> [15:42] DBDev1: andi havent seen yet that meleeattackspeed is different from rangedattackspeed
[15:42] DBDev1: in sniffs
[15:42] DBDev1: unless there's auras
[15:42] DBDev1: so it makes sense to only store one of them in c_template

and from my chat with db-dev2:

> [23:39] DBDev2: actually wait a minute this might be SMSG_UPDATE_OBJECT for Player not creature :D
[23:40] DBDev2: okay nvm actually. The only occurrences I see of UNIT_FIELD_MINRANGEDDAMAGE in my Full SSC ssniff
[23:40] DBDev2: are updates to players
[23:40] DBDev2: not creatures
[23:43] Me: I think you should make a comment about that on the PR so killerwife can see that.
[23:45] DBDev2: ok 
[23:51] DBDev2: talking to him now about it. I actually am looking in some other sniffs from TBC
[23:51] DBDev2: and it seems like same thing
[23:51] DBDev2: only occurs for player update

Some of you guys have sniffs. Go on. Prove me wrong and those db devs for liars.

And let's not forget this very revealing picture;

https://1drv.ms/i/s!Agpa05rA5GSClWUWxZpwY3kaNbI2

Note there is nothing with 
rangedattackspeed
rangedattackpower
rangeddamage

There is however something with min and max damage (notice there's no melee/ranged distinction.... funny, eh...) as well as attack power and attack speed...

Phatcat:
> There's also a bunch of event npcs with the same entry but different factions.

You just said it. Event npcs. so they can be handled by scripting. As it currently stands there is only 1 single entry in the entire creature_template table with different factions set in FactionAlliance and FactionHorde, and that goes both for classic as well as TBC.

Tobschinski:
This isn't about your PR, yours is fine. This is more of an extension. 
And writing scripts just because event NPCs have a different faction and not do anything else is ... meh.

Phatcat:
You are right, but it just so happens to be that there isn't a single non scripted mob which is supposed to have this very strange and unusual trait of having 2 different factions depending on which player faction is interacted with, so, I'm even willing to bet real money there never is going to be discovered any either, which should only have a script to handle it's faction and nothing else.

cala:
I'm ok with the merge of both fields into one and the handling of specific cases by scripts.

killerwife:
Closed by https://github.com/cmangos/mangos-classic/commit/b72c09425c8c68897861cf1d172f5029198c8851

PR #335, opened 2018-03-07
Replace genrevision with CMake/Git solution

Muehe:
* Removed genrevision project.
* Added revision.h.in file, from which revision.h is generated using the CMake configure_file function.
* Revision is updated when CMake is invoked.
* Removed REVISION_TIME define, as time is now included in REVISION_DATE instead.
* This fixes https://github.com/cmangos/issues/issues/1573.

Pursuant to my discussion with @cyberium I tried using the cmake [add_definitions() function](https://cmake.org/cmake/help/v3.0/command/add_definitions.html) to pass the revision to the projects, but there must have been a problem with where I put it. The errors pointed to places which @namreeb changed in his [reverted commit](https://github.com/cmangos/mangos-wotlk/commit/be7b1cdb287274173e105ceab1d404b1d7c50756#diff-a93b1e97ecb6565bd6a5f5be3b956fc8)<sup>[1]</sup>, so I guess the problem was a similar one. When I moved the function call to the `CMakeLists.txt` in one of the sub directories I got better results, but had problems with escaping the variables correctly.

The approach in this PR keeps the `revision.h` file (still not distributed and in the gitignore file), but also adds a [revision.h.in](https://github.com/Muehe/mangos-classic/blob/revision/src/shared/revision.h.in) as the input for the cmake [configure_file() function](https://cmake.org/cmake/help/v3.2/command/configure_file.html). Cmake uses this function to update/generate the `revision.h` file with values it gets from git [here](https://github.com/Muehe/mangos-classic/blob/revision/CMakeLists.txt#L273-L307).  The `revision.h.in` is also the new correct place to check in version changes to git instead of the `version.txt`.

With this patch the revision is updated every time cmake is run.  
When only make is run, the revision is not updated, unless make triggers a cmake run.  
The currently checked out commit is used.  
If there is no git, or the HEAD cannot be retrieved, the `REVISION_ID` variable is set to an error string and the `REVISION_DATE` variable is set to the zero date, which means cmake/make should not fail.

[1]: Macros `_VERSION` and `_FULLVERSION` in `src/shared/SystemConfig.h` and calls to them.


killerwife:
Closed in https://github.com/cmangos/mangos-classic/commit/f3d58c18a559cfc45f2561fe1b804a67be8e2fd5

PR #336, opened 2018-03-18
 improve first Aid - Triage quest 6622 6624

RichardGe2:
this fixes 2 things :  
- the dead corps will no longer stay on bed.  this avoid to have several bodies on the same bed
- make sure the player can finish the quest if he reaches the worst case scenario :  5 dead + 15 saved


cala:
Merged in https://github.com/cmangos/mangos-classic/commit/296f560c4263449f680417c76d19ceba8e264dee
Thank you!

PR #337, opened 2018-03-24
Issue #1466 [classic]some Boss can't be attacked

nekolepsy:
Allow Witch Doctor Zum'rah to be attacked


Warlockbugs:
Please, use new specialized helpers instead of raw flags manipulations in such cases.
```
        void Unit::SetImmuneToNPC(bool state);
        void Unit::SetImmuneToPlayer(bool state);
```

nekolepsy:
OK, so I can resubmit the pull request using SetImmuneToPlayer instead of RemoveFlag, but I want confirmation that is desired. Currently SetImmuneToPlayer isn't used anywhere in the ScriptDevAI scripts, where as SetFlag/RemoveFlag is used in quite a number of places. If SetImmuneToPlayer is the desired approach then it should probably be changed in the other places too.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/5c4cfbf89d1eb8f785c5c8d56006edb81610eff1
Thank you.

PR #338, opened 2018-05-25
fix uninitialized pointer

Tobschinski:
https://github.com/cmangos/issues/issues/1613


cala:
Good catch. I do not have this issue on mac OS, so I can't test it but you are more than probably right.

Tobschinski:
when did I ever disappoint you

cala:
> when did I ever disappoint you

I cannot remember when, but elephants are famous for their memory, not turtles. So that does not prove a thing. Maybe I forgot when in the past century of my long turtle life.

cala:
Thank you! :+1: :heart:

PR #339, opened 2018-06-03
Fixed incorrect UpdateField.h value BANKBAG_SLOT_LAST

HelloKitty:
I think this value is wrong? 

PLAYER_FIELD_BANKBAG_SLOT_1 = 0x1a8 + UNIT_END, // Size:12
PLAYER_FIELD_BANKBAG_SLOT_LAST= 0xab2 + UNIT_END,

I think the new value is correct. There are no references to this value in Cmangos though so it's just a latent fault. Not an active one thankfully.

### DIFF
--- a/src/game/Entities/UpdateFields.h
+++ b/src/game/Entities/UpdateFields.h
@@ -204,7 +204,7 @@ enum EUnitFields
     PLAYER_FIELD_BANK_SLOT_1                   = 0x178 + UNIT_END, // Size:48
     PLAYER_FIELD_BANK_SLOT_LAST                = 0x1a6 + UNIT_END,
     PLAYER_FIELD_BANKBAG_SLOT_1                = 0x1a8 + UNIT_END, // Size:12
-    PLAYER_FIELD_BANKBAG_SLOT_LAST             = 0xab2 + UNIT_END,
+    PLAYER_FIELD_BANKBAG_SLOT_LAST             = 0x1b3 + UNIT_END,
     PLAYER_FIELD_VENDORBUYBACK_SLOT_1          = 0x1b4 + UNIT_END, // Size:24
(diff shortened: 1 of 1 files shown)


HelloKitty:
Also wrong on Mangos I think, PR here: https://github.com/mangoszero/server/pull/68

HelloKitty:
0x1b2 may be more appropriate. I don't know the size of the update field, 3.3.5 on Trinitycore says it is 2 bytes.

Either way the current value is totally off, so either 0x1b2 or 0x1b3.

HelloKitty:
Switched it, I think this is correct. Was off by 1 maybe, but someone who knows the update fields better needs to chime in. Either way this value is much better than the original either way, if nobody knows.

evil-at-wow:
It's 0x1b2. You can easily calculate it because the info is already in the header:
```
PLAYER_FIELD_BANKBAG_SLOT_1 = 0x1a8 + UNIT_END, // Size:12
```
So bank bag slots start at 0x1a8, and take 12 bytes (this also tells you a single bag slot is 2 bytes, because there were 6 bank bag slots), which seems correct because 0x1a8 + 12 = 0x1b4 matching what follows (PLAYER_FIELD_VENDORBUYBACK_SLOT_1 here). So the last bank bag slot is at 0x1a8 + (12-2) = 0x1b2.

HelloKitty:
@evil-at-wow Seems right, I didn't even know how many bank slot bags there were to be quite honest. I went ahead and made the changes to 0x1b2 in the end too.

HelloKitty:
Mangos merged this PR afew days ago: https://github.com/mangoszero/server/commit/d9a11bbb2e

killerwife:
@HelloKitty i will merge this myself, but i am preparing a big export of commits. Dont worry.

PR #340, opened 2018-06-12
Implement Warden anti-cheat support

BenDol:
Full support with db migration scripts provided.

Let me know if there is anything crucial to fix and I will address it on my next weekend.

The OSX compilation is failing to find the MySQL libraries so I'm assuming its not an issue with this PR
```
CMake Error at cmake/macros/FindMySQL.cmake:151 (message):
  Could not find the MySQL libraries! Please install the
  development-libraries and headers.
Call Stack (most recent call first):
  CMakeLists.txt:231 (find_package)
```

If someone is going to work on completing this PR can you please mention that you are working on it here so if I find time over the weekends we don't double up. Thanks!


killerwife:
We had a serious talk about this last night and we decided to change the ideology surrounding anti-cheat, but this PR in this state will still not be merged.

Warden will only be merged when two rules are met, 1. it has to be an optional cmake build option, just like playerbot or extractors and 2. it has to be a separate module like realmd (because anticheat, not just warden can be resource intensive and should not in any way hinder base realm performance).

Should you be willing to work on a solution which fulfills those two points, I will try to be most helpful. Also not hardcoding modules would be a good idea.

henhouse:
I agree with rule 1, but for 2 I feel it's unnecessary considering that not only is number 1 giving you the option to entirely have it built into the core or not, but also there's a configuration variable to enable/disable it if you chose to include it.

If someone is opting to build it & enable it, they should accept the potential impacts it has on core performance in the first place, just as one would do for playerbots. I have not personally used playerbots, but I am assuming if you went crazy with the feature it could impact things just as well.

killerwife:
Yes but world should be separated into other programs as well to be fully blizzlike. The reason why I want to enforce it on admission is so that in the future one of us dont have to refactor it. Its easier to do it at the beginning.

BenDol:
@killerwife unfortunately I won't have much time to work on this PR due to work. I have put done most of the ground work for the feature, if someone wants to fork my `dev` branch and finalize it the way you are suggesting that would be good. So long as credit goes where it's due 👍

BenDol:
If someone is going to work on completing this PR can you please mention that you are working on it here so if I find time in the weekends we don't double up. Thanks!

BenDol:
@killerwife were you thinking as far as taking the anitcheat to its own database too? Or just making it a seperate program?

killerwife:
Socket architecture is simple. Main world realm decides if an anticheat check is needed, sends the required data to warden to process and then it garbles it up and sends judgement back to the main world realm, which either means nothing, kick or ban. DB is not neeeded, since this stuff renders judgement in the account section. Bans are afaik per account not per char.

cyberium:
@BenDol what the status of this?

cyberium:
No status anymore it seem...

PR #341, opened 2018-06-13
Dire Maul North Tribute

Tobschinski:
Stomper Kreeg and Cho'rush the Observer pay tribute as well.
see https://github.com/cmangos/classic-db/issues/130#issuecomment-394048171 for more information


cala:
Cool. I hope we won't find a case that requires even more instance conditions. :wink:

PR #343, opened 2018-07-21
Allowed MaxPlayerLevel config option to go beyond DEFAULT_MAX_LEVEL

katz1:
The config claims that MaxPlayerLevel should be in range 1 to 100, but in reality the max value is clamped to 60 (DEFAULT_MAX_LEVEL). I like the idea of overleveling, it opens some endgame opportunities in low population realms, so I allowed to go it up until MAX_LEVEL (which is 100 in classic, like what the config currently claims).

I tested this a bit and I've seen no problems. The player level xp requirements are already filled automatically so this works without filling the mangos.player_xp_for_level DB table manually. Attack / defense skill can go until level * 5 as I would expect.

### DIFF
--- a/src/game/Globals/ObjectMgr.cpp
+++ b/src/game/Globals/ObjectMgr.cpp
@@ -1972,8 +1972,21 @@ void ObjectMgr::LoadPetLevelInfo()
         {
             if (pInfo[level].health == 0)
             {
-                sLog.outErrorDb("Creature %u has no data for Level %i pet stats data, using data of Level %i.", itr->first, level + 1, level);
-                pInfo[level] = pInfo[level - 1];
+                if (level > 1)
+                {
+                    sLog.outErrorDb("Creature %u has no data for Level %i pet stats data, using data extrapolated from Level %i and Level %i.",
+                        itr->first, level + 1, level - 1, level);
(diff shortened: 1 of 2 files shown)


katz1:
Even though the automatically filled player_xp_for_level values are useful, the other ones are probably not:

level health/mana definitions
level stats definitions
pet level stats

These are all reusing the last value. I will investigate these a bit later, it would be nice the fill the missing values based on some extrapolation.

dagochen:
I think this is custom changes which dont belong into the classic-core here.

katz1:
I ended up adding a simple linear extrapolation based on the two last values for the missing level data and xp requirements.
I think these are good enough when the max level is raised by a sane amount (like 5-10).
People can always fill the tables as they wish if they want something more sophisticated.

cala:
Looking into this.

cala:
I agree with an optional config value allowing to have players above max level.

However, I'm uneasy with the interpolation. Unless we are 100% sure of the interpolation, this should not be hardcoded in core and should be stored DB side... and as a DB dev, I think adding custom values should be let to users responsibility and not be part of master. So, I will merge the first commit, and let an explanation in it about how to fill the missing level data in DB.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/3e21d60b8c9e95d0be0c8fbae9aa0d1f8618066a

PR #345, opened 2018-08-13
Do not show incomplete quest text for completed quests.

brotalnia:
Quests that don't require items use the RequestItemsText field to store the text that is shown when you talk to the quest giver while the quest is incomplete. Therefore the text should not be shown for them when the quest is complete. For quests that do require items, it is self explanatory.

Here is a gif to illustrate what i am talking about.
ezgif-1-3d4a9bc9c4 gif 94004b57416de09e0893bf5b0e36d12b
 https://youtu.be/jTSyexUNN9E?t=5m20s

You can see that on retail, it goes straight to offering the reward. You should not have to click "continue".

### DIFF
--- a/src/game/Entities/GossipDef.cpp
+++ b/src/game/Entities/GossipDef.cpp
@@ -721,8 +721,8 @@ void PlayerMenu::SendQuestGiverRequestItems(Quest const* pQuest, ObjectGuid npcG
         }
     }
 
-    // We may wish a better check, perhaps checking the real quest requirements
-    if (RequestItemsText.empty())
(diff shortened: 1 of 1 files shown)


cala:
The issue does exist and the proposed behaviour seems correct. I can't think of specific cases. 
 I can't tell about the code itself as I'm on vacation with little access to the Internet. So if you're okay with the code, it can be merged.

cala:
If you are busy, it can also wait that I come back. I intend to clean-up/merge Classic core PRs.

killerwife:
Once I am done setting up my new PC I might do that.

PR #346, opened 2018-11-18
Make spell proc cooldowns apply to creatures

Patman64:
This prevents spells from proc'ing at extreme rates for certain NPCs.
For example, creatures that with Lightning Shield will shed a charge on
*every single hit*, which is particularly bad against players with high
hit rate (e.g. rogues).


killerwife:
I have not forgotten about your PRs. Just reaffirming this is a correct PR and will be merged soon™.

PR #348, opened 2018-11-24
Improve EAI ranged mode behavior for main spells with min range

Patman64:
This drastically improves the AI behavior for creatures that use ranged
mode and ranged weapons. Previously they had a "dead zone" where they
would be unable to cast their shoot spell (due to the spell's minimum
range) but would fail to switch back to melee mode.

Additionally, if they used a crowd control spell (e.g. Net), they would
fail to move far enough back and would end up stuck in the dead zone.

This changes `ACTION_T_CAST` logic to properly handle `CAST_FAIL_TOO_CLOSE` to
switch back to melee mode.

This also changes `CreatureEventAI::DistanceYourself` to add additional
distance if the main spell has a minimum range.

Demo script: https://gist.github.com/Patman64/6799a59be9d2951e37dd6ee9680cd3ca
Video of AI *without* patch: https://mega.nz/#!DqoQ1KIY!r-Eqc5fNwgNa6JbQZc2xOxPATwxwLyryTt-RY2bk32A
Video of AI *with* patch: https://mega.nz/#!Tig2UAxJ!qnbetji5k3cvN5rn_1NcVwI8ecY0cQ-R3bhg30acC0Q

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -148,6 +148,7 @@ void CreatureEventAI::InitAI()
                                 m_mainSpellId = i.action[actionIdx].cast.spellId;
                                 SpellEntry const* spellInfo = sSpellTemplate.LookupEntry<SpellEntry>(m_mainSpellId);
                                 m_mainSpellCost = Spell::CalculatePowerCost(spellInfo, m_creature, nullptr, nullptr);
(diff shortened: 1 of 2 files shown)


killerwife:
Okay, small update. I rechecked client vs server min range calculations, and they were indeed incorrect. Nonetheless, this PR is still applicable. I have a buuusy week or two ahead of me, but I absolutely want to merge this after I do my exports. If I forget, poke me in two weeks. I will do my best not to.

PR #349, opened 2018-11-24
Add support for 36 slot bags

Ghaster:
Described in this issue: https://github.com/cmangos/issues/issues/1777, this PR adds support for 36 slot bags. 

As there are no 36 slot bags in Classic, you will need to add a custom bag item with 36 slots to test.

### DIFF
--- a/src/game/Entities/Bag.h
+++ b/src/game/Entities/Bag.h
@@ -22,8 +22,8 @@
 #include "Common.h"
 #include "Entities/Item.h"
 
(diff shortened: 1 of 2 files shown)


ratkosrb:
Actually, there are bags with more than 28 slots in vanilla, they are just not obtainable (and missing from cmangos db). Here they are from ItemSparse.db2 in the Blizzcon Classic Demo client.
untitled

PR #350, opened 2018-12-02
Script for Heros of Darrowshire NPC

res1n:
This should resolve https://github.com/cmangos/issues/issues/1655

In order to test you must set the creature to use the script, I haven't written the SQL update for the db yet.

SELECT * FROM creature_template where entry = 11064; 
Set script_name to npc_darrowshire_spirit

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/eastern_plaguelands.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/eastern_plaguelands.cpp
@@ -31,6 +31,7 @@ EndContentData */
 
 /*######
 ## npc_eris_havenfire
+## npc_darrowshire_spirit
 ######*/
 
 enum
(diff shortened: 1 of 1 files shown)


cala:
Thank you for this PR. It reminds me of the way it was done in Elysium/Nostalrius core. Unfortunately, this does not match how it was done in retail as Darrowshire Spirit should despawn once talked to, removing the need of making it unselectable.

Incidentally, I was working on two quests from the Darrowshire quest-line today, including this one and you can find the fix [here](https://github.com/cmangos/classic-db/commit/f233563681587077e99782a2080cdcdd3542f9cc) with the source I used.

Thanks again.

res1n:
You're welcome.  I did port it from that core good eyes, I tried it out as a learning experience as I noticed their sd2 is setup in an interesting way.  I'll take a peek at some others scripts in the backlog.  Thanks for linking me the fix i'll check it out, had no idea the DB had that much control much.  Cheers!

PR #351, opened 2018-12-04
ExtractResources.sh CPU conditional statement fix

ekg98:
Fixed CPU select conditional statement.

### DIFF
--- a/contrib/extractor_scripts/ExtractResources.sh
+++ b/contrib/extractor_scripts/ExtractResources.sh
@@ -91,8 +91,7 @@ then
   echo "How many CPUs should be used for extracting mmaps? (1, 2, 4, 8)"
   read line
   echo
-  if [[ ($line == 1) || ($line == 2) || ($line == 4) || ($line == 8) ]]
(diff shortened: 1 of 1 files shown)


killerwife:
I am gonna ask what issue this fixes. I know very little about shell script syntax.

ekg98:
For some reason anytime I execute the script it does not like the previous conditional.  It gives errors.  Looks like somebody for got and was programming c at the time or something.

./ExtractResources.sh: 94 ./ExtractResources.sh Syntax error: word unexpecting (expecting ")")

I am reading the 9th edition of the linux bible.  Its saying -eq and = are the same in some regards.  I could have just used one equals.

[var1 = var2]
[var1 -eq var2]

This book is saying that both these statements are equal.  I have only tried the second one.

cala:
Kind of busy lately but I have not forgotten this PR that was successfully tested with mac OS. I should be able to test it with Windows bash-shell on Monday.

cala:
Merged, thank you!

PR #352, opened 2018-12-11
Fixed: Create user failed

dvdvideo1234:


cala:
Correct: now that we use different database names for each core, we need this change if we still want to be able to set up different versions on the same host.

PR #353, opened 2018-12-11
Fixed: ERROR 1136 (21S01) at line 1257: Column count doesn't match va…

dvdvideo1234:
…lue count at row 1


cala:
Thank you for this PR. The change was already done elsewhere, so I'm closing this.

PR #354, opened 2018-12-20
Don't interrupt a healer when an enemy target dies

ainslievaux:
This commit fixes an issue where a healer bot's spellcast is interrupted whenever an enemy target dies.

I've tried setting an appropriate `m_ai->SetIgnoreUpdateTime()` in the relevant class AI files but it still doesn't work (even if you give it a ridiculous time like 5 seconds) and I can't figure out why.


cala:
Thank you for this PR. That is a good addition, however some healers also have a handful of agressive spells in their rotation, so this would let them finish casting it. I think the check would be more efficient against spell effects like heal or periodic heal.

cala:
Added in a different way in https://github.com/cmangos/mangos-classic/commit/2b69d3c4ee8ba7f32e11a289318174a5f390e72b
Thank you for your contribution! :+1:

PR #357, opened 2019-01-19
Misc

Phatcat:
Corrections based on this analysis;

https://medium.com/@Coder_HarryLee/checking-the-world-of-warcraft-cmangos-open-source-server-5405f364fb32

### DIFF
--- a/contrib/mmap/src/MapBuilder.cpp
+++ b/contrib/mmap/src/MapBuilder.cpp
@@ -276,7 +276,7 @@ namespace MMAP
         m_terrainBuilder->loadMap(mapID, tileX, tileY, meshData);
 
         // get model data
-        m_terrainBuilder->loadVMap(mapID, tileY, tileX, meshData);
+        m_terrainBuilder->loadVMap(mapID, tileX, tileY, meshData);
 
         // if there is no data, give up now
         if (!meshData.solidVerts.size() && !meshData.liquidVerts.size())
--- a/contrib/vmap_extractor/vmapextract/vmapexport.cpp
+++ b/contrib/vmap_extractor/vmapextract/vmapexport.cpp
@@ -358,7 +358,7 @@ bool processArgv(int argc, char** argv)
             {
                 hasInputPathParam = true;
                 strcpy(input_path, argv[i + 1]);
-                if (input_path[strlen(input_path) - 1] != '\\' || input_path[strlen(input_path) - 1] != '/')
+                if (input_path[strlen(input_path) - 1] != '\\' && input_path[strlen(input_path) - 1] != '/')
                     strcat(input_path, "/");
                 ++i;
             }


CabrasDaPeste:
I remember that I saw this article a long time ago, nice changes.

VladimirMangos:
arguments order must be cerreful checked before apply and possible requred raname/order changes in other code parts.

Phatcat:
@VladimirMangos 
I've been looking through the core and  LoadVMap very much does expect an order of arguments going: ID, TileX, TileY, Mesh.

Here is the function in the header file with the correct order of arguments;
https://github.com/cmangos/mangos-classic/blob/2b8ba9eef1dc1f0b8a37f8b90d9955211830e511/contrib/mmap/src/TerrainBuilder.h#L88

And here is the function itself, also with correct order;
https://github.com/cmangos/mangos-classic/blob/2b8ba9eef1dc1f0b8a37f8b90d9955211830e511/contrib/mmap/src/TerrainBuilder.cpp#L563

As we can see in the function there is no compensation for mixed up TileX and TileY arguments. It expects the correct order and nothing else.

Here is another area which calls LoadVMap with the correct order of arguments:
https://github.com/cmangos/mangos-classic/blob/f7a04c831438f8bf975a73f9c6cddb5d5f6b9b1c/contrib/recastdemomod/Source/GeomData.cpp#L202

I really can't find where there'd be a need for a rename/reorder.

Phatcat:
Yeah no need to apologize, rather check and double check and tripple check to be sure. I've been looking through both MapBuilder and TerrainBuilder and I can't find anything relevant to change. Maybe I'm not looking the right places.

Warlockbugs:
This likely will require map data version bump and rebuilding from what i see? Tagging @cyberium

cyberium:
@VladimirMangos is correct. The change is not so trivial. So @Phatcat after that change try to build new map and load them in the server. You'll have a big problem. I did try long time ago to fix that but i gave up. We are missing something here.

Phatcat:
Alright, currently extracting resources via this branch, will report back with my results.

Phatcat:
You are right something is very much wrong. Outside in the world mobs behave as expected, but in the water or inside caves mobs seem to wanna move to the top (in caves they move above and on top of the cave with every step they take, in the water they move to be above the surface line.. So as @cyberium said this change alone brings about trouble, but I am not quite ready to give up on it so I'll try and investigate what exactly is going on, if it's with the pathfinding, the mapbuilding or what ever else it could be.

I have compressed and made the resources available (temporarily) [here](https://www.dropbox.com/s/fy954qma2urmatf/branchresources.zip?dl=0) for anyone else.

Phatcat:
The swapped order of arguments is proving too much of a challenge for me so I am closing this PR, at least for now.

PR #359, opened 2019-01-23
Fix always true if statement

Phatcat:
This statement would always be true.

This puts it in line and smiliar functionality with the statements already fixed at line 125 and 231 in MapTree.cpp

Ref: https://medium.com/@Coder_HarryLee/checking-the-world-of-warcraft-cmangos-open-source-server-5405f364fb32

the re-extracted resources are made temporarily available [here](https://www.dropbox.com/s/6hcxi94n5l6yovp/fixifResources.rar?dl=0)

### DIFF
--- a/contrib/vmap_extractor/vmapextract/vmapexport.cpp
+++ b/contrib/vmap_extractor/vmapextract/vmapexport.cpp
@@ -358,7 +358,7 @@ bool processArgv(int argc, char** argv)
             {
                 hasInputPathParam = true;
                 strcpy(input_path, argv[i + 1]);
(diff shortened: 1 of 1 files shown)


cala:
I suppose with have the same issue on the three cores?

Phatcat:
I'd think so yes but i havn't actually checked, can't check right now.

evil-at-wow:
I can confirm this issue is present in the TBC core as well, so most likely it's in all three.

Re-extracting resources will most likely make no difference, because this is only used in code to handle an explicitly passed input path (-d option), which is rarely used I think (ExtractResources.sh also doesn't use it). And I would assume one would notice it quickly when the MPQs aren't found by the extract tool to extract anything...

cala:
Thanks everyone. Merged.

PR #361, opened 2019-01-24
remove remaining unused dep TBB

Phatcat:
Remove the remaining part of unused dep TBB

It has already been removed from the other repos, even the cata one..


cala:
I remember some discussion about not removing this though being unused. @cyberium ?

killerwife:
vengeance still uses it cos its more stable than stdmalloc for all its worth

Phatcat:
It's only still present in classic, it is gone from tbc, it is gone from wotlk, it is gone from cata even.

Also it's all cmake and make files and visual studio project files, I don't see much use of it. It's not even really much of tbb is it?

I did run these changed by cyberium before opening the pr as well actually

evil-at-wow:
> more stable than stdmalloc

Could you elaborate a bit, killerwife? More scalable or even faster, I could understand, but more stable? It's not like the standard library allocator is crashing a lot or causing lots of problems as far as I know (at least not on any platform I know)?

cala:
> I did run these changed by cyberium before opening the pr as well actually

I wasn't aware. 😉 
Merged, thanks!

PR #363, opened 2019-01-25
add git_id to cmake build options (also clean out unneeded visual studio project files)

Phatcat:
pretty much the title


cala:
As I don't use Visual Studio, I'd like to have feedback from another VS user or at least an explanation regarding this change before merging it.

Phatcat:
I'm a visual studio user, this isn't needed since we use cmake to make our own visual studio project files now. This is leftovers from back when cmangos provided the project files.

killerwife:
Last time I checked, git_id was not in cmake. So this needs a cmake addition before its removed.

Phatcat:
It probably does yea.

Phatcat:
Now git_id is a selectable build option in cmake.

PR #365, opened 2019-01-30
Clean PetAI a bit

Phatcat:
remove a bit of unused mini_pet code from PetAI

Forgot to mention in the commit that mini pets use GuardianAI and can never have their AI's changed either since they are immune to npcs and immune to players, probably what causes the issue of them being immune to npc aoe's (which if I recall correctly was reported as they shouldn't be immune)

Also we already set mini pets to be passive on creation through spellEffects;

```
if (UnitAI* critterAI = critter->AI())
    critterAI->SetReactState(REACT_PASSIVE);
```

https://github.com/cmangos/mangos-classic/blob/master/src/game/Spells/SpellEffects.cpp#L5298

also nest pet attack from behind inside a config bool and turn it off by default on classic since it was added on retail in 2.3, should be turned on by default on tbc and wotlk.


src/game/AI/BaseAI/PetAI.cpp
```diff
@@ -40,17 +40,9 @@ PetAI::PetAI(Creature* creature) : UnitAI(creature), m_creature(creature), inCom
...
-        case HUNTER_PET:    //hunter pets attack from behind
-            m_attackAngle = M_PI_F;
-            break;
-        case MINI_PET:
-            SetReactState(REACT_PASSIVE);
-            break;
-        default:
-            break;
-    }
+    //hunter pets attack from behind
```
Phatcat: This thing here is wrong for classic, you mentioned something about that @Warlockbugs ? it should be removed entirely then?

Warlockbugs: https://wow.gamepedia.com/Patch_2.3.0#General
> Pets will try to get behind their targets when engaging in melee combat.

We can comment this functionality out for the time being and later put under config value which would default to disabled for vanilla. This kind of QoL is purely server-side and can be very useful, simply removing feels like a waste.

src/game/AI/BaseAI/PetAI.cpp
```diff
@@ -59,13 +51,6 @@ PetAI::PetAI(Creature* creature) : UnitAI(creature), m_creature(creature), inCom
             m_meleeEnabled = false;
             break;
     }
-
-    // TODO: Remove this when unit support is removed
```
killerwife: Unit support is still not removed. Players use this AI when in charm. The code below is still valid.

src/game/AI/BaseAI/PetAI.cpp
```diff
@@ -40,9 +40,8 @@ PetAI::PetAI(Creature* creature) : UnitAI(creature), m_creature(creature), inCom
     m_AllySet.clear();
     UpdateAllies();
 
-    /* hunter pets attack from behind (tbc and later)
-    if (((Pet*)creature)->getPetType() == HUNTER_PET)
-        m_attackAngle = M_PI_F; */
+    if (((Pet*)creature)->getPetType() == HUNTER_PET && sWorld.getConfig(CONFIG_BOOL_PET_ATTACK_FROM_BEHIND))
```
Phatcat: @Warlockbugs - this should be good now, yea?

Warlockbugs: Any pet spawned by a player should do this, not just hunter pets.

Phatcat: Yeah, that is what I was thinking as well, but guardians behave like their own entities, you don't control them, I think it's only for controlled pets. So either remove the (Pet*)creature)->getPetType() == HUNTER_PET part entirely or replace with (Pet*)creature)->isControlled()?

Warlockbugs: `static_cast<Pet*>(creature)->isControlled()` :smiley:

Phatcat: Hmm, why dynamic and not static cast?

evil-at-wow: In this case your safe downcasting is rather worthless, because you're using the result unconditionally. If ```creature``` isn't really a pointer to Pet, you'll get a nullptr which you then dereference. You could as well use a static_cast then, because the end result is going to be the same anyway: :boom:

Warlockbugs:
Ready for merge. Requires a final touch so it can be merged from github without manual alterations to the branch:
- [ ] Update definition of `_MANGOSDCONFVERSION` in src/shared/SystemConfig.h
- [ ] Update config file version `[MangosdConf]` to the updated value
- [ ] Squash all commits into one

Phatcat:
> Squash all commits into one

Why is that a requirement? These are 2 different commits doing 2 seperate things.. Does it not just obscure things more? I know these aren't major changes so it'd be easy to decipher what is what, but still.

Also isn't it easier to port this before updating definition of _MANGOSDCONFVERSION and config file version [MangosdConf]?

PR #124, opened 2016-05-02
Pet Patch 1.0: Amazing fun times with pets (somewhat satisfied for now)

Phatcat:
Contains a bunchload of fixes and tweaks to pet and summon functions.

[Referencing PR 116](https://github.com/cmangos/mangos-classic/pull/116)

OBS! I will be squashing and organizing this patch into proper commits before merging. I won't do that until I am completely done, though. Commits provided by other developers will be left entirely in the state they are in this pr (ie I won't be squashing other people's work, nor change their commit titles or comments).

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-classic/124)

PR #366, opened 2019-02-02
Update/fix xp gain calculations

Phatcat:
Update xp formulas and rewrite to be based off of
https://wowwiki.fandom.com/wiki/Formulas:Mob_XP?oldid=228414
which is based off of
https://web.archive.org/web/20060722002320/http://mosa.unity.ncsu.edu/WoW/library/xpgrid.html

Rewrite xp formulas to support units and not just players (in support of pets)

Don't factor in any group rate when giving out reputation as there is no group rate bonus for that, it's equally divided by all the members regardless of level or anything else

Calculate in level difference between the pet and the killed target for xp gain instead of the pet's owner and killed target

convert interger types to float before using them in point decimal calculations. Also round up when giving xp so use std::round before converting back to uint32

Contributes to https://github.com/cmangos/issues/issues/931
> Pet does not gain exp as it should; it gains exp based on the level difference between its owner and the kill, not itself and the kill as it should.

### DIFF
--- a/src/game/Tools/Formulas.h
+++ b/src/game/Tools/Formulas.h
@@ -352,7 +352,7 @@ namespace MaNGOS
 
             xp_gain *= target->GetCreatureInfo()->ExperienceMultiplier;
 
-            return (uint32)(std::round(xp_gain * sWorld.getConfig(CONFIG_FLOAT_RATE_XP_KILL)));
+            return (uint32)(std::nearbyint(xp_gain * sWorld.getConfig(CONFIG_FLOAT_RATE_XP_KILL)));
         }
 
         inline float xp_in_group_rate(uint32 count, bool isRaid)
--- a/src/game/Tools/Formulas.h
+++ b/src/game/Tools/Formulas.h
@@ -297,61 +297,62 @@ namespace MaNGOS
             return GRAY;
         }
 
-        inline uint32 GetZeroDifference(uint32 pl_level)
+        inline uint32 GetZeroDifference(uint32 unit_level)
         {
-            if (pl_level < 8)  return 5;
-            if (pl_level < 10) return 6;
-            if (pl_level < 12) return 7;
-            if (pl_level < 16) return 8;
-            if (pl_level < 20) return 9;
-            if (pl_level < 30) return 11;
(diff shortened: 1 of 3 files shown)


src/game/Tools/Formulas.h
```diff
@@ -297,61 +297,62 @@ namespace MaNGOS
...
-                return (pl_level * 5 + nBaseExp) * (ZD + mob_level - pl_level) / ZD;
+                uint32 ZD = GetZeroDifference(unit_level);
+                uint32 nLevelDiff = unit_level - mob_level;
+                return nBaseExp * (1.0f - (float(nLevelDiff) / ZD));
             }
             return 0;
         }
 
-        inline uint32 Gain(Player* player, Creature* target)
+        inline uint32 Gain(Unit* unit, Creature* target)
```
VladimirMangos: Masybe possible use Unit const* for avoid static casts?

Phatcat: Not quite sure what you're saying, so if I change this inline to use Unit const* instead of Unit* then I don't have to do static casting from the other functions calling this? so I can just pass whatever instead (pet/player)? Wouldn't I still have to cast them as unit?

VladimirMangos: Yes, if in function 'unit' not use non-const functions

Levitanious: Владимир, можешь по-русски, а я переведу? :>

VladimirMangos: Why need any explicit cast code in this function call? Just pass Pet* to arg.

Phatcat: Really?

VladimirMangos: I see you drop explict casts. Now this is more clean code.

Phatcat:

Phatcat:
Took me 3 years but finally got around to my old TODO; https://github.com/cmangos/mangos-classic/pull/124/commits/b22343b1e448d998ba2c4302e174fdd11ddf88e2#diff-81afb145f9a28dec972d3fc484bebf40R1111

PR #367, opened 2019-02-03
Pass real killer to client and not BeneficiaryPlayer

Phatcat:
Properly inform client who really carried out the killing blow, the one who dealt the damage calling for the Unit::Kill function, and not the BeneficiaryPlayer, the one who stands to get the kill credit (loot, honor, quest count, etc)

Attempt at fixing https://github.com/cmangos/issues/issues/87


killerwife:
I do not have time for this, but since this is packet sending alteration, it needs verification from 1 of the 2 following:
- Client decompilation
- Sniff verification
It is unlikely I will have time for either in the next month.

Phatcat:
Yeah this change isn't needed.

PR #369, opened 2019-02-05
Allow linking graveyards by area id.

norcimoix:
After dying in frostmane hold and spawning in coldridge valley I suspected something was wrong with how a graveyard is selected. I found [this comment](https://www.wowhead.com/quest=287/frostmane-hold#comments:id=215696) on wowhead: 

>The nearest graveyard is in Kharanos, so if you die it is an annoyingly long run back.

This indicates that there is indeed something wrong with how the graveyards are selected. Even some areas more north of frostmane hold make you spawn in coldridge valley instead of Kharanos. I suspected that subzone/area id is related.

It's unlikely that frostmane hold is linked to the kharanos graveyard. Instead it's more likely that the coldridge valley graveyard is linked to coldridge valley and coldridge pass but NOT to dun morogh. I think this because north of frostmane hold there are areas that have no subzone but currently still make you respawn in coldridge valley. This makes no sense if frostmane hold itself didn't even make you respawn there (as the comment suggests).

I've implemented a patch that allows you to link graveyards to a subzone (in addition to linking to zones as it currently allows). Here's the commit log:

    Changes how `game_graveyard_zone`.`ghost_zone` is interpreted. This
    column can now also be an area id.
    
    When looking for a nearby graveyard the server will first attempt to
    spawn at graveyards linked to your current area and only if that fails
    continue to the graveyards linked to your current zone.

And here is an SQL file to test the changes:

    -- Unlink the graveyard in coldridge valley from dun morogh
    DELETE FROM `game_graveyard_zone` WHERE `id`=100;
    
    -- Link the graveyard to the subzones. TODO: alliance?
    INSERT INTO `game_graveyard_zone` (`id`, `ghost_zone`, `alliance`) VALUES
        (100, 132, 0),  -- coldridge valley
        (100, 800, 0);  -- coldridge pass

It's possible that all closed off starting zones have their own exclusive graveyard that is not linked to the wider zone, but I haven't investigated that at all. I also don't know if the graveyard should be exclusive to alliance, but it was currently available to all factions so I haven't changed that in this SQL file.

This is the first patch I've written for mangos so I'm open to any input. If something is unclear I can clear that up or if something needs to be changed I can do that too.

### DIFF
--- a/sql/updates/mangos/z27xx_01_mangos_graveyard_map_links.sql
+++ b/sql/updates/mangos/z27xx_01_mangos_graveyard_map_links.sql
@@ -1,0 +1,12 @@
+-- Rename ghost_zone to ghost_loc because ghost_loc can now also refer to a map, not just an area
+-- Add link_kind to indicate what `ghost_loc` refers to.
+-- There are two
+-- kinds:
+--   0 --> `ghost_loc` refers to an area
+--   1 --> `ghost_loc` refers to a map
+
+ALTER TABLE `game_graveyard_zone`
+    CHANGE `ghost_zone` `ghost_loc` MEDIUMINT UNSIGNED NOT NULL,
+    ADD COLUMN `link_kind` TINYINT UNSIGNED NOT NULL DEFAULT 0 AFTER `ghost_loc`,
+    DROP PRIMARY KEY,
+    ADD PRIMARY KEY(`id`, `ghost_loc`, `link_kind`);
--- a/src/game/Chat/Level3.cpp
+++ b/src/game/Chat/Level3.cpp
@@ -3710,7 +3710,7 @@ bool ChatHandler::HandleLinkGraveCommand(char* args)
         return false;
     }
 
-    if (sObjectMgr.AddGraveYardLink(g_id, zoneId, g_team))
+    if (sObjectMgr.AddGraveYardLink(g_id, zoneId, GRAVEYARD_AREALINK, g_team))
         PSendSysMessage(LANG_COMMAND_GRAVEYARDLINKED, g_id, zoneId);
     else
         PSendSysMessage(LANG_COMMAND_GRAVEYARDALRLINKED, g_id, zoneId);
--- a/src/game/Chat/Level3.cpp
+++ b/src/game/Chat/Level3.cpp
@@ -3735,6 +3735,7 @@ bool ChatHandler::HandleNearGraveCommand(char* args)
 
     Player* player = m_session->GetPlayer();
     uint32 zone_id = player->GetZoneId();
+    uint32 area_id = player->GetAreaId();
 
     WorldSafeLocsEntry const* graveyard = sObjectMgr.GetClosestGraveYard(player->GetPositionX(), player->GetPositionY(), player->GetPositionZ(), player->GetMapId(), g_team);
 
@@ -3742,7 +3743,10 @@ bool ChatHandler::HandleNearGraveCommand(char* args)
     {
         uint32 g_id = graveyard->ID;
 
-        GraveYardData const* data = sObjectMgr.FindGraveYardData(g_id, zone_id);
+        GraveYardData const* data = sObjectMgr.FindGraveYardData(g_id, area_id);
+        if (!data || (g_team != TEAM_BOTH_ALLOWED && data->team != g_team && data->team != TEAM_BOTH_ALLOWED))
+            data = sObjectMgr.FindGraveYardData(g_id, zone_id);
+
         if (!data)
         {
             PSendSysMessage(LANG_COMMAND_GRAVEYARDERROR, g_id);
--- a/src/game/Globals/ObjectMgr.cpp
+++ b/src/game/Globals/ObjectMgr.cpp
@@ -5436,12 +5436,6 @@ void ObjectMgr::LoadGraveyardZones()
             continue;
         }
 
-        if (areaEntry->zone != 0)
-        {
-            sLog.outErrorDb("Table `game_graveyard_zone` has record subzone id (%u) instead of zone, skipped.", zoneId);
-            continue;
-        }
-
         if (team != TEAM_BOTH_ALLOWED && team != HORDE && team != ALLIANCE)
         {
             sLog.outErrorDb("Table `game_graveyard_zone` has record for non player faction (%u), skipped.", team);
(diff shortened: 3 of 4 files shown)


Warlockbugs:
It is not limited to opposing faction's areas. I was reworking mirror timers (such as Fatigue), and i noticed that graveyard selection issue applies to it as well. Once you get off the shore far enough, upon death you will release spirit directly on the corpse. We can't obviously link oceans to any specific graveyards.

norcimoix:
That's interesting, many large areas of deep sea do have a zone assigned to them but some don't indeed. I found at least one location off the coast of [Kalimdor](https://user-images.githubusercontent.com/47334129/52330039-a1833380-29f4-11e9-9087-ea89d8e08f7c.png) and one zone off the coast of [Eastern Kingdoms](https://user-images.githubusercontent.com/47334129/52330066-b4960380-29f4-11e9-8e98-a85c85e79ccd.png).

How would you propose this is resolved? I don't see why we can't link the graveyards of coastal facing zones to the larger map.

Phatcat:
> We can't obviously link oceans to any specific graveyards.

Well, if we can't, then getting this behaviour right may be difficult;

> You can also use a trick to travel there from Darkshore. If you keep swimming far north from the northern end of its coast, the atmosphere will change to green hills and your zone will be 'Moonglade'. You can enter the land, but not climb the mountains. Enter the water again, but take care that you do not reach 'The Veiled Sea', you must pass through 'The Veiled Sea' however on your way north. Drown yourself, and you will be in Moonglade. 

https://wow.gamepedia.com/Moonglade#Getting_there

This would very much indicate that it's not simply picking the nearest graveyard or this part;

*Enter the water again, but take care that you do not reach 'The Veiled Sea'* would be needless.

norcimoix:
That moonglade behavior already works correctly. Since you are considered in the moonglade zone it will select the nearest graveyard linked to moonglade. The issue Warlockbugs points out is for locations in the world that are not linked to any zone at all.

Phatcat:
That is not the part I was trying to highlight, it's the part I picked out and repeated under. Since if you do enter 'The Veiled Sea' again, but you having just been in moonglade, does it also correctly send you back to darkshore or would it send you to the moonglade graveyard? - does the server ever consider you crossing back over to darkshore when entering the veiled sea after having just zoned to moonglade?

killerwife:
Then how about custom resolution of the "sea" zone. Locate closest possible GY to your current location which is valid for team, and go there. I would guess the same issue could happen in Twisting Nether maybe?

cyberium:
First -> Well done for your first PR. You know how to talk with devs!

So for this PR its ok and should also fix the ocean but need more data wich will be a hard task to fill all those cases.

norcimoix:
I think it makes sense to have to link graveyards to the entire map too.

 * Find the closest graveyard based on subzone, continue if none are found
 * Find the closest graveyard based on zone, continue if none are found
 * Find the closest graveyard based on map, spawn on corpse if none are found

It probably does make a lot of sense to link graveyards from the coastal zones to the entire map. That way if you die in the sea you'll spawn in some graveyard with access to the sea. Most of the sea _does_ have a zone attached to it, so it would only be used in cases where there is no zone.

norcimoix:
I'll implement the map linking as soon as I have some free time and add it to this pull request as a separate commit. We can still decide to squash it or leave it out later then. The change should be minor but it would fix the graveyard selection in the ocean and also make the moonglade behavior as described by Phatcat work like it did on retail servers.

norcimoix:
While looking into how to implement the map_id linking I bumped into an issue. Adding `area_id` was relatively simple because `area_id` and `zone_id` are technically the same kind of id so they never overlap. That's not true for `map_id`, Kalimdor has `map_id` 1 and `Dun Morogh` has `area_id` 1.

I propose the following resolution to this problem:

 * Add a column `id_kind` that dictates whether or not the linked id is a map or an area. This is better than adding a `map_id` column because map and area links need to be mutually exclusive, using the proposed column makes that clear.
 * Because id_kind is a binary option (either area or map) we can combine both columns into a single `uint32` when loading the table.  We can use the most significant bit of the id to signal if it's a map (MSB = 1) or an area id (MSB = 0). The advantage of this is that you require almost no changes to the core because that combined id can be the index in the multimap. All current indices to areas and zones remain the same so no existing code will break because of this change.

I'm not aware of special meaning being assigned the `area_id` or `map_id` range starting at `0x80000000`. Should that be the case then my proposal would need to be tweaked.

VladimirMangos:
Maybe just use negative for mapid?

norcimoix:
@VladimirMangos: In the database or in the multimap? I spoke about this with @cyberium and he brought up that it's not very clear if the database column has a different interpretation like that for negative values. I tend to agree with that, which is why I propose to add a second column that makes it clear you need to choose between map or area.

Currently all those values are unsigned in both the core and the database. It's probably a good idea to keep them unsigned so using the MSB to split the domain of the ids in two seems appropriate.

VladimirMangos:
I agree that special type field is more clear way. Just note that diff interpretation signed values wide used in db/code.

norcimoix:
To easily test this patch add [these three test sql files](https://gist.github.com/norcimoix/af9d294a26f6d8469cafcd0a6f46c61a) to the dev/ directory in your classic-db repo.

The first file contains data to test the first area link patch in dun morogh. The second file updates the database to have the correct format for the map patch and the third file adds map graveyard links to test dying in [this place](https://user-images.githubusercontent.com/47334129/52330039-a1833380-29f4-11e9-9087-ea89d8e08f7c.png).

I have tried my best to test the patch in various places but it would be nice if someone could review it.

@cyberium  Lastly I'm not sure what to do with the sql update, how are those files handled? What decides the filename? For now I placed it in the sql/updates directory with a dummy name. If someone could instruct me how I can correctly do that sql update that would be awesome.

cala:
@cyberium : please, leave you keg of beer and come work on this. 👅

Zarant:
I noticed while playing classic some exceptional graveyard behavior:

- The Twilight Vale graveyard in southern darkshore is a valid graveyard for people in ashenvale
-  If you die in The Barrens you can get ported to orgrimmar/razor hill if one those 2 GYs are the closest to you 
- Horde players can get ported to the Bulwark graveyard in Tirisfal if they die in Western Plaguelands.
- Ironforge GY despite being located in dun morogh can only be accessed by people dying in IF or right outside the gates of IF, otherwise the Kharanos GY is used

cala:
Poking @cyberium with a stick. 😸

cyberium:
Well graveyard does not work as it should. But this pr make it possible to fix some issue with it.
In the wait of a full rewrite of graveyard system i am merging this.

Fallback to nearest graveyard faction usable should always work and only some exception should be done.
- Like maybe for start area only linked to start zone
- pvp graveyards that are currently set to other team
- some exception here and there where nearest graveyard is in another zone.

PR #370, opened 2019-02-06
Set autocast default to enabled for pets

Phatcat:
Autocast spells should be set to autocast by default when creatures are first summon/tamed
Touched up the same functionality for charmed units (they determined autocast spells in a really weird manner)

Proof; https://www.youtube.com/watch?v=ILsu79gPj00 (at 4:29)

Addresses the last issue remaining mentioned in https://github.com/cmangos/issues/issues/882
> the pet bar doesn't actually reflect what the default settings are properly. Waterbolt is set to autocast and Freeze is set to manual cast by default,


Phatcat:

cala:
Merged, thank you! :+1:

PR #371, opened 2019-02-06
CMakelist alterations for git_id

Phatcat:
The folder architecture stuff isn't needed and neither is the properties link flag since it'll fail linking on windows anyway

I based the inclusion on how recastdemomod is included but I should have based it on how extractors are included, which it is now

Ref: https://github.com/cmangos/mangos-tbc/commit/30a29c5593b2bf4e4d3c1dea86fd4dac3dd7a44c#commitcomment-32214692

I did test it underway and everything was working fine but I must have added the 
```
  # Define windows application type to not see a console
  set_target_properties(git_id PROPERTIES LINK_FLAGS "/SUBSYSTEM:WINDOWS")

```
after and forgot to delete cache and reload cmake.. It should be working as intended with this change.

### DIFF
--- a/contrib/git_id/CMakeLists.txt
+++ b/contrib/git_id/CMakeLists.txt
@@ -19,20 +19,12 @@ cmake_minimum_required(VERSION 2.8)
 include(../../cmake/common.cmake)
 include(CheckPlatform)
 
+
 add_executable(git_id git_id.cpp)
 
-if(WIN32 AND MSVC)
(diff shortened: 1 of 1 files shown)


Phatcat:
@Warlockbugs I messed up a bit with my git_id PR, this needs to be merged in to all 3 cores to actually get git_id building to work in visual studio on windows

Phatcat:
Thank you.

PR #372, opened 2019-02-07
Fix player-controlled pet spellcasting by manual casting

Phatcat:
Stored pet spells don't execute properly because the spell is never applied the targets properly, but autocast spells do, so move stored spells further down the code and handle it along with autocast spells

Active pet spells cast manually have the same issue
in pethandler we never set a target for the spell so looking for a target in the spell is silly, remove that, we need to also set the target of the spell before casting it or the pet will cast it on itself.

This solves 
https://github.com/cmangos/issues/issues/1036
and https://github.com/cmangos/issues/issues/870

### DIFF
--- a/src/game/AI/BaseAI/PetAI.cpp
+++ b/src/game/AI/BaseAI/PetAI.cpp
@@ -150,52 +150,34 @@ void PetAI::UpdateAI(const uint32 diff)
         }
         charminfo->SetIsRetreating();
     }
-    else if (charminfo->GetSpellOpener() != 0) // have opener stored
-    {
-        uint32 minRange = charminfo->GetSpellOpenerMinRange();
-        victim = m_unit->getVictim();
 
-        if (!victim
-                || (minRange != 0 && m_unit->IsWithinDistInMap(victim, minRange)))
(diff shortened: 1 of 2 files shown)


Phatcat:

Phatcat:
Solves 2 open issues (somewhat old ones even) and does it by pulling out more code than is introduced, and most of the code removed is code I personally wrote in back in 2016.

So... why has it gone to a stale? no-one can be bothered to do a 2 min test to confirm?

cala:
Hmm... Pet code is not really something I have a good knowledge of, but I'll have a look at this this week.

cala:
Next on my list. Sorry: TBC-DB porting and Maws took more time than expected…

cala:
Merged, thank you ! @Phatcat : this does apply to TBC as I understand it, do you think this is also required on WotLK ? (based on what I understand, I think it does)

Phatcat:
Yes, this applies to all the cores, as far as I know pets still had both passives and autocastables all throughout the 3 OG expansions. And since I added the stored pet spell capabilities to all 3 cores back in 2016, all the cores need this update.

PR #373, opened 2019-02-11
CLUCK!

Phatcat:
Thanks to @MantisLord who was invaluable in this and who in the end solved the puzzling issue. It was all down to a missing IsAutoComplete() check. You guys were right @cala , @Tobschinski , it was a sneaky little core issue.

Furthermore correct a tiny bit in the quest markers (for classic only) to make it correctly display the completion marker as yellow and not blue. (it already correctly displays them on tbc and wotlk since questmarker logic is slightly more complex there with stuff not available to classic sadly)

Also contain some corrections to the chicken SD2 script. Lower chance, much lower timer, generally more blizzlike behaviour.

The [Fix chicken SD2 script for quest Cluck](/cmangos/mangos-classic/pull/373/commits/894dfe7d24be90ddc9bf57c726c82d53f2783f76) commit is the only one also applicable to tbc and wotlk (respecting the fact that on wotlk it should be available to horde as well), the rest of the commits are classic-only.

Solves:
https://github.com/cmangos/classic-db/issues/213
https://github.com/classicdb/database/issues/228

Improvements remaining:
- The spell that spawns the egg seems to be coming from the player and not the chicken making it seem like the player laid the egg..
- The chicken when it is 'activated' needs to stay in place and not move until it resets
- The 'hello emote' that the chicken does should not only be for alliance, it should also be readable to horde, even though they shouldn't be able to interact with it, because...
- In classic and tbc the chicken would become friendly only to alliance as per this comment (and many others from horde players saying they can't interact with it after they got the emote);
https://www.wowhead.com/quest=3861/cluck#comments:id=556749
>After maybe 40-50 repetitions, I got the message "Chicken looks up at you quizzically. Maybe you should inspect it?" (I'm actually copying this verbatim from my screen.)
>
>I got excited and turned towards the chicken to discover that it did NOT have the question mark over its head, nor did my cursor change to the gear when I hovered over it. The cursor remained a sword.
>
>Instead, the guard over at the town boundary RAN OVER AND KILLED THE CHICKEN!
>
>That has to have been the single funniest thing I've seen while playing this game. Apparently, activating the quest script got the chicken flagged as an Alliance chicken. Ha ha ha!

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
@@ -62,7 +62,7 @@ struct npc_chicken_cluckAI : public ScriptedAI
 
     void Reset() override
     {
-        m_uiResetFlagTimer = 120000;
+        m_uiResetFlagTimer = 20000;
 
         m_creature->setFaction(FACTION_CHICKEN);
         m_creature->RemoveFlag(UNIT_NPC_FLAGS, UNIT_NPC_FLAG_QUESTGIVER);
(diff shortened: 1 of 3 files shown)


killerwife:
Good job, but please make a version (or check if its needed) for each expansion for the quest markers. TBC added dailies and whatnot, and we need to be sure it plays ball with all the flags.

Phatcat:
The quest marker changes are only needed in classic, in tbc and wotlk it is already correctly shown with yellow markers, the logic is slightly different there since dailes/weeklies are a thing and is included in the logic there. Basically on tbc to have a blue question mark it checks that it's both a repeatable as well as a daily or weekly, which this quest is not, so it comes back with normal question mark. On classic we can't do such a check, so we need to check slightly differently.

Only commit that is needed on the other cores is the sd2 changes.

cala:
Wonderful! Just one question before I merge this: why remove the reset code in SD2 script?

Phatcat:
Because the chicken doesn't reset when you take the quest nor when you turn it in, can be seen in both the videos (:

cala:
Ok. You can call me lazy.

Phatcat:
Not trying to call you lazy at all, I just don't know what else to say than, I saw in the videos linked that they don't reset, so I made them not reset, not being snarky at all, just happy it's being merged =)

cala:
No worries, I should have added a winking smiley. 😉 

Would I have looked at the videos more thoroughly, I would not have need to ask. 😛

PR #374, opened 2019-02-11
Fix base\mangos.sql

Phatcat:
ERROR 1062 (23000) at line 8074: Duplicate entry '0-8-176' for key 'PRIMARY'

Doesn't require any sql update as it's only the base file which is affected, it is however preventing us from running the sql file.
It is only classic which has this error in the base file.


Levitanious:
So that's what was going on with that .sql, eh? You've been talking in private with the guy who asked for help in discord?

Phatcat:
... Yes I have..

PR #375, opened 2019-02-14
Partly revert 2670768

Phatcat:
Partly revert 2670768 which there is no gained functionality in

### DIFF
--- a/sql/base/characters.sql
+++ b/sql/base/characters.sql
@@ -21,7 +21,7 @@
 
 DROP TABLE IF EXISTS `character_db_version`;
 CREATE TABLE `character_db_version` (
-  `required_z2737_00_characters_cooldown` bit(1) DEFAULT NULL
+  `required_z2738_00_characters_pet_loyalty_fix` bit(1) DEFAULT NULL
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC COMMENT='Last applied sql update to DB';
 
 --
@@ -323,7 +323,6 @@ CREATE TABLE `character_pet` (
   `Reactstate` tinyint(1) unsigned NOT NULL DEFAULT '0',
   `loyaltypoints` int(11) NOT NULL DEFAULT '0',
   `loyalty` int(11) unsigned NOT NULL DEFAULT '0',
-  `xpForNextLoyalty` int(11) unsigned NOT NULL default '0',
   `trainpoint` int(11) NOT NULL DEFAULT '0',
   `name` varchar(100) DEFAULT 'Pet',
   `renamed` tinyint(1) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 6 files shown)


Phatcat:
@killerwife @devmaximus I would love your thoughts on the matter since you were involved with the merging of it.

killerwife:
Where is no gained functionality. We are saving loyalty experience to database.

Phatcat:
... Yes... we do that in field 7. It's called m_loyaltypoints and it was there before the commit which I am reverting from. What is being saved by the removed code is m_xpRequiredForNextLoyaltyLevel which is so very much redundant.

Tell me exactly what those functions I have removed provide?
m_xpRequiredForNextLoyaltyLevel? Why is that needed? We have the check for (m_loyaltyPoints > int32(GetMaxLoyaltyPoints(loyaltylevel))) in ModifyLoyalty which checks if we can level up. You don't have a field in in the character database for players with a field of 'xprequiredfornextlevel', so why is it needed for pets with loyalty?
Pet::SetRequiredXpForNextLoyaltyLevel() is completely needless when Pet::ModifyLoyalty already sets m_loyaltyPoints with what you modify it with...

So you tell me what exactly those functions provide of added functionality? The calculations in SetRequiredXpForNextLoyaltyLevel() ? Why aren't those integrated in GetMaxLoyaltyPoints(loyaltylevel)?
Actually, where do those calculations even come from?

You should be aware pet's can't level up loyalty currently and that your added command is broken in it's current state, not reliant on the other changes in the commit it was included with, and what a surprise, is fixed by these proposed changes?

killerwife:
I did review the code, and there are some minor differences, but nothing major. On that front you are right. What I would like you do first though is find where the current implementation breaks, so that I have a clue what to do with some outliers when merging this PR. Eventually I will make time for that too, but I assume you'd like this merged asap.

Phatcat:
> What I would like you do first though is find where the current implementation breaks

I swear I tried with several hunters with several beasts and could not get it to work yesterday but now it works on a frsh branch so you can disregard the last paragraph of my previous comment. The PR still stands however as I think whatever functionality was added with that commit besides the command should be treated as I previously mentioned.

Phatcat:
@killerwife If nothing else then [Levelrequirement of pet learning is in the taught spell](/cmangos/mangos-classic/pull/375/commits/f283a71436be17516dea60867c3c4667bcd3860b) is all ready to merge and independent from the other commit, I think it should be merged perhaps without the other commit for now so that the related issue can finally be closed.
Then you can afterwards ponder the fate of [Party revert](/cmangos/mangos-classic/pull/375/commits/a5033bdd61775e378fb0b26d868046289d67cf54) [2670768](https://github.com/cmangos/mangos-classic/commit/2670768f33f3203b5399c956ebb057d411675cd5) (which by the way was supposed to be Partly, not party*)

PR #377, opened 2019-02-16
pin/totp impl by Chaosvex

Phatcat:
- Rebased, refitted and squashed
- Added commands

Difference to the original PR: 
- Adds field uint8 'Security' and uses that for determining security type and not 'Locked'.
- Redefine type of and use already present token field and not security for storing the token pin.
- Uses, but initializes slightly differently, already present 'pinData' struct (sAuthLogonPinData_C)

Ref: https://github.com/cmangos/mangos-classic/pull/193

### DIFF
--- a/sql/base/realmd.sql
+++ b/sql/base/realmd.sql
@@ -58,7 +58,8 @@ CREATE TABLE `account` (
   `expansion` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `mutetime` bigint(40) unsigned NOT NULL DEFAULT '0',
   `locale` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `token` text,
+  `security` int(11) DEFAULT '0',
+  `token` varchar(16) DEFAULT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `idx_username` (`username`),
   KEY `idx_gmlevel` (`gmlevel`)
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -21,7 +21,7 @@
 */
 
 #include "Common.h"
-#include "Auth/HMACSHA1.h"
+#include "Auth/Hmac.h"
 #include "Auth/base32.h"
 #include "Database/DatabaseEnv.h"
 #include "Config/Config.h"
@@ -29,6 +29,7 @@
 #include "RealmList.h"
 #include "AuthSocket.h"
 #include "AuthCodes.h"
+#include "Util.h"
 
 #include <openssl/md5.h>
 #include <ctime>
@@ -79,11 +80,11 @@ typedef struct AUTH_LOGON_CHALLENGE_C
     uint8   I[1];
 } sAuthLogonChallenge_C;
 
-typedef struct AUTH_LOGON_PIN_DATA_C
+struct AUTH_LOGON_PIN_DATA_C
 {
     uint8 salt[16];
     uint8 hash[20];
-} sAuthLogonPinData_C;
+};
 
 typedef struct AUTH_LOGON_AUTHENTICATOR_DATA_C
 {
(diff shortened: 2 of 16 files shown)


sql/base/realmd.sql
```diff
@@ -58,7 +58,9 @@ CREATE TABLE `account` (
   `expansion` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `mutetime` bigint(40) unsigned NOT NULL DEFAULT '0',
   `locale` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `token` text,
+  `security` int(11) DEFAULT '0',
+  `pin` int(11) DEFAULT '0',
+  `token` varchar(16) DEFAULT NULL,
```
killerwife: Why this change right here? it was "text" before.

Phatcat: Because varchar takes up less space, it's indexable and has a variable max size (so we can set it to 16)
Text on the other hand...

https://stackoverflow.com/questions/25300821/difference-between-varchar-and-text-in-mysql

src/realmd/AuthSocket.cpp
```diff
@@ -182,7 +189,7 @@ std::array<uint8, 16> VersionChallenge = { { 0xBA, 0xA3, 0x1E, 0x99, 0xA0, 0x0B,
 
 /// Constructor - set the N and g values for SRP6
 AuthSocket::AuthSocket(boost::asio::io_service& service, std::function<void (Socket*)> closeHandler)
-    : Socket(service, std::move(closeHandler)), _status(STATUS_CHALLENGE), _build(0), _accountSecurityLevel(SEC_PLAYER)
+    : Socket(service, std::move(closeHandler)), _status(STATUS_CHALLENGE), promptPin(false), _build(0), _accountSecurityLevel(SEC_PLAYER)
```
evil-at-wow: In the header the member promptPin is declared before _status, so this will result in compiler warnings about members being initialized in an order that doesn't match the declared order (that's also what actually happens, even though the initializer list order suggests otherwise). 

Please re-order them here, or even better, use C++11 and remove promptPin in the initializer list here and declare the member as ```bool promptPin { false };``` for example.

src/realmd/AuthSocket.cpp
```diff
@@ -636,8 +650,17 @@ bool AuthSocket::_HandleLogonProof()
     BigNumber M;
     M.SetBinary(sha.GetDigest(), 20);
 
+    ///- Check PIN data is correct
+    bool pinResult = true;
+
+    if(securityFlags & SECURITY_FLAG_PIN && !lp.securityFlags & SECURITY_FLAG_PIN)
```
evil-at-wow: This looks quite suspicious. Remember that in C++ the logical not (operator!) has priority over the bitwise and (operator&), so you're effectively checking first if ```lp.securityFlags``` is non zero and then you're doing a bitwise and of that bool with SECURITY_FLAG_PIN. I think this needs to be ```!(lp.securityFlags & SECURITY_FLAG_PIN)``` or equivalent like ```~lp.securityFlags & SECURITY_FLAG_PIN```, but the first is probably easier to understand.

Phatcat: The code is refactored in a later commit and looks like this now; https://github.com/cmangos/mangos-classic/pull/377/files#diff-d7bd4c0307846a89da8bec996236180fR648 - It is no longer using a logical not operator on the flags.

src/realmd/AuthSocket.cpp
Phatcat: I got no compiler warnings on my end, I guess the VS c++ compiler doesn't care much.

So solution is just to initialize promptPin before _status? or are there other members being initialized in wrong order compared to header declaration order?

evil-at-wow: Indeed, the VS compiler doesn't warn about that by default. In fact, I only recently found out it can do so after all, but only in pretty recent versions according to [Microsoft's documentation](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5038?view=vs-2017) ("This warning is new in Visual Studio 2017 version 15.3, and is off by default."). But GCC and clang do warn about it for ages, to avoid subtle bugs like one member being initialized with a value based on another uninitialized member, although the order in the code seems to suggest something else. In the case here that's not a problem of course and the code is just fine, but the warnings get annoying.

And yes, just swap the order and initialize promptPin before _status, so it matches the order members are declared in the header, and all is perfect.

Phatcat: I did it the other way around, switched places of them in the header instead, thought that made more sense. I am also gonna squash his 2 commits as they touch on the same code making reviewing them a bit messy.

Phatcat:
I just force-pushed some changes @cyberium , just informing in case you just pulled in and are working with an now outdated branch.

I found there was already a struct for pinData, also squashed his 2 commits into one, to not change the same parts of the code twice in the same PR, which def. didn't make reviewing easy either. Should be somewhat clearer now.

I also added some data validation checks on the commands so people don't enter invalid stuff

Phatcat:
.. Just had to do another force push as I noticed a mistake in my commands commit. Everything should be all good now, I'll leave it be for now until I hear otherwise.

Phatcat:
Just curious, but did you self assign to this so that you could lock it out of getting merged, Cyberium?

Considering you were the one to state on the original PR that it was a 'nice proof of concept' but that you doubted 'anyone would use it', I don't think you are the right person to be overseeing the merging of this, because... you're perhaps the most biased person against this (besides killerwife and the rest of VoA, since their half-arsed lackluster version got merged despite all of this already being an open PR on cmangos for what, a year prior to that?)...

PR #379, opened 2019-02-20
SQL 8 (MySQL 8.0) compliance

Phatcat:
rank is now a keyword. groups is also a keyword.

This has full backwards compatibility with older versions of SQL

I had already made this for classic so I'm gonna PR it as well 'cause why not.


cala:
Merged, thank you!

PR #380, opened 2019-02-22
Correct CLUCK emotes

Phatcat:
https://www.youtube.com/watch?v=Pn2qPJWBCdw (2:25)
We see a horde here get the same emote as an alliance, and also we see that the emote being done on /cheer is the unused one in the scriptdev sql while the one that is currently used on /cheer there seems to be no evidence of anywhere.
https://www.youtube.com/watch?v=az8BNDNqzc4 (3:20)


cala:
Merged, thank you! :+1:

PR #383, opened 2019-03-21
add a null ptr check when cast a SummonWild spell

RichardGe2:
for example if you execute :  
.cast 7975
you are making the server crash if you don't have this fix.


cala:
Good catch, thank you!

PR #384, opened 2019-04-11
Fix minor code issues

bdebaere:
Remove duplicate check.
Fix dereferencing of null pointer.
Switch checks to allow null check to come first.
Clean up args to prevent memory leak.
Fix potential overflow.
Simplify check.

### DIFF
--- a/src/game/Entities/CharacterHandler.cpp
+++ b/src/game/Entities/CharacterHandler.cpp
@@ -457,18 +457,18 @@ void WorldSession::HandlePlayerLoginOpcode(WorldPacket& recv_data)
     {
         // player is reconnecting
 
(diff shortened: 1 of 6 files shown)


cala:
Merged, thank you. I removed one of the check simplification you made (session socket) because this one is required for PlayerBot IIRC.

bdebaere:
@cala I assume you're talking about the one below? Care to elaborate?

```cpp
if (!m_Socket || (m_Socket && m_Socket->IsClosed()))
if (!m_Socket || m_Socket->IsClosed())
```
Unless I am gravely mistaken, you cannot get into the second part of the or i.e. `|| m_Socket->IsClosed()` without `m_Socket` being not null.

PR #386, opened 2019-04-28
[PlayerBot] Fix bots going into idle state after looting

ainslievaux:
9c19a96a seems to have introduced a regression that caused bots to stop
following the master after looting.

This commit adds a check after bots are finished looting to see if they are 
ordered to follow, and resets their motion if so.


ainslievaux:
It seems that this regression also breaks bots' buffing, because we never reach the code that handles that if bots are set to follow. I'll have a fix for it soon.

cala:
I'm quite busy at the moment and did not had the time to look at your PR. Is the issue resolved ?

cala:
The issue was more global than only looting and should be fixed by: https://github.com/cmangos/mangos-classic/commit/2967b54998ee42d8b4db3b85f2073dae3e834ded

PR #387, opened 2019-05-10
[Playerbot] Fix crash bug when training skills

ainslievaux:
Fix a segfault when we attempt to learn a spell from a trainer that doesn't teach that spell.

To reproduce:
- Create two bots of different classes 
- Target one of the class trainers (the ones below are the troll mage & priest trainers in Durotar) and instruct the bots to learn a class-specific spell that that trainer teaches, e.g. 
-- `/tar mai'ah`
-- `/p skill learn [Arcane Intellect]`
- Target the other class trainer and instruct the bots to learn the same spell, e.g. 
-- `/tar ken'jai`
-- `/p skill learn [Arcane Intellect]`


cala:
Good catch, thank you!

PR #388, opened 2019-05-24
Update VS labels to be consistent to the modules name

BenDol:
Not sure if this is something the dev team will accept, but figured I'd get your thoughts on this considering it is a minor change.

**Before:**
devenv_2019-05-25_04-14-00
**After:**
devenv_2019-05-25_04-05-30 

Let me know if there is a naming standard and I can update it? But this seems better.

### DIFF
--- a/src/game/CMakeLists.txt
+++ b/src/game/CMakeLists.txt
@@ -109,7 +109,7 @@ if(PCH)
     #set_property(SOURCE ${PrecompiledSource} PROPERTY COMPILE_FLAGS "/Yc\"${PrecompiledHeader}\" /Fp\"${PrecompiledBinary}\"")
     set_target_properties(${LIBRARY_NAME} PROPERTIES COMPILE_FLAGS " /Yu\"${PrecompiledHeader}\" /FI\"${PrecompiledHeader}\" /Fp\"${PrecompiledBinary}\" /bigobj")
     set_source_files_properties(${PrecompiledSource} PROPERTIES COMPILE_FLAGS "/Yc\"${PrecompiledHeader}\" /Fp\"${PrecompiledBinary}\"")
-    set_target_properties(${LIBRARY_NAME} PROPERTIES PROJECT_LABEL "Game")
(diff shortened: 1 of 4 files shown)


Levitanious:
Anticheat isn't in the commit, but on the screenshot `:3`

BenDol:
Yeah it's from my `dev` branch.

killerwife:
Since @cyberium changed it initially in the cmake rework, I will tag him to see his opinion.

cyberium:
I would prefer to rename executable. It was just a start to make exe name more readable.
realmd -> AuthServer
mangosd -> WorldServer
And probably some more services in the future.
What's your thought?

BenDol:
I personally prefer `realmd` & `mangosd` for the sake of consistency since my OCD brain gets triggered by modules with caps naming mixed in with modules without it, _schizo little modules_. Probably not referring to me here though haha but that's my 2 cents anyway :)

evil-at-wow:
@BenDol We could still rename it to ```authserver``` and ```worldserver``` in that case :smile:

The ```<something>d``` is rather typical for the UNIX world (at least for daemons), so that's fitting, but there's more than Linux/BSDs in this project. I don't mind either way, but whatever the name is, it's probably good to make it consistent in all places.

Laizerox:
+1 authserver and worldserver.

Reason? Imagine if we get someday to implement clustering, worldserver will be just and instance that is connecting to realmserver or masterserver that accepts and serves character lists and redirects to worldserver.

cala:
Poking you guys on this PR. This is VS related and I don't use VS so I'm not the best one to review it. Thanks.

cyberium:
We will still put that aside but please give your opinion about it.

To summarize we are heading to rename executable to worldserver and authserver.
We may also need later to rename db accordingly.

killerwife:
Closing this for now, as the current stance is fine for compatibility vs TC and vmangos and whatnot, and we need to be a bit friendly with the cross project naming for sake of sanity, even if others arent.

PR #393, opened 2019-06-13
Don't downrank spells in classic

Phatcat:
## 🍰 Pullrequest
The whole auto downranking of buff/healing spells is a tbc+ thing

### Proof
http://wowwiki.wikia.com/wiki/Patch_2.0.1
"When casting a buff on a low level target, the appropriate rank of the buff will automatically be applied."

### Issues
- relates to:
https://github.com/cmangos/issues/issues/950
https://github.com/cmangos/issues/issues/1860

### How2Test

- None


Warlockbugs:
This looks like a convenient QoL functionality for local testing to just throw it away. Maybe adding a config value with appropriate default setting for corresponding expansion would be more desirable so user can bring it back on demand?

Phatcat:
I dunno, perhaps, one thing is for sure, what is being done currently is not blizzlike, furthermore the 'upranking' could be achieved with macros back in 1.12

Problem is that it's not working at all atm,it doesn't uprank the spellcast, it locks the caster from using any spells, it does as described in one of the issues which Jimmy can't help himself from closing even though it's a still present issue (and will continue to be until someone works on it)

Phatcat:
So since it not only doesn't work, it's not blizzlike, my first instinct was to simply get everything of it out until the core worked and was blizzlike, since 'blizzlike > QoL' all day every day

Defending something which is not blizzlike, can be fine
Defending something which is not blizzlike which doesn't work and breaks the core.... hmmm... I don't really have much patience left over for that.

But you can approach this issue however you see fit, I am not gonna waste any more time on it.

cala:
I won't comment the code by itself, but I agree with the validity of the change and I confirm than auto-rank selection for spell was not a thing in Classic. If for some reasons a player wanted to cast a low level spell, it has to be done manually. Healers were often using this to reduce overhealing.

Phatcat:
@Warlockbugs Now it has the config option that you asked for and I tested, it works with both settings, and actually works now, so it automatically downranks a buff on a low level char with the config on and with the config off (default) it will display the 'target is too low level' error message and not hang like on current master.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/bbea27363200bf650671dd535be55164ad985b1a and https://github.com/cmangos/mangos-classic/commit/7410acf0755c70ee12eb8d9daf1182754266b90e

PR #397, opened 2019-06-30
drop sha_pass_hash support

esno:
## 🍰 Pullrequest
due to the fact that sha1 is broken it should not be used to store passwords as sha1 hashed string.
cmangos team  decided to drop that support.


src/realmd/AuthSocket.cpp
```diff
@@ -184,8 +185,7 @@ std::array<uint8, 16> VersionChallenge = { { 0xBA, 0xA3, 0x1E, 0x99, 0xA0, 0x0B,
 AuthSocket::AuthSocket(boost::asio::io_service& service, std::function<void (Socket*)> closeHandler)
     : Socket(service, std::move(closeHandler)), _status(STATUS_CHALLENGE), _build(0), _accountSecurityLevel(SEC_PLAYER)
 {
-    N.SetHexStr("894B645E89E1535BBDAD5B8B290650530801B18EBFBF5E8FAB3C82872A3E9BB7");
-    g.SetDword(7);
+    srp = new SRP6();
```
evil-at-wow: Who deletes this? I can't find a matching delete, so it looks like this introduced a memory lead.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -37,6 +38,9 @@ AccountMgr::~AccountMgr()
 
 AccountOpResult AccountMgr::CreateAccount(std::string username, std::string password) const
 {
+    SRP6* srp = new SRP6();
```
evil-at-wow: I see several of these local variables without matching deletes in this commit, so I think you've introduced several memory leaks. Since this is a local variable, just create the variable on the stack instead, it's not such a big object anyway (```SRP6 srp;```). See also remarks below.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -37,6 +38,9 @@ AccountMgr::~AccountMgr()
...
+    std::string sha_pass_hash;
```
evil-at-wow: You don't need to declare all variables at the top of a function. It's not good old C, so please move this to where it's used for the first time.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -48,7 +52,9 @@ AccountOpResult AccountMgr::CreateAccount(std::string username, std::string pass
         return AOR_NAME_ALREADY_EXIST;                       // username does already exist
     }
 
-    if (!LoginDatabase.PExecute("INSERT INTO account(username,sha_pass_hash,joindate) VALUES('%s','%s',NOW())", username.c_str(), CalculateShaPassHash(username, password).c_str()))
+    sha_pass_hash = CalculateShaPassHash(username, password);
```
evil-at-wow: See earlier remark: ```std::string sha_pass_has = CalculateShaPassHash(username, password);```.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -100,6 +106,9 @@ AccountOpResult AccountMgr::DeleteAccount(uint32 accid) const
 
 AccountOpResult AccountMgr::ChangeUsername(uint32 accid, std::string new_uname, std::string new_passwd) const
 {
+    SRP6* srp = new SRP6();
+    std::string sha_pass_hash;
+
```
evil-at-wow: See earlier remarks, here as well for both lines.

src/game/Accounts/AccountMgr.cpp
```diff
@@ -113,20 +122,24 @@ AccountOpResult AccountMgr::ChangeUsername(uint32 accid, std::string new_uname,
...
         return AOR_DB_INTERNAL_ERROR;                       // unexpected error
 
     return AOR_OK;
 }
 
 AccountOpResult AccountMgr::ChangePassword(uint32 accid, std::string new_passwd) const
 {
+    SRP6* srp = new SRP6();
     std::string username;
+    std::string sha_pass_hash;
```
evil-at-wow: See earlier remarks, here as well for both new lines.

src/shared/SRP6/SRP6.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "Common.h"
+#include "Auth/HMACSHA1.h"
+#include "Auth/base32.h"
+#include "SRP6.h"
+#include "Log.h"
+
+#include <openssl/md5.h>
+#include <ctime>
+#include <utility>
+
```
evil-at-wow: This list of includes is probably a bit overkill. What do we need Log.h for, or md5.h or ctime?

src/shared/SRP6/SRP6.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "Common.h"
+#include "Auth/BigNumber.h"
+#include "Auth/Sha1.h"
+#include "ByteBuffer.h"
+
+#include <boost/asio.hpp>
+
+#include <functional>
+
```
evil-at-wow: It looks like a few of these headers are not needed. At least I don't see why we need asio.hpp, ByteBuffer.h and functional here in this header. You're probably missing <string> though (I'm sure it's indirectly included via another header, but perhaps we should not rely on that).

esno:
thx @evil-at-wow for your review. this is still a prototype where I have to work on functionallity. rebase and code cleanup will follow soon. I'll keep an eye on your recommendations

evil-at-wow:
No problem, and thank you for the effort. I just wanted to give some (early) feedback. 
If only to make sure cyberium wouldn't just hit the merge button :grin:

src/realmd/AuthSocket.cpp
esno: will be deleted in `~AuthSocket()`

src/game/Accounts/AccountMgr.cpp
esno: old habbits. moved the declarations. thx

src/game/Accounts/AccountMgr.cpp
esno: since I rebased the code `sha_pass_hash` variable is obsolete and dropped

src/game/Accounts/AccountMgr.cpp
esno: I'm not sure how `AccountMgr` is instanciated. The reason why I didn't create that variable on the stack was to ensure that sensitive data from SRP computation will not be shared/mixed with other actions. Meanwhile I added the delete statements for `srp`. Are you fine with that?

evil-at-wow: The AccountMgr is a MaNGOS::Singleton, so it's created when the instance is first accessed I think.

As long as we're not leaking memory, I don't mind having it on the heap in cases like this. It probably doesn't add a lot of security (the pointer is on the stack instead, so the data is just one hop away) but it obviously also doesn't make it worse.

Remember that instead of adding delete statements, you can also use std::unique_ptr for example to do that for you (even in the face of exceptions). But I might be barking up a tree for no reason here because I didn't check your updated code yet. I'll try to do that later this week.

esno: no man. it's my lack of up2date cpp knowledge. I'll push an updated version after testing

metacreo:
"due to the fact that sha1 is broken" it a mistake - this is fact. The whole panic is based on a huge number of known hashes. Add a secret variable to user password and the whole problem disappears.
Like $paasword = SHA1(CONCAT('my_super_user'),':',UPPER('my_super_pass'),'<',('my_secret_var'))

esno:
imho the whole panic about sha1 is the fact that it is broken not in a way of known hashes but to the fact that there are theoretical as well as practical attacks to build a collision which means that different input values will be computed to the same hash

metacreo:
where to read example? thanks

esno:
is that proof enough? https://shattered.io

metacreo:
Yep thanks I see... 
I think it not 100% proof. They have one example with signed documents and unfortunately they only theoretically describe the algorithm. I think you should not hurry. Probably identical hashes received with certain conditions. In addition, except from Google, there is no more information. Perhaps this is a commercial move.

esno:
I'm not in hurry ;) and I don't think it's a commercial move it's more like the researchers are proud of their findings. in professional IT sha1 will not used anymore for crypto stuff - I hope so. At least when it comes to cmangos there is no need to have the hashes stored in database.

ulfgebhardt:
Do you think that is a real threat?

image

It takes 110 GPU's 1 year to find a collision. That does not mean that it will calculate the correct hash.

Therefore an attacker who has access to the database can find a collision with one hash within 1 year if he finds himself in the possession and with the money to run 110 GPUs. That does not mean he will find the correct Cleartext?!

Also: adding a secret string which is hashed with the username and pw will not add much of security since the attacker needs access to the database to obtain the hash - therefore its quite likely he has also access to the rest of the system which will include the secret string.

Just some uneducated thoughts on the matter <3

metacreo:
How do you plan to keep passwords in the database? I see a simple hex. What is it like?
and... "access to the database", server... etc... direct to wallet :) this is exactly panic, sorry

esno:
@ulfgebhardt my point is not about current computation times it's about shattered is just one practical proof of a theory which can be speed up by some enhancements. At least for me it is also not a question about having access to the database because a common practice (for me) is that also admins shouldn't keep secret data that aren't necessary for technical reasons.

@metacreo `this is exactly panic` isn't the way I like to discuss. Nobody in here has any kind of panic about that otherwise I would work on that topic as long as it's finished - Just to point that out sha1 is broken for a while now and NIST already planned to phase it out till 2010. Is it really kind of panic when someone starts working on a topic a decade after a technique is deprecated?

My point and the reason for this PR is that there is **no** technical reason to store the password hash in database because the auth mechanism for the client works without it. Also for CMS's it's not quiet necessary because all the stuff can be implemented in a proper way.

Last but not least to answer your question about the hex values stored in DB. They represent results from SRP computation which can be used to store and verify credentials. Please have a look to the SRP class to see how it works (https://github.com/blizzlike-org/cmangos-vanilla/blob/vanilla/bugfix/sha_pass_hash/src/shared/SRP6/SRP6.h).
If the attached comments doesn't fit to you feel free to add a comment and I am willing to give my best to add some additional lines of documentation

Laizerox:
> How do you plan to keep passwords in the database? I see a simple hex. What is it like?
> and... "access to the database", server... etc... direct to wallet :) this is exactly panic, sorry
> secret string add top security for against known hashes

Adding another secret word doesn't work, why? Because client doesn't do that. So even if you could that on your website it would be literally impossible for client to do that. This sha string is used in SRP6 protocol to calculate session keys and everything around it. If session keys or proof of session key doesn't match then your client will not be able to authenticate against server.

However fear not I've been working on PHP compliant client & server for "legacy" backwards compatibility and once JS counterparts of client & server are done you should be able to implement proper SRP6 authentication using rest api or something similar.

metacreo:
Yes, thanks, I understood everything. I meant only password encryption in the database. I don’t see any problem in implementing PHP connection with cmangos. As I understand there is already a SOAP service? If so, then it is very good. Only I did not try to use it.

t880216t:
@esno  @metacreo  Sir ,  how to create a user account with SQL? 
old sql 
``` php 
$entry = "INSERT INTO account (username, sha_pass_hash, gmlevel, email, last_login) VALUES (UPPER('" . $mysqluser . "'), SHA1(CONCAT(UPPER('" . $mysqluser . "'),':',UPPER('" . $pw . "'))),0, '" . $email . "', '2019-01-21')";
```
now exec error log
``` bash 
Unknown column 'sha_pass_hash' in 'field list'
```

trumpetx:
@t880216t - I spent yesterday trying to figure this out.  Here's the answer using no external libraries (raw PHP).  Note, this is the BARE MINIMUM that you need to insert S/V values, it will not perform an hash authentication via SRP challenge/response.  For that, I suggest you use https://github.com/Laizerox/php-wowemu-auth

```
#!/usr/bin/php
<?
function bchexdec($hex) {
    if(strlen($hex) == 1) {
        return hexdec($hex);
    } else {
        $remain = substr($hex, 0, -1);
        $last = substr($hex, -1);
        return bcadd(bcmul(16, bchexdec($remain)), hexdec($last));
    }
}

function bcdechex($dec) {
    $last = bcmod($dec, 16);
    $remain = bcdiv(bcsub($dec, $last), 16);

    if($remain == 0) {
        return dechex($last);
    } else {
        return bcdechex($remain).dechex($last);
    }
}

function reverseHex($string)
{
    for ($i = 0, $length = strlen($string); $i < $length; $i += 2) {
        $bytes[] = substr($string, $i, 2);
    }
    return implode(array_reverse($bytes ?? []));
}

$username="trumpetx";
$password="trumpetx";
$sha = hash('sha1', strtoupper($username . ":" . $password), true);

$n = bchexdec("894B645E89E1535BBDAD5B8B290650530801B18EBFBF5E8FAB3C82872A3E9BB7");
$generator = 7;

$s = bin2hex(random_bytes(32));
$salt = hex2bin(reverseHex($s));
$x = bchexdec(reverseHex(hash('sha1', $salt . $sha)));
$v = bcdechex(bcpowmod($generator, $x, $n));


echo("S: " . $s . PHP_EOL);
echo("V: " . $v . PHP_EOL);
?>
```

Aveyder:
Same on Python:
```
import os
import hashlib

g = 7
N = int("894B645E89E1535BBDAD5B8B290650530801B18EBFBF5E8FAB3C82872A3E9BB7", 16)


def register(login, password):
    salt_hex = os.urandom(32).hex()

    return salt_hex, get_verifier(login, password, salt_hex)


def auth(login, password, salt_db, verifier_db):
    verifier_calc = get_verifier(login, password, salt_db)

    return verifier_calc == verifier_db


def get_verifier(login, password, salt_hex):
    salt_bytes = reverse_bytes(bytes.fromhex(salt_hex))

    h1_bytes = hashlib.sha1(
        b"%s:%s" % (login.encode().upper(), password.encode().upper())
    ).digest()
    h2_bytes = hashlib.sha1(salt_bytes + h1_bytes).digest()

    h2_int = int.from_bytes(h2_bytes, byteorder="little")

    verifier_int = pow(g, h2_int, N)

    return format(verifier_int, 'x').upper()


def reverse_bytes(b):
    b_bytearray = bytearray(b)
    b_bytearray.reverse()
    return bytes(b_bytearray)


if __name__ == "__main__":
    login = "my_user"
    password = "my_pass"

    s, v = register(login, password)

    assert auth(login, password, s, v) is True
    assert auth(login, "another_pass", s, v) is False

```

PR #399, opened 2019-07-01
add optional output path to extractor scripts

esno:
## 🍰 Pullrequest
this PR introduces an optional output path used by the extractor scripts.
It should prevent the requirement of writing into the client directory which could be in fact readonly.

### Issues
- https://github.com/cmangos/issues/issues/1965

### How2Test
```
./ExtractResources.sh a <clientdir> <outputdir>
```


killerwife:
Resolved

PR #400, opened 2019-07-02
Move VisibilityDistanceType in CreatureInfo match with CreatureTemplate. 

darklordsqk:
## 🍰 Pullrequest
Exchange the sequence of two members in CreatureInfo, VisibilityDistanceType and EquipmentTemplateId, to avoid data filling errors. 

### Proof
Login and find a mob with weapon on its hand. 

### Issues
In the Creature defines, member VisibilityDistanceType is behind of EquipmentTemplateId. 
While in DB table creature_template, VisibilityDistanceType is in front of EquipmentTemplateId. 
This will lead to a misplaced columns data filling on the Init of sStorageCreature. 
The directly appears issue is that creatures with EquipmenId!=0 have no equepment on their hands. 

### How2Test
- None

### Todo / Checklist
- [X] None


darklordsqk:
No.... Unhandle the origin commits.

PR #405, opened 2019-07-21
Sartura

Tobschinski:
AQ40 - Battleguard Sartura

add missing spell Sundering Cleave
update incorrect spellids of Enrage and Berserk
add spelleffect of Whirlwind to mimic fixation
no longer use ScriptedAI for Royal Guards

-------------------------------------------------------
introduce modifyAllThreatPercent

to avoid iterating threatlists on multiple occasions

### DIFF
--- a/src/game/Combat/ThreatManager.cpp
+++ b/src/game/Combat/ThreatManager.cpp
@@ -240,6 +240,7 @@ void ThreatContainer::clearReferences()
 
 //============================================================
 // Return the HostileReference of nullptr, if not found
+
 HostileReference* ThreatContainer::getReferenceByTarget(Unit* victim)
 {
     if (!victim)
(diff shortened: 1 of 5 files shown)


src/game/Combat/ThreatManager.cpp
```diff
@@ -281,6 +281,20 @@ void ThreatContainer::modifyThreatPercent(Unit* victim, int32 threatPercent)
...
+    ThreatList const& tList = getThreatList();
+    UnitList uList;
+
+    // store attacker
+    for (auto itr : tList)
+        uList.push_back(itr->getTarget());
+
+    // process the threat modification
+    for (auto itr : uList)
+        modifyThreatPercent(itr, threatPercent);
```
evil-at-wow: Allow me to suggest even less iterating by not building a unit list first. You're now basically iterating over the ThreatList once to build a unit list of equal size, then you iterate over that unit list. Unless I've missed something, this seems even simpler:

```
for (auto itr: tList)
    modifyThreatPercent(itr->getTarget(), threatPercent);
```
If duplicates would be possible and should be avoided, your solution is better as long as you use a set instead of a list.

killerwife:
I have even better solution to avoid altering ThreatManager. Will ping tomorrow.

src/game/Combat/ThreatManager.cpp
cyberium: @evil-at-wow it sound tempting but no, the modifyThreatPercent method can modify the tList storage resulting to a crash.
Yeah not so evident at first look. Its an old code that probably need some upgrade to simplify it and make it more clear.
He choose the safest way to do it, maybe there is a more optimized way to do it tho.

cala:
Thanks @killerwife I'll wait for your update before merging this PR, then.

killerwife:
@cala here it is:
https://github.com/cmangos/mangos-tbc/commit/08da03d269705975e8be28cac50550ceabff3905

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/72560324218b54e4ac454cc4c45c52a8096c93eb
As per @cyberium advice, I kept the original threat approach.

PR #406, opened 2019-11-22
Don't set double health/power for guardians

canidae:
## 🍰 Pullrequest
With commit 694b7391b85a4d7df42291960fb19efdcfa991ab health/power ended
up being set twice the intended value.
Max health/power was already set by the CLS function SelectLevel(), and
UpdateAllStats() caused these values to be doubled.

### Proof
- None

### Issues
- Fixes https://github.com/cmangos/issues/issues/2064

### How2Test
- Find any NPC with guardian (Except Imp Minion guardians, they do not use CLS, https://github.com/cmangos/classic-db/pull/239 fixes this), check health before and after this change, health should be halved and set to a sensible amount.

### Checklist
- UpdateAllStats() is not invoked for guardians, but I believe this is entirely covered by CLS.
- ~POWER_RAGE is not forced to 0 for guardians (unless CLS handles this as well)~. It seems like CLS do handle this, see https://github.com/cmangos/mangos-classic/blob/master/src/game/Entities/Creature.cpp#L1313-L1339

### DIFF
--- a/src/game/Entities/Pet.cpp
+++ b/src/game/Entities/Pet.cpp
@@ -1377,21 +1377,21 @@ void Pet::InitStatsForLevel(uint32 petlevel)
         SetCreateMana(mana);
         // Set base Armor
         SetModifierValue(UNIT_MOD_ARMOR, BASE_VALUE, armor);
+
+        // Need to update stats - calculates max health/mana etc
+        UpdateAllStats();
+
+        // Need to set Health to full
+        SetHealth(GetMaxHealth());
+
+        // Need to set Mana to full
+        if (GetPowerType() == POWER_MANA)
+            SetPower(POWER_MANA, GetMaxPower(POWER_MANA));
+
+        // Remove rage bar from pets (By setting rage = 0, and ensuring it stays that way by setting max rage = 0 as well)
+        SetMaxPower(POWER_RAGE, 0);
+        SetPower(POWER_RAGE, 0);
     }
-
-    // Need to update stats - calculates max health/mana etc
-    UpdateAllStats();
-
-    // Need to set Health to full
-    SetHealth(GetMaxHealth());
-
-    // Need to set Mana to full
-    if (GetPowerType() == POWER_MANA)
-        SetPower(POWER_MANA, GetMaxPower(POWER_MANA));
-
-    // Remove rage bar from pets (By setting rage = 0, and ensuring it stays that way by setting max rage = 0 as well)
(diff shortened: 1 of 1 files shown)


canidae:
The failing check does not appear to be related to my changes (last commit to master fails with same error)

cala:
Thank you for this PR.

It is a long time since the last time I looked at the stats or pet code. So I'll need to dig a bit before being able to give an opinion on the matter. It is possible that other devs that worked more recently on that part of the code can give hints also.

canidae:
For debugging I printed out `GetMaxHealth()` before and after `UpdateAllStats()`, which showed that the health of guardians was doubled.
I'm running a server with this fix and guardians have more sensible health now. I've not discovered any side effects yet, and when I looked over the code it _seemed_ like a fair fix, but as you know I'm not experienced with the code yet, so it's a good idea to have someone more experienced look over it.

cala:
After looking at the code and your suggested update, I think this PR is correct. I just need to test the change before merging. :+1:

cala:
After investigating further, there seem to be two reasons for this issue.

The first one is indeed the stats of guardian pets that are calculated twice (increasingly). I should merge the related PR soon.

The second one needs to be investigated. It seems that we are wrongly setting the level of guardian pets when their summoner is a NPC.

Could someone check on retail the level of the guardians summoned by the following NPCs/creatures:
- Silithid Swarmer (summons Silithid Swarm in combat), found in the Barrens, south of entrance to Mulgore and north of entrance to Dustwallow Marsh
- Centipaar Swarmer (summons Silithid Swarm in combat), found in Tanaris in the hive right north of Un'Goro entrance
- Dark Strand Cultist (summons Imp Minion), found in northern Ashenvale near the entrance to Darkshore
- Bloodsail Warlock (summons Imp Minion), found in Stranglethorn Vale, on the shore south of Booty Bay

These are two combinaisons of the same NPC being summoned as guardian by two different NPCs (of different level). I chose these because Silithid Swarm and Imp Minion are summoned by spells with different attributes, so this can be helpful.

Thanks to anyone that can look at this.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/10b092f7dcf92df44b5e192a3c8d1a25bdf5b1bc
Thank you.

Yatzii93:
Is this working correctly, guaridan pets seem to be doing like 1-2 damage now. Seems to be missing their attack power.

killerwife:
I have a fix inc in export. Keep u posted.

PR #408, opened 2019-12-12
Teach Sapphiron how to fly. (Naxx)

Yatzii93:
## 🍰 Pullrequest

Sapphiron needs additional code to do his proper fly animation and not be ground locked.
I figured @cala might appreciate this considering he's reworking naxx and it took me a little while to perfect this for Felmyst in Sunwell. 
### Proof

- None

### Issues

- Give Sapphiron wings.

### How2Test

- Do the boss fight until Sapphiron does his airbolt phase.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/naxxramas/boss_sapphiron.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/naxxramas/boss_sapphiron.cpp
@@ -108,6 +108,10 @@ struct boss_sapphironAI : public ScriptedAI
         m_uiIceboltCount = 0;
 
         SetCombatMovement(true);
+        m_creature->RemoveByteFlag(UNIT_FIELD_BYTES_1, 3, UNIT_BYTE1_FLAG_FLY_ANIM);
+        m_creature->SetByteFlag(UNIT_FIELD_BYTES_1, 3, UNIT_BYTE1_FLAG_ALWAYS_STAND);
+        m_creature->SetCanFly(false);
+        m_creature->SetHover(false);
         m_creature->SetLevitate(false);
     }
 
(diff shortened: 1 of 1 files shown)


cala:
Great, thank you. :+1: 
This will indeed save me some times as this was one of the issues I've already identified with this fight.
This will also benefit to Onyxia encounter phase 2 where the boss is standing in the air instead of flying.

cala:
@Yatzii93 I assume this is a direct port from TBC code because `UNIT_BYTE1_FLAG_FLY_ANIM` does not exist in Classic.

I will look at this.

Yatzii93:
> @Yatzii93 I assume this is a direct port from TBC code because `UNIT_BYTE1_FLAG_FLY_ANIM` does not exist in Classic.

@cala yep! Apologies if it's not as helpful as I thought, I only work on the TBC core.

cala:
No worries, just to be sure. :wink:

killerwife:
I am gonna do a small possibly useless note. Nightbane and Onyxia are utterly different. That being said, Nightbane was in development when Sapphiron was in development. ^^

cala:
I tested this PR on Classic and TBC. On TBC, it works fine: Sapphiron takes off, hover and nicely play his flying animation when casting spells and when being idle in the air and finally, he lands correctly.

However, on Classic, this does not work. Take off/landing are fine. Flying animation is played when casting spells, but when Sapphiron should be idle in the air he stands still as if on the ground instead of playing his flying/hovering animation. Basically, this looks like the same issue than with Onyxia encounter in phase 2.

It looks that we are missing the Classic equivalent of `UNIT_BYTE1_FLAG_FLY_ANIM` and we should investigate on this.

killerwife:
As someone who raged at onyxia very recently, you will find that making these dragons work is extremely difficult. I hope to do another classic export round, and I will explain what I am on about. The code above did not work for onyxia even in TBC, even though the identical code works perfectly for nightbane and felmyst. I'll link you cala on the commit when I port it.

cala:
I'm not surprised at all that having nice flying dragons in Classic is extremely difficult. Flying was not a real thing in Classic, so I guess Blizzard used workarounds for the handful of NPCs that should emulate fly rather that writing a full system for them...

I've merged this commit for code consistency with TBC with a few changes : I commented out the non-existing flag and I disabled flying/levitating/hovering when Sapphiron is on ground because he was hover towards his target instead of walking/running.

I'll be happy to update this part of the code once we have a working code for flying creatures in Classic, but in the meantime, I prefer to have things moving forward. :)

PR #409, opened 2019-12-27
Load world safe locs orientation values.

ratkosrb:
The new classic client includes orientation values in WorldSafeLocs.db2. This pull requests adds support for loading these values from the database.

Blizzard defined the orientation for all graveyards back in patch 1.8 but they didn't export the orientation value to the client dbc until a later expansion.

> World of Warcraft Client Patch 1.8.0 (2005-10-11)
> - All graveyards that needed adjustment were changed so that a 
>   character's spirit comes into the world facing toward the Spirit Healer.

Database update https://github.com/cmangos/classic-db/pull/246

### DIFF
--- a/sql/base/mangos.sql
+++ b/sql/base/mangos.sql
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `db_version`;
 CREATE TABLE `db_version` (
   `version` varchar(120) DEFAULT NULL,
   `creature_ai_version` varchar(120) DEFAULT NULL,
-  `required_z2745_01_mangos_game_event_split` bit(1) DEFAULT NULL
+  `required_z2746_01_mangos_world_safe_locs_facing` bit(1) DEFAULT NULL
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC COMMENT='Used DB version notes';
 
 --
--- a/sql/updates/mangos/z2746_01_mangos_world_safe_locs_facing.sql
+++ b/sql/updates/mangos/z2746_01_mangos_world_safe_locs_facing.sql
@@ -1,0 +1,11 @@
+ALTER TABLE db_version CHANGE COLUMN required_z2745_01_mangos_game_event_split required_z2746_01_mangos_world_safe_locs_facing bit;
+
+CREATE TABLE IF NOT EXISTS `world_safe_locs_facing` (
+  `id` int(10) unsigned NOT NULL,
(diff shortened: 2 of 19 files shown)


killerwife:
Hm, would it be too much of a bother to move the xyz to DB as well? That way its all in one place.

ratkosrb:
You mean ignore the dbc file completely and move this data to db? I can do that, sure.

killerwife:
Yeah. It would be more consistent :)

ratkosrb:
I have updated it to load everything from the database.

cala:
Thank you for this PR.
Sounds very nice and mergeable as it is to me. I also prefer to have as much content data as possible in the DB.
@killerwife, your opinion ?

killerwife:
2 things - Travis failed so compilation hit a snag somewhere
And we will need to add more data for tbc and wotlk branches for this. @ratkosrb is it possible to gather the tbc and wotlk data from like a legion client, do you have the data or should I go ask nicely for it to TC.
When you fix compilation, all we need is that tbc and wotlk data and we can merge it.

ratkosrb:
I am assuming you mean only take the orientation values from a newer client, and the rest from the actual patch appropriate dbc? Cause I don't think its a good idea to take all data from a post-cataclysm version where the world is changed. I can convert the 2.4.3 and 3.3.5 dbcs into sql and assign orientation for old world from Classic, and for Outlands/Northrend from Mists of Pandaria (that's when they added it to dbc).

About the compilation error, it seems MSVC is happy to compile without the needed header includes, but Travis is not so generous.

killerwife:
Correct, since you have all of vanilla from classic, then use this here, and then take tbc dbc xyz (and wotlk xyz) and add orientation to them from earliest possible, if convenient.

ratkosrb:
Travis compiles fine on Linux now, and the error on OSX is not related to my changes.

Here is the tbc and wotlk data.
https://gist.github.com/ratkosrb/fabe874468c4c51f68ffa66428152ee5
https://gist.github.com/ratkosrb/85659c91216a1905222a1968560b5f4e

cala:
No that we have relevant data for also TBC and WotLK. Any objection on merging or can I proceed?

killerwife:
Please do merge it to all cores if you have the time to do so :) I like this PR quite a bit.

cala:
Merged. Thank you very much. 👍 
I will port this to TBC and WotLK on next Classic -> TBC -> WotLK round (probably this week).

PR #410, opened 2019-12-28
Improvements to conditions system.

ratkosrb:
Porting some changes I did to the db conditions system for Light's Hope, as discussed with killerwife.

- Move the conditions code to its own source file.
- Add fields value3 and value4. You can now check up to 4 other conditions with the OR and AND conditions, reducing the amount of chaining required.
- Add a flags field, with REVERSE_RESULT and SWAP_TARGETS options. This allows for several redundant condition types that just return the reverse of another condition to be removed.
- Generalize the way target requirements for conditions are checked. Now instead of writing specific target checks for every condition in a switch case, you just define the target requirement for this type using an enum value.
- You can now provide any WorldObject as the target, if the condition type supports it. Before the target param was only Player, limiting the potential use cases of db conditions.
- Added a condition_id field to the dbscript tables. You can have some script actions only execute in specific circumstances this way.

List of conditions that were made redundant and removed:
- CONDITION_NO_AURA replaced by CONDITION_AURA + FLAG_REVERSE_RESULT.
- CONDITION_NOITEM replaced by CONDITION_ITEM + FLAG_REVERSE_RESULT.
- CONDITION_NOITEM_WITH_BANK replaced by CONDITION_ITEM_WITH_BANK + FLAG_REVERSE_RESULT.
- CONDITION_NOT_ACTIVE_GAME_EVENT replaced by CONDITION_ACTIVE_GAME_EVENT + FLAG_REVERSE_RESULT.
- CONDITION_NOT_ACTIVE_HOLIDAY replaced by CONDITION_ACTIVE_HOLIDAY + FLAG_REVERSE_RESULT.
- CONDITION_SOURCE_AURA replaced by CONDITION_AURA + FLAG_SWAP_TARGETS
- CONDITION_GENDER_NPC replaced by CONDITION_GENDER + FLAG_SWAP_TARGETS


killerwife:
@cala please do review this. I had a long talk with him about it and seems like its beneficial to me. You know my domain is more c++ than DB. @Grz3s might want to look too

cyberium:
Nice PR, thank you for your effort.

src/game/DBScripts/ScriptMgr.h
```diff
@@ -433,7 +433,7 @@ struct ScriptInfo
     float y;
     float z;
     float o;
-    uint32 condition_id;
+    uint32 condition_id = 0;
```
evil-at-wow: It's a bit strange - and certainly not very consistent - to see this is the only member that's initialized here. I would assume the other members, all built-in types I think, also need some initialization. And I think that's done in ```ScriptMgr::LoadScripts()``` for the other members (whether that's a good idea or not is another discussion).

You have rightlfully adjusted ```ScriptMgr::LoadScripts()``` as well to initialize condition_id, so I'm wondering where this uninitialized condition_id was coming from? I'm going to guess it's related to ```ScriptInfo generateActivateCommand()``` in SpellEffects.cpp that you didn't update (but probably should).

ratkosrb: Yep, I made sure it is always initialized because of generateActivateCommand. When we first made this change on LH we had the displeasure of having to figure out why stuff is just randomly not working for seemingly no reason, and it turned out that a ScriptInfo can be generated from inside the core without initializing all members of the structure, thus leading to the condition_id sometimes being a random number preventing the script command from executing.

I thought it's a good idea to just make sure it is always initialized in all possible circumstances, saving future people the pain of debugging it if they decide to do what is done in generateActivateCommand and utilize a ScriptInfo object but forget to initialize the condition id.

Is it okay if I add a constructor to ScriptInfo and initialize all members for consistency? I think it's a better solution than just setting the condition to 0 inside generateActivateCommand, as it prevents this situation from occurring in the future as well. The C++ core guidelines says to always initialize your member variables.

ratkosrb:
I have moved the condition files to Globals as requested.

src/game/DBScripts/ScriptMgr.h
evil-at-wow: Yes, in my opinion it's certainly fine to add a constructor and initialize all members there. That's actually what I was referring to when I questioned whether initializing all members in ```ScriptMgr::LoadScripts()``` was a good idea. Because as you mention: any time you create a ScriptInfo object somewhere else, you'll basically start with random values for the members and forgetting to initialize one somewhere is too easy.

An alternative to writing a default constructor could be braced initialization of all the members, but we don't seem to do that often, so I'm fine with just a constructor as well.

cala:
Looks nice at first glance. I need to think over it a bit more (very tired, here). 👍

cala:
Unless someone vouches against it, I'm going to merge this very nice PR extremely soon.

cala:
Merged in https://github.com/cmangos/mangos-classic/commit/01bf6303cd9d8f9ee76b84dcef6e49583bba016c
Thank you very much. 👍 

My apologies, I hit the merge/push button before squashing these commits into a single one with a neat commit message from the first message here. :confirmed

PR #411, opened 2019-12-29
Implement modify commands for stats.

ratkosrb:
This adds .modify commands for all stats. It's a very nice feature to have when testing things, or even just for fun. And unlike the commands to modify hp and mana, these will not reset when you change items or are hit with an aura that changes your stats. Implemented using spells as requested by killerwife.

You can use this to clear all modifications:

> .reset mods

### DIFF
--- a/sql/base/dbc/cmangos_fixes/Spell.sql
+++ b/sql/base/dbc/cmangos_fixes/Spell.sql
@@ -78,3 +78,24 @@ UPDATE spell_template SET EffectRadiusIndex1=23 WHERE id=29234;
 
 -- Remove incorrect spell attribute for spell 30122 (Plague Cloud) used in Heigan the Unclean encounter
 UPDATE spell_template SET AttributesEx=(AttributesEx&~0x00000040) WHERE id=30122;
+
+-- Custom spells to be used with .modify commands. These spell ids are free in all expansions.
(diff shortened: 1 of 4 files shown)


cala:
Poking @cyberium for review.

ratkosrb:
It appears that only the spell power spell is problematic. It is not used by anything in vanilla, but it is used by some items in tbc/wotlk. I'll look for a replacement. The other client spells for the base stats are unused in all expansions.

The custom spell ids that I've added are free in all expansions.

ratkosrb:
I have replaced the spell power spell. This one isn't used.
http://classic.cavernoftime.com/spell=22747
http://tbc.cavernoftime.com/spell=22747
http://wotlk.cavernoftime.com/spell=22747

cala:
Merged. Thank you !

Yatzii93:
If there was a PogU emote on github I would of used it here! This is BIG for testing purposes. Awesome stuff. Will most likely test it in my tbc core sometimes this week.

miraco:
> 
>     * Are those spell used buy any spell script/creature? What will happen in that case? Why not create our own custom spell like with spell id 65000? (now possible since we have spell_template)

Just a question to this, is it so simple to add a custom spell using spell_template without editing dbc? Could use this for some fun server scripts then

ratkosrb:
Yes, it's easy to add a custom spell server side, but you can't see it in the client unless you modify Spell.dbc and make a MPQ patch. So if you were thinking of making custom abilities, that won't work without client modifications. For the purposes of these commands we don't need to see the auras in the client, so its perfectly fine to use custom spells.

PR #412, opened 2020-01-05
Improve implementation of learn all command.

ratkosrb:
Instead of hardcoding a giant list of spell ids, iterate over the spells template and learn all non passive and non talent player abilities that are meant to be teachable to players, so only spells that have an associated teaching spell with effect 36.

The current list appears to be completely arbitrary. It contains the following:
- Some mage class spells.
- Spells with effect 36 (learn spell) that are meant to be used by a trainer, instead of the actual ability. 
- Some random pet and creature spells.
- Druid cat and bear form abilities.
- Deprecated spells with [PH] and zzOLD in the name.
- Human racial abilities.
- Some weapon skills.
- A few enchanting recipes.
- A few blacksmithing recipes.
- Random stuff like "Honorless Target", "Opening", "SPELLDEFENSE (DND)".

The list's author also appears to have been completely oblivious to the fact that spells are different per expansion, since the list is the same in all cores, and some of the stuff it attempts to teach you does not even exist.


cala:
Looks very nice to me and far more logical than the previous spell list.
If @killerwife sees no objection, I will merge this.

killerwife:
I approve.

Yatzii93:
image

Oddly it doesnt seem to work well my end. These are the spells it learnt on a fresh warrior. Unless this is intended?

ratkosrb:
The intent is to learn all player spells regardless of class, excluding passives and talents. If you want to learn only things for your own class, there already exists "all_myclass", "all_myspells", and "all_mytalent" sub commands.

cala:
Merged, thank you!

PR #414, opened 2020-01-13
Adjust the source code and build environment to enable ARM32 support.

sterlingalexander:
## 🍰 Pullrequest
Hi,

I'm submitting this pull request for arch support, the changes will enable building/running cMangos on ARM32.  I have been carrying this commit for about 5 months now with no issues, periodically merging it into cmangos's master branch with no conflicts (I also successfully tested x86_64 each time I did a merge).  It works really well on a RPi4 with 4G RAM.  I also offer to maintain ARM32 if required to get the arch enabled.  I tested all available build options, the only issue is the extractors.  See the issues section.

### Proof
I can provide access to my running instance for verification, or you can pull and build my branch, which I last sync'ed 1/5/2020.

### Issues
- Extractors do not build on ARM32 at this time - I did not investigate this as I used an x86_64 host for extraction (and I expect most people will as well).  I would be willing to investigate this issue if it is required to make this PR successful.

### How2Test
On a 32-bit ARM host:
- $ git clone https://github.com/sterlingalexander/mangos-classic.git
- build and run as usual

### Todo / Checklist
- Investigate why the extractors fail to build
- Stress test with more players?  My max in testing was 11, however we made heavy use of playerbot's.


cala:
Requesting @cyberium review for that.

sterlingalexander:
I see the clang build for OSX failed, I don't have access to an OSX environment to test on, but I'm actively looking for one.  Looking at other 3 recent pull requests, I see the same failure on all OSX builds since Dec 11th.  I also randomly sampled the travis-ci logs and see the same type of failure consistently since build #2467 (commit c665a9c).  I don't see anything in that commit that looks contributory either...so at this point I feel pretty good that this PR didn't break OSX builds.

If I get an environment to work on I'll see if I can figure anything out.

sterlingalexander:
Small correction, the extractors build but segfault upon running.  I think the approach of warning and refusing to build them is reasonable.  I'll get it worked in today if I can, otherwise I'll be on vacation for a week, so it will be early February.

cyberium:
Perfect no hurry.

evil-at-wow:
@sterlingalexander Don't worry about the macOS builds, those were already broken because of changes on the Travis CI machines we don't handle well - homebrew updating the MySQL version from 5.7 to 8 started it all. It's being sorting out via cmangos/mangos-tbc#359.

Also, I'm not surprised to see you need fixes for unaligned memory access on ARM. We usually get away with them on x86/x64 because those processors are very forgiving and _most_ instructions can handle an unaligned acces. But the fact remains this is very likely undefined behavior in C++, so I'm wondering if we shouldn't just go for the memcpy on all platforms. It's a very portable and safe option.

sterlingalexander:
I just pushed the modified cmake rules to warn/avoid extractor builds on ARM.  Let me know if there is anything else you would like me to change/work on.

cyberium:
Merged thank you. it will be nice if someone can confirm that it work on tbc, wotlk.

PR #416, opened 2020-02-01
Only create databases when they don't already exist

canidae:
## 🍰 Pullrequest
A quick way to reset the various databases is to drop them and recreate them using the sql scripts, but you don't always want to reset them all.
This change will cause non-existing databases to be created while leaving the existing databases unchanged.


cyberium:
Thank you for this.

I understand your concern. But its not meant to be used to update a db.
`create/db_create_mysql.sql`
Its clear that we want to create db and if it fail we'll get an wanted error.

This merged can silently make some to think that they created a fresh db while they will reuse an older one.

PR #417, opened 2020-05-12
fixed issue #2203

yosefl20:
## 🍰 Pullrequest
fixed issue #2203

### Proof
- My home server runs 18 hours till now. And between 00:00-07:00 no one is playing there.

### Issues
- [2203](https://github.com/cmangos/issues/issues/2203)The client cannot connect to the game server. and the log of mangosd tells "MySQL has gone away" after a serveral hours when there is no one playing the game.

### How2Test
- Run mangosd and realmd with MySQL 5.6
- Setup a demo account
- Don't login in 8 hours
- Launch a client and login

### Todo / Checklist
- [X] None

### DIFF
--- a/contrib/mysql_to_pgsql/src/main.cpp
+++ b/contrib/mysql_to_pgsql/src/main.cpp
@@ -18,7 +18,7 @@
 
 #include "defines.h"
 
-int main(int argc, char* argv[])
+int main(int argc, char *argv[])
 {
     char sPGhost[26], sPGport[26], sPGdb[26], sPGuser[26], sPGpass[26];
     printf("Postgres connection settings\n Host>");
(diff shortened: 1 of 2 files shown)


evil-at-wow:
I know it's rather inconsistent in the current code, but I believe our code style puts the pointers and references on the side of the type, so:
```
const char* something;
somefunction( const std::string& name ) { ...}
```

rather than

```
const char *something;
somefunction( const std::string &name ) { ...}
```
So no need to change those.

cala:
Will look at this soon. Thank you.

yosefl20:
I fixed that issue by decrease mysql ping time to 10min yesterday. And it is fixed without changing any code. So this PR is irrelevant. Thanks evil-at-wow and cala for reviewing it.

PR #418, opened 2020-05-16
Fixed fishing from pools while in group

TarasHots:
## 🍰 Pullrequest
Fixed issue [2074](https://github.com/cmangos/issues/issues/2074)

### Proof
Fixed not only group loot but also master and round robin loots. Player can easily fish from pools and can loot them. Other party members just see what he received, and loot type has no influence on fishing from pools. Fishing outside of pools worked before and works after fix.

### Issues
- fixes https://github.com/cmangos/issues/issues/2074

### How2Test
- Create or take character with enough fishing skill to fish from pools
- Find any kind of pool (e.g. Oily Blackmouth School or Floating Debris) 
- Be in group with group loot (master and round robin can should also be checked)
- Group member should be near you
- Try fishing from the pool
- On successful try, see that you can loot goodies safely

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Loot/LootMgr.cpp
+++ b/src/game/Loot/LootMgr.cpp
@@ -471,7 +471,7 @@ LootSlotType LootItem::GetSlotTypeForSharedLoot(Player const* player, Loot const
         {
             if (!isBlocked)
             {
-                if (loot->m_isChest)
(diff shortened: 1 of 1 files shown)


cala:
Poking @cyberium who is working on loots right now.

TarasHots:
@cyberium please, let me know if I need correct something

PR #419, opened 2020-07-11
Triage quest fix

TarasHots:
## 🍰 Pullrequest
Fixed issue [2074](https://github.com/cmangos/issues/issues/2226)

### Proof
This is how patients looks like. They have owner now, and quest works correctly, they can even die!
Screenshot from 2020-07-11 11-39-50

Quest objectives reached! No Evade when casting the spell
Screenshot from 2020-07-11 11-40-58

Quest completed as usual, so let's try to finish it
Screenshot from 2020-07-11 11-41-15

First Aid skill before...
Screenshot from 2020-07-11 11-41-36

First Aid skill after
Screenshot from 2020-07-11 11-41-48

I haven't tried for Alliance, but as long as fix is not faction-specific it should work good

### Issues
- fixes https://github.com/cmangos/issues/issues/2226

### How2Test

- Have First Aid skill topped near 225
- Visit faction-specific NPC to get this quest
- Take quest and start `healing` the soldiers, witness they are not evading your heals
- Be a good doc and heal 15 patients
- Complete the quest

### Todo / Checklist

- [ ] Is it fine for patients to have this sub name?
- [ ] Maybe this change should not be in `npcs_special.cpp` and somewhere globally, let me know

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
@@ -263,7 +263,7 @@ struct npc_injured_patientAI : public ScriptedAI
         if (pCaster->GetTypeId() == TYPEID_PLAYER && m_creature->IsAlive() && pSpell->Id == 20804)
         {
             Player* pPlayer = static_cast<Player*>(pCaster);
-            if (pPlayer->GetQuestStatus(6624) == QUEST_STATUS_INCOMPLETE || pPlayer->GetQuestStatus(6622) == QUEST_STATUS_INCOMPLETE)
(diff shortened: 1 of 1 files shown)


TarasHots:
I believe this is not a final solution yet. Issue with quest is because patients are in combat and this helps to not regenerate theirs health. Adding owner only helps to prevent evade, but that sub name is not cool + sounds this is not really fixed in the end. I've tried to set `RegenerateStats` to 0, it helped to not regenerate health, but soldiers despawned really quickly even before you manage to cast the bandage...

canidae:
I've tested this change and it will make the quest possible to complete.
Not sure if this is the correct way to do it, but it's a simple fix until a proper solution is available.

TarasHots:
I came up even with better fix, works just in the same way as previous one. No sub-name this time, looks quite clean so far

PR #421, opened 2020-08-16
Hearthstone pet fix

TarasHots:
## 🍰 Pullrequest
While playing WoW locally, I noticed that hearthstone did not teleport pet correctly.  They become unresponsive, you can't dismiss them and pet appears only after logout. If distance is small, pet will try running to you, which does not look "natural". 

### Proof
Captured few gifs so you can witness how fix works

warlock
hunter

### Issues

- None. Decided to create PR straight away

### How2Test

1. Have class with pet as it's part (e.g. warlock/hunter)
2. Use hearthstone 
3. See that pet is teleport along with you

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -9949,7 +9949,7 @@ void Unit::RemoveAurasAtMechanicImmunity(uint32 mechMask, uint32 exceptSpellId, 
 void Unit::NearTeleportTo(float x, float y, float z, float orientation, bool casting /*= false*/)
 {
     if (GetTypeId() == TYPEID_PLAYER)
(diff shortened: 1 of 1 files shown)


Crypticaz:
I believe this issue also occurs in tbc version as well. However, is this how it's handled in classic wow? Is the pet supposed to disappear on teleport or stay with the player?

TarasHots:
Specifically this flag removes pet only temporarily, and it is returned back later.

To be honest, I don't really know how it should be on vanilla, but state without this change is broken for sure, because pet is neither unsummoned nor teleported, it just left as it is, and pet, tries to reach player with pathing, which might lead to crashes sometimes.

At this point I made it convenient from player perspective, how I'd expect and it may be wrong.

In tbc, pet is taken with player afaik. But I played it 7-8 years ago, so...

cala:
> Is the pet supposed to disappear on teleport or stay with the player?

Pet is teleported with player.

cala:
Merged. Thanks @TarasHots :

PR #424, opened 2020-09-22
Feature cleanup playerbot nullable args

bogglez:
## 🍰 Pullrequest
I found the Playerbot code to be difficult to comprehend, because it uses nullable parameters everywhere.
In this PR I changed manager, master, bot, and ai from nullable pointers to references. This makes dead code and redundantly defensive programming obvious (code like `if (!always_set_pointer) return ERROR` or `if (always_set_pointer && f())`.

The code repeats itself a lot, therefore the changes in this PR are unfortunately many, but follow a fixed pattern. They were mostly performed using regex patterns. Mostly `a->b` had to be changed to `a.b`, and tautologies were removed (`if (!m_ai) return ERROR)` etc).

In a future PR I intend to do the same with targets, but didn't want to make this PR too big. I have already noticed a couple of bugs that I intend to submit fixes for thanks to these changes.

By the way, I think the Playerbot code could be simplified a lot by using behavior trees instead of a state machines (https://arxiv.org/pdf/1709.00084.pdf). Not sure how well one could retrofit this into the existing implementation, but there's a lot of duplication among class AIs right now. Playing a dungeon together with the AI also didn't feel "production ready", as bots will stop reacting quite often, run into mob groups to loot something, fail to crowd control, crowd control their kill target, etc. Perhaps this logic could be shared among different classes more easily with behavior trees, while reducing the complexity.

### How2Test
- Play with Playerbot bots. I ran Lower Blackrock Spire with 8 bots in a raid for ~20 minutes.

### DIFF
--- a/src/game/PlayerBot/AI/PlayerbotDruidAI.cpp
+++ b/src/game/PlayerBot/AI/PlayerbotDruidAI.cpp
@@ -20,7 +20,7 @@
 
 class PlayerbotAI;
 
-PlayerbotDruidAI::PlayerbotDruidAI(Player* const master, Player* const bot, PlayerbotAI* const ai) : PlayerbotClassAI(master, bot, ai)
+PlayerbotDruidAI::PlayerbotDruidAI(Player& master, Player& bot, PlayerbotAI& ai) : PlayerbotClassAI(&master, &bot, &ai)
 {
     MOONFIRE                      = m_ai->initSpell(MOONFIRE_1); // attacks
     STARFIRE                      = m_ai->initSpell(STARFIRE_1);
--- a/src/game/PlayerBot/AI/PlayerbotDruidAI.cpp
+++ b/src/game/PlayerBot/AI/PlayerbotDruidAI.cpp
@@ -20,84 +20,84 @@
 
 class PlayerbotAI;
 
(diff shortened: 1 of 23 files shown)


cala:
Greetings and thank you very much for this PR. I will need some time to review it as it brings many changes but this looks very interesting.

Playerbot was developed over the years by several people, with different mind-set, without real guidelines and without always following core policy, sometimes focusing on a specific feature without a global approach. It resulted in what you can actually see in the code that is really heterogeneous, with redundancy though it still does the job of providing bot buddies to play with in a variety of cases.

I try to streamline the code, remove unneeded part of it, simplify it or refactor it when I can spare the time, but this is a very slow process that comes after usual work on the core, scripts and database. Thus, every hand that can be given is much welcome. Like this PR.

bogglez:
Hi cala,

it's only natural for code to grow like this. I'm not complaining, just identifying some issues to work on, and try to fix them. :)
But before I add new features or rewrite behavior I think I have to refactor a bit in order to avoid breakage. So sorry for the big commit.

cala:
Just a side note: this is still on my to do list.

cala:
@bogglez Sorry for the delay in reviewing this, but like many people this year, I had a lot of crisis-things to handle lately.
Merging this, thank you. If you have more improvements to bring to PlayerBot, I'll be happy to review and test them.

cala:
On a side note, I'm porting this to TBC and WotLK core, so you don't have to worry about that.

PR #425, opened 2020-09-22
Playerbot: Fix 2 crash bugs

bogglez:
## 🍰 Pullrequest
Fixes two server crashing bugs when Playerbot is active.

### How2Test
- Form a group with bots.
- Leave the group.
- Type "/p .bot add x" although not in a group.
- Observe server crash due to NPE.


cala:
Nice catch, thank you!

PR #426, opened 2020-09-22
Playerbot Mage AI: Fix Polymorph immunity check on wrong target

bogglez:
## 🍰 Pullrequest
The Playerbot mage AI will attempt to sheep elite mobs attacking it.
However the arcane magic immunity check is performed on the combat target, not on the attacking "target". This isn't necessarily the same.

### How2Test
- Add mage playerbot.
- Have mage attack a target immune to arcane magic.
- Have mage draw aggro from a beast.
- Mage will not try to sheep the beast, because his main target is immune to arcane magic.


cala:
Greetings and thank you very much for this PR. This error is not a big deal as very little creatures are immune to arcane, but for those who are not, we need a correct check. Merging. :+1:

PR #433, opened 2020-11-03
Update names of unit byte fields

ratkosrb:
## 🍰 Pullrequest
- Adds offsets enum for accessing the byte fields, like it's done in Trinity.
- Correct names of Bytes1 flags based on research.

### Proof
The flags that reside in UNIT_FIELD_BYTES_1 offset 3 in Vanilla and TBC are visibility related flags. In Wotlk+ these flags were moved to offset 2.
Based on data from many Classic sniffs, I've come to the conclusion that flag 1 is for Ghosts. All the creatures that I see with it are ghosts, and in my own data they all have an aura of type SPELL_AURA_GHOST. However killerwife seems to have found some creatures in Legion sniffs who do not have an aura of that type, and instead use an invisibility aura, but all of them are still ghosts, so this name fits the flag. It is also applied to players when they die and release spirit (by the ghost aura that is applied on them).
Flag 2 is only found on creatures with an aura of type SPELL_AURA_MOD_STEALTH.
Flag 4 is only found on creatures with an aura of type SPELL_AURA_UNTRACKABLE.

### Issues
In my view, we shouldn't be setting the VisFlags when loading `creature_addon`, as they seem to be applied by auras, but I didn't change this in this pull request. Also it is not clear if all of the current classic db data for `creature_addon` even comes from vanilla sniffs, which is very important, because these flags were moved to offset 2 in Wotlk, which would mean they are being toggled incorrectly in some circumstances.

Another thing we don't know is what UNIT_FIELD_BYTES_2 offset 3 means, it's not currently used in core. In TBC+ this is where the shapeshift form is written, but in Vanilla shapeshift form is in UNIT_FIELD_BYTES_1.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/blackwing_lair/boss_nefarian.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/blackwing_lair/boss_nefarian.cpp
@@ -107,7 +107,7 @@ struct boss_nefarianAI : public CombatAI
         m_creature->SetHover(true);
         SetMeleeEnabled(false);
         SetCombatMovement(false);
-        // m_creature->SetByteValue(UNIT_FIELD_BYTES_1, 3, UNIT_BYTE1_FLAG_ALWAYS_STAND);
+        // m_creature->SetByteValue(UNIT_FIELD_BYTES_1, 3, UNIT_BYTE1_FLAG_ALWAYS_STAND); // wotlk+
     }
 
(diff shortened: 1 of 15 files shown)


killerwife:
I requested this PR. Its not merged because of merge insanity for me when porting.

cyberium:
@killerwife request accepted but you will have to do it sooner or later.

killerwife:
Merged in https://github.com/cmangos/mangos-classic/commit/81f20f4f47412a9e84ab600621c88692d1747d56

PR #435, opened 2020-11-23
Prevent stealth ability from getting stuck when Fearie Fired

canidae:
## 🍰 Pullrequest
Attempting to cast Stealth while under the influence of Faerie Fire resulted in Stealth ability becoming permanently locked until the player relogged.
This may not be a perfect solution, but it does improve on an annoying bug.
I've tested Prowl (Druid), Shadowmeld (Night Elf), Stealth (Rogue), Vanish (Rogue) and "Lesser Invisibility Potion".
Prowl, Shadowmeld, Stealth and the potion were all prevented from being used. Potion was not lost.
Vanish was not prevented, ability entered cooldown and a Flash Powder was consumed. It looks like it's correct behavior that Faerie Fire prevents Vanish from working, but I'm uncertain whether it's correct that ability should enter cooldown and a reagent to be consumed.

Credit where credit is due; @darklordsqk wrote the patch for resolving this issue. I slightly modified the code and moved it to `CheckCast()` rather than `CheckCasterAuras()`, but I'm not sure whether that was an improvement.

### Proof
- None

### Issues
- Fixes https://github.com/cmangos/issues/issues/1880

### How2Test
- .aura 9907 (Faerie Fire debuff)
- Attempt to stealth, fails with "You can't do that yet", stealth ability will not become permanently locked after this patch

### Todo / Checklist
- See pull description

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -4214,6 +4214,17 @@ SpellCastResult Spell::CheckCast(bool strict)
         return SPELL_FAILED_CASTER_AURASTATE;
 
     Unit* target = m_targets.getUnitTarget();
+    uint32 affectedMask = GetCheckCastEffectMask(m_spellInfo);
+    // Check if caster is prevented from entering stealth/invisibility (e.g. Faerie Fire)
+    if (!target)
+    {
+        uint32 selfImmuneMask = GetCheckCastSelfEffectMask(m_spellInfo);
+        if (selfImmuneMask)
+        {
+            target = m_caster;
+            affectedMask = selfImmuneMask;
+        }
+    }
     if (target)
     {
         // TARGET_UNIT_SCRIPT_NEAR_CASTER fills unit target per client requirement but should not be checked against common things
--- a/src/game/Spells/Spell.h
+++ b/src/game/Spells/Spell.h
@@ -383,6 +383,53 @@ class Spell
         void TakeReagents();
         void TakeCastItem();
 
+        inline bool HasSpellTarget(SpellEntry const* spellInfo, uint32 target)
+        {
+            for (uint32 i = 0; i < MAX_EFFECT_INDEX; ++i)
+                if (spellInfo->EffectImplicitTargetA[i] == target)
+                    return true;
+
+            return false;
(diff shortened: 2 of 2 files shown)


killerwife:
The reason why this is broken right now, is this piece of code here that we discovered to be faulty last week.
`            if (IsPositiveSpell(m_spellInfo->Id) && !m_spellInfo->HasAttribute(SPELL_ATTR_UNAFFECTED_BY_INVULNERABILITY))
                if (target->IsImmuneToSpell(m_spellInfo, target == m_caster, 7) && !target->hasUnitState(UNIT_STAT_ISOLATED))
                    return SPELL_FAILED_TARGET_AURASTATE;`

This checks for dispel immunity, but atm its not routed there on cmangos for self.

`    if (!target)
    {
        uint32 selfImmuneMask = GetCheckCastSelfEffectMask(m_spellInfo);
        if (selfImmuneMask)
        {
            target = m_caster;
            affectedMask = selfImmuneMask;
        }
    }`

and

`inline bool HasSpellTarget(SpellEntry const* spellInfo, uint32 target)
{
    for (uint32 i = 0; i < MAX_EFFECT_INDEX; ++i)
        if (spellInfo->EffectImplicitTargetA[i] == target)
            return true;

    return false;
}

inline uint32 GetCheckCastSelfEffectMask(SpellEntry const* spellInfo)
{
    uint32 resultingMask = 0;
    for (uint32 i = 0; i < MAX_EFFECT_INDEX; ++i)
        if (HasSpellTarget(spellInfo, TARGET_UNIT_CASTER))
            resultingMask |= (1 << i);
    return resultingMask;
}`

This should fix the issue, when put in Spell::CheckCast before if (target) block. Check it out, and let me know if it works.

canidae:
Great, I will give it a try in a moment. If it fixes the issue I'll assume you've already fixed it somewhere and that it'll be merged to the cores shortly, and I'll close this PR.

canidae:
I will close this PR as @killerwife likely will merge a better fix in the future.
For reference and the impatient, I pushed the necessary changes to this branch: https://github.com/canidae/mangos-classic/tree/faerie_fix

PR #436, opened 2020-11-23
Replaced "my_bool" with "bool" used by newer MySQL

canidae:
## 🍰 Pullrequest
The cores use an outdated `my_bool` datatype which was removed from MySQL in version 8.0.1.
According to a MySQL dev: «The recommended fix is to use bool or int, which will work both in 8.0 and older versions.»

### Proof
- https://bugs.mysql.com/bug.php?id=85131

### Issues
- https://github.com/cmangos/issues/issues/2285

### How2Test
- None

### Todo / Checklist
- I've not built any cores on Windows, but classic core compiles fine both in Ubuntu:focal (libmysqlclient-dev) and ubuntu:bionic (libmysqlclient20) when datatype is changed from "my_bool" to "bool".
- [ ] Check for (unprotected) usage of reserved words in MySQL (there are no issues with MariaDB), e.g: https://github.com/cmangos/mangos-classic/commit/b2a4ffbf2bad3b7a1d515edd6bca71dc1d3c080d

### DIFF
--- a/src/shared/Database/DatabaseMysql.cpp
+++ b/src/shared/Database/DatabaseMysql.cpp
@@ -417,7 +417,7 @@ void MySqlPreparedStatement::addParam(unsigned int nIndex, const SqlStmtFieldDat
 
     MYSQL_BIND& pData = m_pInputArgs[nIndex];
 
-    my_bool bUnsigned = 0;
+    bool bUnsigned = 0;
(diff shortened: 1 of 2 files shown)


metacreo:
Please pull this in WotLk too.

canidae:
Fixed in WotLK, closing this as classic likely will see same fix.

PR #446, opened 2021-05-17
Implement hiding of update fields based on flags.

ratkosrb:
## 🍰 Pullrequest
This reduces network traffic by only sending the update fields that should be visible to the client.

### How2Test
- Login with sniffer attached and examine sniff to confirm only public update fields are sent to client.
- Use beast lore on a mob to confirm private fields are properly revealed on aura application.


al3xc1985:
@ratkosrb u are the man

PR #448, opened 2021-07-21
Improve mmap generator

deiteris:
## 🍰 Pullrequest
The pull request introduces simple multithreading support to the mmap generator and consolidates some threading and generation logic from the bash scripts.
With multithreading support, the mmap generation takes ~18 minutes with 12 threads, while with current implementation it takes >2hr due to large maps being processed by a single CPU thread (even though maps are distributed between multiple mmap generator instances).

### Proof
Test configuration:
CPU: AMD Ryzen 5 2600 @ 4.00 GHz 6 cores / 12 threads
RAM: DDR4-2933 32GB
OS: Windows 10 Pro 21H1

Run with no mmaps generated - ~18 minutes:
[MaNGOSExtractor_detailed.log](https://github.com/cmangos/mangos-classic/files/6855970/MaNGOSExtractor_detailed.log)

Run with all mmaps generated - 1min 30s:
[MaNGOSExtractor_detailed.log](https://github.com/cmangos/mangos-classic/files/6855784/MaNGOSExtractor_detailed.log)

Memory consumption: approx. 20-30mb per thread (i.e., with 12 threads may reach up to 450mb in total depending on a map).
CPU utilization: approx. 70-80% depending on a map.

Before:
image

After:
image

### Issues
- None

### How2Test
- Copy prebuilt extractors to the client directory and run `./ExtractResources.sh`.
- Choose extraction options and wait until the generation finishes.
- Verify that all mmaps and offmeshes generated (should be 1999 elements in total, tested on 1.21.1 client), record the extraction time.

### Todo / Checklist
- New argument and changes arguments positioning in the `MoveMapGen.sh`. Current order is:
  1. Extraction mode (`maps` or `offmesh`).
  2. Log file.
  3. Detailed log file.
  4. Number of threads.
- `MoveMapGen.sh` no longer spawns new processes to multiprocess the maps. `MoveMapGen.exe` is now responsible for multiprocessing.
- `ExtractResources.sh` accepts any number of threads. If the number is not provided - all threads will be used (the argument `--threads` will not be passed to `MoveMapGen.exe` in this case).
- In `MoveMapGen.exe`:
  * `--onlyGO` was renamed to `--buildGameObjects`.
  * Added new `--threads [#]` argument that accepts a number of CPU threads. If not provided, will be equal to `thread::hardware_concurrency()`.
  * `--buildGameObjects` now may be additionally supplied to build game objects, while ` --onlyGO` was building only game objects. This way we can build game objects only (by supplying only this argument), or in combination with map IDs (what was not possible in current implementation, a separate process was launched).
  * Map IDs are now accepted as a space-separated list instead of a single number.

### DIFF
--- a/contrib/extractor_scripts/ExtractResources.sh
+++ b/contrib/extractor_scripts/ExtractResources.sh
@@ -12,7 +12,7 @@
 
 ## Expected param 1 to be 'a' for all, else ask some questions
 
-## Normal log file (if not overwritten by second param
+## Normal log file (if not overwritten by second param)
 LOG_FILE="MaNGOSExtractor.log"
 ## Detailed log file
 DETAIL_LOG_FILE="MaNGOSExtractor_detailed.log"
--- a/contrib/extractor_scripts/ExtractResources.sh
+++ b/contrib/extractor_scripts/ExtractResources.sh
@@ -39,7 +39,7 @@ else
   ## do some questioning!
   echo
   echo "Welcome to helper script to extract required dataz for MaNGOS!"
-  echo "Should all dataz (dbc, maps, vmaps and mmaps be extracted? (y/n)"
+  echo "Should all dataz (dbc, maps, vmaps and mmaps) be extracted? (y/n)"
   read line
   if [ "$line" = "y" ]
   then
@@ -60,7 +60,7 @@ else
 
     echo
     echo "Should mmaps be extracted? (y/n)"
-    echo "WARNING! This will take several hours! (you can later tell to start delayed)"
+    echo "WARNING! This may take several hours with small number of CPU threads!"
     read line
     if [ "$line" = "y" ]
     then
@@ -80,7 +80,7 @@ fi
 ## Special case: Only reextract offmesh tiles
 if [ "$USE_MMAPS_OFFMESH" = "1" ]
 then
-  echo "Only extracting offmesh meshes"
+  echo "Only extracting offmesh tiles"
   MoveMapGen.sh offmesh $LOG_FILE $DETAIL_LOG_FILE
   exit 0
 fi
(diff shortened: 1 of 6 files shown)


deiteris:
By the way, I see the tile coordinates swapped in `m_terrainBuilder->loadVMap(mapID, tileY, tileX, meshData)`
https://github.com/cmangos/mangos-classic/blob/12335d6f02a9e27854cbe74c7b498f573cb7f8d9/contrib/mmap/src/MapBuilder.cpp#L565

But `TerrainBuilder::loadVMap` accepts tileX, tileY. Is that actually ok?
https://github.com/cmangos/mangos-classic/blob/12335d6f02a9e27854cbe74c7b498f573cb7f8d9/contrib/mmap/src/TerrainBuilder.cpp#L563

killerwife:
If I remember correctly some tile value is actually swapped, so yes, its working fine right now. Good PR, I am currently unable to stare at my screen for long periods of time, I will merge this once my vision is back to normal, should anyone get to it before me, I approve this PR.

deiteris:
I've updated the code to use `std::future` instead of `std::thread`. This improves performance a bit with the same config.

Run with no mmaps generated - 16 minutes:
[MaNGOSExtractor_detailed.log](https://github.com/cmangos/mangos-classic/files/6867364/MaNGOSExtractor_detailed.log)

Run with all mmaps generated - 1min:
[MaNGOSExtractor_detailed.log](https://github.com/cmangos/mangos-classic/files/6867363/MaNGOSExtractor_detailed.log)

CPU utilization is ~99% and memory consumption has dropped to 200-250mb in total depending on a map.
image

I guess this is the best result that can be achieved. @killerwife, you might need to reapprove again.

On the side note, multiprocessing can be improved by processing smaller maps on multiple threads (with tiles multiprocessing) first if number of tiles < number of threads, and use only tiles multiprocessing for large maps. This might cut off a bit more time.

cyberium:
Thank you very much for this PR. I did some change to make it use full queue. I hope its not a problem for you that i kept your authorship you did the hard work for this ;)

deiteris:
Sure, no problem. Glad the PR finally made it to the master branch :D

killerwife:
Yeah luckily cyberium took over cos i entangled myself with smth else.

PR #451, opened 2021-08-15
SpellAura guard against nullptr access

insunaa:
## 🍰 Pullrequest

aura is initialized as a nullptr and only gets assigned with a real object inside of an if function, that could also fail, but then a method of the Aura class is called without a guard, which can (and apparently does) lead to the server trying to execute a method from a nullptr, which should be an access violation

### Proof

- None

### Issues

- possibly related to https://github.com/cmangos/issues/issues/2513

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -388,15 +388,6 @@ Aura* CreateAura(SpellEntry const* spellproto, SpellEffectIndex eff, int32 const
         aura = new AreaAura(spellproto, eff, currentDamage, currentBasePoints, holder, target, caster, castItem);
 
     if (!aura)
-    {
-        uint32 triggeredSpellId = spellproto->EffectTriggerSpell[eff];
-        if (SpellEntry const* triggeredSpellInfo = sSpellTemplate.LookupEntry<SpellEntry>(triggeredSpellId))
-            for (unsigned int i : triggeredSpellInfo->EffectImplicitTargetA)
-                if (i == TARGET_UNIT_CHANNEL_TARGET)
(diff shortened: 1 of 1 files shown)


insunaa:
For reference, this is what the function looks like in mangos-tbc
image
```cpp
Aura* CreateAura(SpellEntry const* spellproto, SpellEffectIndex eff, int32 const* currentDamage, int32 const* currentBasePoints, SpellAuraHolder* holder, Unit* target, Unit* caster, Item* castItem)
{
    Aura* aura = nullptr;
    if (IsAreaAuraEffect(spellproto->Effect[eff]))
        aura = new AreaAura(spellproto, eff, currentDamage, currentBasePoints, holder, target, caster, castItem);

    if (!aura)
    {
        uint32 triggeredSpellId = spellproto->EffectTriggerSpell[eff];
        if (SpellEntry const* triggeredSpellInfo = sSpellTemplate.LookupEntry<SpellEntry>(triggeredSpellId))
            for (unsigned int i : triggeredSpellInfo->EffectImplicitTargetA)
                if (i == TARGET_UNIT_CHANNEL_TARGET)
                    aura = new SingleEnemyTargetAura(spellproto, eff, currentDamage, currentBasePoints, holder, target, caster, castItem);
    }

    if (!aura)
        aura = new Aura(spellproto, eff, currentDamage, currentBasePoints, holder, target, caster, castItem);

    aura->OnAuraInit();
    return aura;
}
```

kderholt:
This is the commit done to the mangos-classic that breaks it, resulting in the error you describe:

https://github.com/cmangos/mangos-classic/commit/cfd76ed8951e9c9d804e9dc4191b56ebad523a55

PR #457, opened 2022-01-30
Make CMake copy config files for PlayerBot and AHBot

Niam5:
## 🍰 Pullrequest
Currently post build will not copy AHBot or PlayerBot configs on Window even if built. Linux/Mac OS copies the PlayerBot config file even if feature was not selected. This PR gates the configs behind their respective defines and also fixes the non-copying of configs on windows.

### Proof
- None

### Issues
- None

### How2Test
-Build project with default options(AHBot and PlayerBot off). No configs copied
-Build with either or both options enabled. Corresponding configs are now copied to correct directory.

### Todo / Checklist
- [x] None


killerwife:
@insunaa can u pls check linux if this works for you? I am checking windows

killerwife:
Can confirm it fixes the issue on windows.

PR #460, opened 2022-02-16
Implement areatrigger event involving Justine Demalier and Melris Malagan in Stormwind

Nytanath:
## 🍰 Pullrequest

When a player enters areatrigger 2746, if they satisfy certain conditions (level at or above 56 and quest [The First and the Last](https://tbc.wowhead.com/quest=6182/the-first-and-the-last) not rewarded), [Justine Demalier](https://classic.wowhead.com/npc=12481/justine-demalier) and [Melris Malagan](https://classic.wowhead.com/npc=12480/melris-malagan) do a small roleplay event (they turn toward player, talk and do some emotes). This event then won't be triggered again for roughly 8 minutes after Melris Malagan resets his orientation (the last event they do).

I don't know if that cooldown would be shortened if a different player enters the areatrigger - I only have one account on the PTR to test it and Stormwind is always empty.

This is valid for Classic and TBC. I don't think it's valid for WotLK since the male NPC's text explicitly references Bolvar Fordragon, which in WotLK is no longer the acting king of Stormwind. The relay_script id I used is free on both DBs as of the time of this PR.

### Proof

- A lot of comment on [Justine Demalier's Wowhead page](https://classic.wowhead.com/npc=12481/justine-demalier#comments).

### Issues

- None

### How2Test

- Have your character be at level 56 or higher, and without having done the quest [The First and the Last](https://tbc.wowhead.com/quest=6182/the-first-and-the-last).
- .go creature id 12481
- Notice the NPCs do their event.
- If you want, try to retrigger the event before its cooldown is up by exiting and re-entering the areatrigger (it covers a small patch of land in front of the tree - not big by any means. Move to the innkeeper and back to retrigger it for sure). I already made sure it wouldn't retrigger before the cooldown was up.

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -34,6 +34,7 @@ INSERT INTO scripted_areatrigger VALUES
 (2046,'at_blackrock_spire'),
 (2066,'at_blackrock_spire'),
 (2067,'at_blackrock_spire'),
+(2746,'at_stormwind_recruiter'),
 (3066,'at_ravenholdt'),
 (3146,'at_hive_tower'),
 -- Darnassian bank
--- a/src/game/AI/ScriptDevAI/scripts/world/areatrigger_scripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/areatrigger_scripts.cpp
@@ -387,11 +387,12 @@ enum
 
 bool AreaTrigger_at_stormwind_recruiter(Player* player, AreaTriggerEntry const* /*at*/)
 {
-    if (player->IsGameMaster() || player->GetLevel() < 56 || player->GetQuestRewardStatus(QUEST_THE_FIRST_AND_THE_LAST))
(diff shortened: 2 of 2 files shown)


Grz3s:
```This is valid for Classic and TBC. I don't think it's valid for WotLK since the male NPC's text explicitly references Bolvar Fordragon, which in WotLK is no longer the acting king of Stormwind```

May be worth to check if another simmilar text exist in broadcast_texts .. with name changed because q.6182 still works in wotlk..



> By 128712 on 2008/10/27 (Patch 3.0.2)
>  (Patch 3.0.2)		
>King Varian Wrynn in Stormwind City now starts this quest as of patch 3.0.2 on live realms.


> By Aviry on 2009/07/13 (Patch 3.1.3)
> This quest counts toward Eastern Kingdom Loremaster

Nytanath:
I already checked for similar texts, nothing. Even searching for LIKE "%varian%" returned nothing related to this. Wouldn't be the first rp event that was nuked in WotLK, sadly.

cala:
Thank you. Merged in https://github.com/cmangos/mangos-classic/commit/dd8e5f2cc8603582bca0b07d70939e5e216ef9c9
Sorry for the delay in handling this.

PR #461, opened 2022-02-18
Spell: prevent Lunar Fortune from stacking with itself

Nytanath:
## 🍰 Pullrequest

[Lunar Fortune](https://classic.wowhead.com/spell=26522/lunar-fortune) should not stack with itself. Depending on luck (you have a 50% chance of obtaining this item from the mail you receive when you do the Elder quests during the Lunar Festival) you could get 30+ of those items, giving a 7500 increase of hp if it was allowed to stack.

### Proof

Tested on TBC PTR, it does not stack when using two of those.

### Issues

- None

### How2Test

- .event start 7
- .add 21744 2
- .go obj id 180859
- Use the two Lucky Rocket Clusters.

### DIFF
--- a/src/game/Spells/SpellMgr.h
+++ b/src/game/Spells/SpellMgr.h
@@ -1791,7 +1791,7 @@ inline bool IsStackableAuraEffect(SpellEntry const* entry, SpellEntry const* ent
             nonmui = true;
             break;
         case SPELL_AURA_MOD_INCREASE_HEALTH:
-            if (entry->Id == 26522) // Lunar Fortune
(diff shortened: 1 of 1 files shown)


killerwife:
When preventing stacking, you need to check also for entry2 if it has this entry. This way it will not stack with anything. Also this is like one of the systems that needs a rework in the future really badly :D this is like V 1.5 from the old mangos system, and we really need V2 one day.

Nytanath:
Done, thanks!

killerwife:
Oki, I will merge this later this weekend. Cos i need to push it to other cores too.

PR #462, opened 2022-02-18
Fix behavior for the fireworks that can be used by players during the Lunar Festival

Nytanath:
## 🍰 Pullrequest

The fireworks you can use during the Lunar Festival ([listed here](https://classic.wowhead.com/item=21557/small-red-rocket#shared-cooldown)) have graphical effects that are currently missing.

When you use a firework, a trigger NPC is instantly spawned on the position of the firework/cluster launcher gameobject. This NPC summons one (for fireworks) or five (for clusters) other gameobjects, at certain position offsets compared to its own position.

One second later, the custom animation 3 is sent to the launcher gameobject, which plays its launching animation. At the same time, quest credit for [Lunar Fireworks](https://classic.wowhead.com/quest=8867/lunar-fireworks) is given, and the summoned gameobjects are despawned, which triggers their own firework animation one second later.

One second later, you get the buff from the [Lucky Lunar Cluster](https://classic.wowhead.com/item=21744/lucky-rocket-cluster) if that's the firework you fired.

This results in pretty, shiny fireworks.

All this behavior was confirmed via sniffs. I could not find usage of the serverside spells used to summon the firework gameobjects, but those might just be hidden to the client.

**NOTE**: the spells scripted via SD2 used in this PR do not exist in TBC and WotLK's DBC. They are probably serverside spells in those cores.

TBC is a little different from Classic in that you don't get quest credit directly, but a quest credit spell ([Lunar Festival Cluster Credit Marker](https://www.wowhead.com/spell=44429/lunar-festival-cluster-credit-marker) or [Lunar Festival Rocket Credit Marker](https://www.wowhead.com/spell=44430/lunar-festival-rocket-credit-marker)) is cast from trigger NPCs [Lunar Cluster Credit Marker](https://www.wowhead.com/npc=15894/lunar-cluster-credit-marker) and [Lunar Firework Credit Marker](https://www.wowhead.com/npc=15893/lunar-firework-credit-marker). This does not seem to happen on Classic since those spells were introduced in TBC, but I still spawned the credit NPCs because they exists on Classic. No real use for them though.

Offset between firework positions are calculated from repeated sniffs. They vary depending on the cluster being normal or large (large ones are further away from each other since they're bigger).

Lucky Lunar Cluster always spawns the same five large fireworks: three blue and two white.

There's a bug I could not understand yet: sometime, the timing of the firework gameobjects going off for the clusters are wrong: instead of going off all five at the same time, one goes off and one second later the other four go off. Other times it works correctly. I don't know why.

Also prevent Omen from being summonable via Engineering-created rocket launchers.

### Proof

- None

### Issues

- Fixes [#1547](https://github.com/cmangos/issues/issues/1547) (the issue is closed right now, but the quest was bugged again)

### How2Test
- Create a Horde character (since I'm giving the guid of a NPC in Orgrimmar)
- .event start 7
- .go creature 91608
- Accept the quest Lunar Fireworks
- Buy fireworks from the Lunar Festival Vendor at her right
- Knock yourself out with the fireworks!
- Try some large rockets and clusters too, from [this list](https://classic.wowhead.com/item=21557/small-red-rocket#shared-cooldown) (purple and yellow aren't available to players)

### Todo / Checklist

- [ ] Sometime the five fireworks from a cluster do not fire at the same time. This is just a cosmetic bug though.


src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
```diff
@@ -1107,6 +1110,241 @@ struct npc_advanced_target_dummyAI : public ScriptedAI
...
+    { PATS_FIREWORK_CLUSTER_GUY_YELLOW, { 26321, 26322, 26323, 26324, 26362 } },
+
+    { PATS_FIREWORK_CLUSTER_GUY_LARGE_RED,    { 26501, 26502, 26503, 26504, 26505 } },
+    { PATS_FIREWORK_CLUSTER_GUY_LARGE_BLUE,   { 26483, 26484, 26485, 26486, 26487 } },
+    { PATS_FIREWORK_CLUSTER_GUY_LARGE_GREEN,  { 26491, 26492, 26493, 26494, 26495 } },
+    { PATS_FIREWORK_CLUSTER_GUY_LARGE_WHITE,  { 26506, 26507, 26508, 26509, 26510 } },
+    { PATS_FIREWORK_CLUSTER_GUY_LARGE_YELLOW, { 26511, 26512, 26513, 26514, 26515 } },
+};
+
+struct npc_pats_firework_guyAI : public ScriptedAI
```
killerwife, line 1177: Use the AddCustomTimer function. You will not need to handle the timer and it will just trigger your lambda. The idea is to never have to override UpdateAI function in any script. This is because the base AI has a lot of functionality now, which is generally not the case in other cores.

killerwife: Also the intended usecase of calling AddCustomTimer is in constructor and then setting timer or resetting it on some event. Its akin to EAI GENERIC_TIMER event

killerwife:
Ok here is my analysis paralysis of this. TARGET_LOCATION_SCRIPT_NEAR_CASTER implies it should target an object (creature or GO). It shouldnt have the XYZ supplied from outside. Next, void OnDestTarget(Spell* /*spell*/) const {} exists for exactly this XYZ manipulation off of what TARGET_LOCATION_SCRIPT_NEAR_CASTER  supplies. I will ofc not force you to rework it because I can see that just typing this shit must have been a lot of work :D But putting it out here if you want to make it work like the design likely intended. Blizzard AI is generally what I consider as "stupid", in that they are not capable of putting the code you wrote in AI, but instead it would be housed in a spell script, which is way more powerful. TARGET_LOCATION_SCRIPT_NEAR_CASTER uses spell_script_target table for any targeting, and so far we have not been wrong on that asumption.

killerwife:
Also, I think all the spells in your PR use TARGET_LOCATION_SCRIPT_NEAR_CASTER. If I reviewed it properly. Except the base ones.

Nytanath:
Thanks a lot for your review! I'm learning as I go so I don't mind rewriting it as many times as it takes. OnDestTarget() sounds perfect for this, I didn't notice it when I was searching for hooks to use on Classic. I'll mark this as draft and put it back when I'm done.

Nytanath:
Now it uses a much cleaner script. Thanks for the tips!

al3xc1985:
Guys just a question, woudn't it be better to have the scripts from lunar festival (and all other events this type) sorted out in a special folder and special name... for example lunarfestival.cpp ?? 
To keep it more organise?

killerwife:
Yep, since there is so many of them, a separate file is not a bad idea. We have one for midsummer and others too.

Nytanath:
Made a separate file for Lunar Festival scripts!

sql/base/dbc/cmangos_fixes/Spell.sql
```diff
@@ -207,6 +207,12 @@ UPDATE `spell_template` SET `RangeIndex` = 0 WHERE `Id` IN (16613,16619,16630,16
 -- Toxic Gas - used by Garden Gas in Naxx, remove SPELL_ATTR_EX_CHANNELED_1 and CHANNEL_FLAG_MOVEMENT to prevent aura being removed
 UPDATE spell_template SET AttributesEx=0, ChannelInterruptFlags=0 WHERE Id=30074;
 
+-- Lunar Festival fireworks spells have no missile speed serverside
+UPDATE `spell_template` SET `speed`=0 WHERE `Id` IN (26286, 26291, 26292, 26294, 26295, 26333, 26334, 26336, 26337, 26338, 26304, 26325, 26327, 26328, 26329, 26488, 26490, 26517, 26519, 26518, 26521);
```
killerwife, line 211: Now I realized something. Do you have a sniff of these being casted by any chance? The only reason why I am asking is because speed is ignored for GO casted spells. Which would explain the bad data and lack of "use" in the original spell because a GO cannot use another GO.

killerwife: Nvm, I rechecked, and my asumption is bogus :D

Nytanath: [Here's the relevant packets for one of those spells](https://pastebin.com/K3CEz1S8), just in case :P

The effect is executed instantly, and it is even visible (if you use the firework as far as possible from the launcher, the launcher will trigger before the firework projectile reaches it. Strange, but blizzlike :D).

killerwife:
Closed by https://github.com/cmangos/mangos-classic/commit/7af20f1f8855b28490bfdb1bb05eed18f7b18045

PR #465, opened 2022-02-19
Allow gameobjects activated via ANIMATE_CUSTOM GameObjectAction to trigger scripts

Nytanath:
## 🍰 Pullrequest

Commit https://github.com/cmangos/mangos-classic/commit/ad13e6770a99a6682ff7119929f15f731111520d made it so gameobjects activated via a spell effect using action AnimateCustom0/1/2/3 (example: [Lucky Lunar Rocket](https://classic.wowhead.com/spell=26521/lucky-lunar-rocket) no longer call GameObject::Use().
This prevents activation of gameobject scripts, both from SD2 and from DB.

I can't say whether or not that specific activation should call Use() too, but having it still be scriptable is needed for some scripts.

### Proof

- None

### Issues

- None

### How2Test

- Create a Horde character (since I'm giving the guid of a NPC in Orgrimmar)
- .event start 7
- .go creature 91608
- Accept the quest Lunar Fireworks
- Buy fireworks from the Lunar Festival Vendor at her right

This quest was fixed in https://github.com/cmangos/tbc-db/commit/2a0ae8ff47014b5ef640f116983c5ec4bc7fad09, it relies on this behavior to work.

### DIFF
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -4701,8 +4701,14 @@ void Spell::EffectActivateObject(SpellEffectIndex effIdx)
         case GameObjectActions::ANIMATE_CUSTOM_1:
         case GameObjectActions::ANIMATE_CUSTOM_2:
         case GameObjectActions::ANIMATE_CUSTOM_3:
+        {
             gameObjTarget->SendGameObjectCustomAnim(gameObjTarget->GetObjectGuid(), uint32(action) - uint32(GameObjectActions::ANIMATE_CUSTOM_0));
+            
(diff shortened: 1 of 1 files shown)


killerwife:
You dont need to do it like this. You can just handle OnEffectExecute and use GetGOTarget and do anything you want there. This is the official meaning of the enum so if you want anything extra you need to attach some extra logic to it. Sadly we dont atm have any alternative to EAI for GOs. Need a bigger sample size of what we actually need since GO scripts are hardly researched. So adding OnGameObjectUse is redundant here for sure, not entirely sure about dbscripts autocall.

Nytanath:
I see, thanks. I can just call Use() from the spell script so the db script will still be run :D

PR #468, opened 2022-02-20
Unsummon active Warlock demons when trying to summon a new one

Nytanath:
## 🍰 Pullrequest

On Classic and Classic only, trying to summon a Warlock demon while you already have one summoned will force the current one to despawn as soon as you start casting the new demon's summoning spell.

Added this check here in Spell::CheckCast() instead of Spell::prepare() because there's already code handling pets in the former, and TC handles the Summoning Disorientation stun in there too.

### Proof

- None

### Issues

- Fixes https://github.com/cmangos/issues/issues/2762

### How2Test

- Create a Warlock.
- Summon a pet.
- Try to summon another pet.

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -2853,6 +2853,12 @@ void Spell::Prepare()
 
     OnSuccessfulStart();
 
+    // Unsummon active Warlock demons when trying to summon a new one
+    if (Unit* unitCaster = dynamic_cast<Unit*>(m_trueCaster))
+        if (m_spellInfo->HasAttribute(SPELL_ATTR_EX_DISMISS_PET))
+            if (Pet* pet = unitCaster->GetPet())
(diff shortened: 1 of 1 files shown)


killerwife:
I particularly dislike this being in Spell::CheckCast because CheckCast is not meant to alter the state of any other object than the spell and this interferes with that philosophy. Also it will clash with this change https://github.com/cmangos/mangos-wotlk/commit/39c2848670a0cc6d2d8ccf6cf3358ae6a7f9230e I actually pushed an hour ago, because suddenly checkcast differs. I can change this myself in the merge but I think its important so you know the philosophy. This is also why there exists OnSuccessfulStart and OnCheckCast separately, even though you could derp it like in TC.

Nytanath:
I understand, thanks for the explanation! I see now why that's bad. Moved to Spell::Prepare().

PR #469, opened 2022-02-21
Vanilla LFG (port from vmangos)

celguar:
## 🍰 Pullrequest
Vanilla LFG system (Meeting stones & dbscript support) port from Vmangos, **with tweaks**:
a) fix error of player being queued both solo and in group
b) fix not leaving current solo queue if invited to a group that is also in lfg queue
c) fix requeue after being kicked from group that was in queue
d) tweak default role calculation to make this tool more useful:
   - queued groups always calculate roles by highest talent tree picked (to know what role group is missing)
   - queued solo players calculate role by talents, and if not invited after 5 minutes of queue, rolemask is filled with all possible roles (defaul behavior in vmangos)

Required core mods included:
a) add enums for group add/remove member methods
b) fix missing "You have been removed from the group" message when you are kicked. (Common bug when instead of this message you see "member X left the group" for all members)

**Note**: dbscript data (inkeepers' menus/scripts) will be added in classic-db PR. (if this thing is added, ofc)

### Todo / Checklist
Maybe add some custom features like teleport to Meeting Stone, but I guess this is never gonna be added.


killerwife:
Linked issue https://github.com/cmangos/issues/issues/128

src/game/Groups/Group.cpp
```diff
@@ -354,6 +356,19 @@ bool Group::AddMember(ObjectGuid guid, const char* name)
...
+
+        if (sLFGMgr.IsPlayerInQueue(player->GetObjectGuid()))
+            sLFGMgr.RemovePlayerFromQueue(player->GetObjectGuid());
+        
+        if (isInLFG())
+        {
+            if (joinMethod != GROUP_LFG)
+            {
+                player->GetSession()->SendMeetingstoneSetqueue(m_LFGAreaId, MEETINGSTONE_STATUS_JOINED_QUEUE);
+
```
killerwife, line 368: This space seems redundant

src/game/Groups/Group.cpp
```diff
@@ -479,10 +523,21 @@ void Group::Disband(bool hideDestroy)
...
+            {
+                player->GetSession()->SendMeetingstoneSetqueue(0, MEETINGSTONE_STATUS_NONE);
+            }
         }
 
         _homebindIfInstance(player);
     }
+
+    if (isInLFG())
+    {
```
killerwife, line 537: Brackets also redundant

src/game/Groups/Group.h
```diff
@@ -265,6 +279,14 @@ class Group
         InstanceGroupBind* GetBoundInstance(uint32 mapid);
         BoundInstancesMap& GetBoundInstances() { return m_boundInstances; }
 
+        // LFG
+        void SetLFGAreaId(uint32 areaId) { m_LFGAreaId = areaId; }
+        uint32 GetLFGAreaId() { return m_LFGAreaId; }
+        bool isInLFG() { return (m_LFGAreaId > 0) ? true : false; }
```
killerwife, line 285: As its a public method, it should be uppercase, IsInLFG

src/game/LFG/LFGMgr.h
```diff
@@ -0,0 +1,138 @@
...
+{
+    public:
+        LFGQueue() {}
+        ~LFGQueue() {}
+
+        void AddToQueue(Player* leader, uint32 queAreaID);
+        void RestoreOfflinePlayer(Player* player);
+        bool IsPlayerInQueue(ObjectGuid const& plrGuid) const;
+        bool IsGroupInQueue(uint32 groupId) const;
+        void RemovePlayerFromQueue(ObjectGuid const& plrGuid, PlayerLeaveMethod leaveMethod = PLAYER_CLIENT_LEAVE); // 0 == by default system (cmsg, leader leave), 1 == by lfg system (no need report text you left queu)
```
killerwife, line 102: Using ObjectGuid as reference if its not being assigned to is overkill. ObjectGuid is a single uint64 value which is same as reference as we are using 64-bit systems, and with value the compiler can optimize the copies away.

src/game/LFG/LFGMgr.h
```diff
@@ -0,0 +1,138 @@
...
+        void GetGroupQueueInfo(LFGGroupQueueInfo* info, uint32 groupId);
+
+        static void BuildSetQueuePacket(WorldPacket& data, uint32 areaId, uint8 status);
+        static void BuildMemberAddedPacket(WorldPacket& data, ObjectGuid plrGuid);
+        static void BuildInProgressPacket(WorldPacket& data);
+        static void BuildCompletePacket(WorldPacket& data);
+
+        static ClassRoles CalculateRoles(Classes playerClass);
+        static ClassRoles CalculateTalentRoles(Player* player);
+        static int GetHighestTalentTree(Player* player);
```
killerwife, line 116: use uint32 or int32 instead of int

src/game/LFG/LFGMgr.h
```diff
@@ -0,0 +1,138 @@
...
+        QueuedPlayersMap m_QueuedPlayers;
+        QueuedPlayersMap m_OfflinePlayers;
+
+        typedef std::map<uint32, LFGGroupQueueInfo> QueuedGroupsMap;
+        QueuedGroupsMap m_QueuedGroups;
+
+        void FindInArea(std::list<ObjectGuid>& players, uint32 area, uint32 team, ObjectGuid const& exclude);
+        bool FindRoleToGroup(ObjectGuid playerGuid, Group* group, ClassRoles role);
+
+        uint32 _groupSize = 5;
```
killerwife, line 133: cmangos uses members with m_

src/game/LFG/LFGMgr.h
```diff
@@ -0,0 +1,138 @@
...
+        static ClassRoles CalculateTalentRoles(Player* player);
+        static int GetHighestTalentTree(Player* player);
+        static std::map<uint32, int32> GetTalentTrees(Player* player);
+        static RolesPriority getPriority(Classes playerClass, ClassRoles playerRoles);
+
+        static uint32 GetMaximumDPSSlots() { return 3u; }
+
+    private:
+        typedef std::map<ObjectGuid, LFGPlayerQueueInfo> QueuedPlayersMap;
+        QueuedPlayersMap m_QueuedPlayers;
```
killerwife, line 124: cmangos uses codestyle of m_lowerCase in the newer code (still trying to unify the old code slowly)

src/game/LFG/LFGMgr.h
```diff
@@ -0,0 +1,138 @@
...
+    PLAYER_SYSTEM_LEAVE = 1
+};
+
+enum GroupLeaveMethod
+{
+    GROUP_CLIENT_LEAVE  = 0,
+    GROUP_SYSTEM_LEAVE  = 1
+};
+
+#define MAX_DPS_COUNT = 3
```
killerwife, line 56: Please for small bit of unification, do the enum like here: https://github.com/cmangos/mangos-wotlk/blob/master/src/game/LFG/LFGDefines.h#L157

So that there is at least a shred of unification for LFG across cores in cmangos. (cos rn there isnt :D )

killerwife:
I have made a small tech related review. That being said, I do not know enough about the details for this to be qualified to actually review the functionality. @cala I am leaving this one to you :)

celguar:
Thanks for fast review! I'll work on fixing those.
Also if talents check tweak gonna be an issue I can remove it too. (since from info I read on internets neither vanilla nor tbc LFG tools did absolutely no class or role checks, and you could have 5 tanks or 5 dps from it. Default vmangos way is to check class only and prioritize warrior = tank, priest = healer etc, so you can have 1 warrior who is not tank but system won't look for proper tank)

al3xc1985:
If that îs not blizzlike I suggest add an option to mangosd.conf for checking by talents

src/game/LFG/LFGMgr.h
celguar: So it's never appropriate? I don't understand why they did it, but it seems it was on purpose, because there is this comment
https://github.com/cmangos/mangos-classic/pull/469/files#diff-f02ca7bf1bf633b9e81516bc6acc4c4c5275176077464cf070473fae85a5fb52R511

Do you have any ideas why could they do it? All I know this system supports working with offline players who got DC'ed so it keeps them in queue and doesn't look for replacement.

killerwife: Its fine to use ObjectGuid parameter event ObjectGuid&, but using ObjectGuid const& is overkill and unneeded, because if you cannot assign to it, there is no additional benefit.

celguar: I'm sorry, how to fix it then? I see many similar things in code like this
https://github.com/cmangos/mangos-classic/search?p=1&q=ObjectGuid+const%26
Should I just use `bool IsPlayerInQueue(ObjectGuid plrGuid) const;` ?

al3xc1985:
cause I don't see why not to keep it @celguar . In classic, not everything they did was good. and something that can help is allways welcome..

My opinion!

src/game/LFG/LFGMgr.h
killerwife: Then chill, we will change it during merge :) You will see whats what then.

src/game/LFG/LFGMgr.h
celguar: I'm sorry, do you mean this one only or all enums in lfg code?

killerwife: This one only, the rest would be too much of a pain to impose on you.

PR #470, opened 2022-02-22
Disable AB graveyards inside bases during BG

celguar:
## 🍰 Pullrequest
Disables graveyards inside bases (where you spawn on teleporting into bg) after BG starts. Players who die near base or if all bases are captured by enemy should teleport to the graveyard to the side of the base, where Spirit Healer is. (Without fix closest GY is in most cases = inside base, which is not correct).


killerwife:
Question, what happens if you join bg dead after battle has started?

PR #474, opened 2022-02-22
Fix nullptr crash

celguar:
## 🍰 Pullrequest
Fixes rare crash. Sadly can't give more details on who caused it, it was long ago.
Graveyard one is new though

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -7759,7 +7759,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP || creatureNotInCombat)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else
+                else if (controller->AI())
                     controller->AI()->AttackStart(enemy);
             }
         }
--- a/src/game/Globals/GraveyardManager.cpp
+++ b/src/game/Globals/GraveyardManager.cpp
@@ -65,7 +65,7 @@ WorldSafeLocsEntry const* GraveyardManager::GetClosestGraveyardHelper(GraveYardM
             continue;
 
         // find now nearest graveyard at other (continent) map
-        if (mapId != entry->map_id)
+        if (mapId != entry->map_id && tempEntry)
         {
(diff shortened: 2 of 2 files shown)


killerwife:
Ok so, I will be a bit disappointing here, the AI crash implies a core logic corruption, and I intentionally did not fix this one because of hiding the root cause of the corruption. So the task there is to indeed crash it but perhaps add a comment that if AI is nullptr, this should never happen.

As to the graveyard manager code. This would imply that some Map Entry is missing Instance Template. To actually know what is going on, I would like to know the mapId for which this happened, because we are again just hiding a potentially worse outcome.

I know these arent exactly easy to digest answers, but sadly they are very necessary. I in particular fixed already several causes of the first one on TBC branch. The asumption is that if a player isnt player controlled, he must always have AI. Just like all creatures must have AI.

killerwife:
So if you could post at least a stack trace, that would be great. In TBC the GY one never happens cos the info is directly in Map Template.

celguar:
TBC code for graveyard is just different and has no such issue

celguar:
These are great answers! I just patched these things because they caused crashes on my server with 2k+ random bots. I'll revert these patches and try to catch them again then. I know bots can behave different than humans and cause crashes in unexpected places, but I have no other way of doing it. I just hope it gives proper answers in the end. Thanks for explanation!

celguar:
Also controller->UI() crash was patched by me in early 2021, so maybe it's no longer an issue. I'll try nonetheless

celguar:
_The asumption is that if a player isnt player controlled, he must always have AI._
Okay so I guess this invalidates my info entirely, since I got this while using custom random bots module. Though this one was rare...So, we can close this I guess?

killerwife:
> TBC code for graveyard is just different and has no such issue, maybe it should be ported then?

The idea is that when the info needs to be available, it should be in DB and on load we should make sure all possible combinations are covered and if not data is added. Because then we can potentially fail to tp to a graveyard which is unacceptable. If you show me a stack trace or the missing map ID, I will definitely know whats up. Porting DBC down to vanilla doesnt make sense atm since there are differences. Thats why the supplementary data is in instance template in vanilla core alone.

PR #477, opened 2022-03-02
Implement optional, customizable per-creature waypoint pause timer on interaction, and allow to change the default value in config

Nytanath:
## 🍰 Pullrequest

When you interact with a creature that is moving through a waypoint path, that creature pauses its movement for a certain amount of time. The most common value is 3 minutes, but that can vary for certain creatures. For example, I have confirmed that ["Cookie" McWeaksauce](https://tbc.wowhead.com/npc=17246/cookie-mcweaksauce) pauses for three seconds, and [Jinar'Zillen](https://tbc.wowhead.com/npc=11317/jinarzillen) pauses for 30 seconds.

This PR adds a way to customize that pause timer via the creature_template table, changes the default pause value from 1 to 3 minutes and makes it configurable from config file.

I've chosen to use the value -1 as default to indicate that the creature should use the global default pause timer, because 0 is a valid value that indicates no pause.

### Proof

https://user-images.githubusercontent.com/99603810/156390262-bc3d5237-0735-4b55-9b93-4affa994f93b.mp4

https://user-images.githubusercontent.com/99603810/156390425-213274d3-c3dc-4bd9-8cc4-4e4c3e2509f4.mp4

### Issues

- None

### How2Test

- Run this query (just for testing, actual data is meaningless):
```sql
UPDATE creature_template SET InteractionPauseTimer=10000 WHERE entry=3296;
```
- Teleport to Orgrimmar, talk with a guard, see how long it takes to resume waypoint movement (10 seconds).


killerwife:
Ok I can safely say that this is smth we need because I know about this. Getting data for it will be, painful :D I will merge this soon.

Nytanath:
Got it, I'll keep it in mind for the next time, thanks!

I don't think it can vary between different spawns of the same creature. Would it be better to have it in creature_template then? I only added it in addon because it was something extra that not all creatures care about.

Nytanath:
Moved it to creature_template since it does not vary by spawn (at least that I know of - but I can't think of any reason why it should).

killerwife:
Yeah, if it doesnt vary by spawn, addon is unnecessary. We kinda get away with creature_template being so humongous cos we dont need to insert it often. I am all for it being moved to addon should we find that its per-spawn.

al3xc1985:
`When you interact with a creature that is moving through a waypoint path, that creature pauses its movement for a certain amount of time. The most common value is 3 minutes, but that can vary for certain creatures. For example, I have confirmed that ["Cookie" McWeaksauce](https://tbc.wowhead.com/npc=17246/cookie-mcweaksauce) pauses for three seconds, and [Jinar'Zillen](https://tbc.wowhead.com/npc=11317/jinarzillen) pauses for 30 seconds.`

It might depend on npc/mob class the wait time... Cookie for example is a trainer also, the other one is only a normal mob.
Also merchants might have full 3 min wait time.

So than we can do I simple sql update and adjust all of them depending on the npc type

al3xc1985:
I remembr Inspector razuvious in dk start zone not stand in place for 1 sec to talk with ya... I believe he has 30 sec

This is my pu

Nytanath:
I wish it was that easy, but from what I've seen so far specific creatures have "special" timers while everyone else has the default one.

Here's the data I have so far:
- [Private Thorsen](https://classic.wowhead.com/npc=738/private-thorsen) (he has the gossip flag while he runs for his out of combat event): 0 ms (no pause).
- [Jinar'Zillen](https://tbc.wowhead.com/npc=11317/jinarzillen): 30000 ms
- ["Cookie" McWeaksauce](https://tbc.wowhead.com/npc=17246/cookie-mcweaksauce): 3000 ms (TBC+)

al3xc1985:
so we can't have a logic with what all these guys have in commune or not? 
Why I am asking this? cause I was wondering if we might find 2-3 similar examples from all those categories that u laied down, and have a calculation, cause otherwise we have some hard time collecting data like this

Nytanath:
> so we can't have a logic with what all these guys have in commune or not?

The examples we have so far have nothing in common, if we find more then sure we could find some common point between them, but consider that this is a rare behavior so there aren't many creatures that do this. I think it was made so that players couldn't stop a npc for 3 minutes each time while they were going to do some event.

AnonXS:
John Mithril [Entry 14508] e.g the guy that starts gurubashi arena has to have 0 too as he does not remove gossip flags in sniff, so i snuck that in to make sure the start time of gurubashi event can not be delayed.

Nytanath:
Added to the list, thanks!

PR #478, opened 2022-03-02
Implement command .debug areatriggers to make areatrigger scripting easier by informing gamemasters when they enter an areatrigger

Nytanath:
## 🍰 Pullrequest

Taken from TrinityCore, so all credits to them.

The command .debug areatriggers is a toggle that starts off. When it's activated, whenever the GM enters an areatrigger it will send a chat message containing the areatrigger's id. This makes identifying areatriggers far easier.
To disable it, use the command again.

### Proof

- None

### Issues

- None

### How2Test

- .debug area
- .tele zul'gurub
- Enter the raid.
- Walk forward: there's an areatrigger just in front of the entrance, before the stairs.
- Once you enter the areatrigger, you will see a chat message.
- Type .debug area to disable it.


killerwife:
I suggest two more features if you can be bothered. List all areatriggers in a zone (or map if in instance) and tp to an areatrigger.

Nytanath:
There is already a teleport to areatrigger command (.go trigger <id>), I'll see if I can get a list areatrigger command working.

PR #480, opened 2022-03-03
Implement command .tele warp to teleport the user along a specified axis by a specified value

Nytanath:
## 🍰 Pullrequest

The command .tele warp teleports the user along a specified axis by a specified amount. This allows forward/backward/left/right/above/below movement, as well as changing the current orientation by a certain amount of degrees.
Currently it is impossible for a gamemaster to teleport below the terrain level of a map; with this command, it is as easy as typing ".tele warp z -5" (not really useful for Vanilla as you can't use .gm fly there, so you fall until you get ported to the nearest graveyard, but good for TBC and WotLK).
Want to see what lies beyond an instance portal? ".tele warp x 20" will allow you to teleport beyond the instance portal.

Credit for the involved math to [this forum post](https://www.getmangos.eu/forums/topic/5918-solved-warp-command-in-all-directions-this-damn-math-caclulations/#comment-58246) by [lillecarl](https://www.getmangos.eu/profile/2666-lillecarl/).

### Proof

- None

### Issues

- None

### How2Test

- Forward: .tele warp x 10
- Backward: .tele warp x -10
- Right: .tele warp y 10
- Left: .tele warp y -10
- Above: .tele warp z 10
- Below: .tele warp z -10
- Face opposite direction: .tele warp o 180
- 360 no scope!: .tele warp o 360


insunaa:
Personally I think this is a good addition, tho I would put it into the `.go` command instead of the tele command. I believe it's grouped better, there

Nytanath:
Makes sense, thanks! I'll change it.

PR #485, opened 2022-05-29
Port group update flags from vmangos

celguar:
Fixes group update flags being a copy paste from tbc
Also fixes debuffs in party/raid default UI in 1.12
I did it because I can't use Hermes Proxy + 1.14 otherwise


killerwife:
Verified with client, good find.

PR #487, opened 2022-06-19
Fix Scourge Invasion mail spam

celguar:
Fixes Scourge invasion letters being sent on every server start

### Issues
- Event mail with quest item is sent on every server start if scourge invasion is enabled. E.g. start server 30 times = get 30 letters to every character.

### How2Test
- Start Scourge Invasion (.worldstate scourge state 1), you receive 1 letter with quest item, restart server. You now have 2 identical letters (only first one will have item though).


celguar:
renamed argument to be more clear

PR #491, opened 2022-07-29
Use #/bin/bash instead of #/bin/sh

Thorsten42:
Use bash in shebang since double brackets used in this script are bash extensions and for example line 147 is causing the following error with sh "./ExtractResources.sh: 147: [[: not found"  (MoveMapGen.sh is also using  #/bin/bash)

### DIFF
--- a/contrib/extractor_scripts/ExtractResources.sh
+++ b/contrib/extractor_scripts/ExtractResources.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


insunaa:
The shebang was changed to /bin/bash deliberately to keep POSIX compliant. Unfortunately code parts that are not POSIX compliant were added which cause the issues you describe.
The proper solution would be to fix the code segments that are not POSIX compliant instead of adjusting the shebang.

This warrants more discussion imo.

PR #493, opened 2022-10-11
Set Autoload.Active to False by default

jrtashjian:
## 🍰 Pullrequest
Currently Autoload.Active is set to True by default even though it's noted to be "useful for running in debug mode." Because of this I think it makes sense to disable it by default. Disabling this reduces the cpu and memory usage significantly preventing a memory issue reported in https://github.com/cmangos/issues/issues/3140.

### Proof
- https://github.com/cmangos/issues/issues/3140#issuecomment-1264133822
- https://github.com/cmangos/issues/issues/3140#issuecomment-1272656983

### Issues
- Relates to https://github.com/cmangos/issues/issues/3140


Grz3s:
I am not recommending this change:
Explained here:
https://github.com/cmangos/issues/issues/3140

killerwife:
This breaks scripts that assume activity of a given creature or gameobject. For example, completely breaks bashir landing if noone is in blades edge. So while this is perhaps sensible for you, this is not valid for a proper server.

PR #494, opened 2022-10-17
Fix mangos.sql import error

ellingtonsantos:
Fix error below on mangos.sql import
ERROR 1136 (21S01) at line 325: Column count doesn't match value count at row 1

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


ellingtonsantos:
It is related to issue #3171.

olegabr:
had the same issue with the same fix

PR #500, opened 2023-05-21
Prevent using dangling pointer to HonorStanding.

mostlikely4r:
## 🍰 Pullrequest
GetHonorStandingByGUID and GetHonorStandingByPosition return pointers to HonorStanding which are dereferenced inside the methods. This pull-request makes these methods return a copy of the object instead.

### Proof

Here (https://github.com/cmangos/mangos-classic/blob/b2ac38debda59413b680575b26e57df2b9b77638/src/game/Globals/ObjectMgr.cpp#L3690) GetInfo() returns a pointer to itself which is a HonorStanding contained in the list defined inside the method.

### DIFF
--- a/src/game/Globals/ObjectMgr.cpp
+++ b/src/game/Globals/ObjectMgr.cpp
@@ -3681,19 +3681,19 @@ HonorStandingList ObjectMgr::GetStandingListBySide(uint32 side)
     }
 }
 
-HonorStanding* ObjectMgr::GetHonorStandingByGUID(uint32 guid, uint32 side)
+HonorStanding ObjectMgr::GetHonorStandingByGUID(uint32 guid, uint32 side)
 {
     HonorStandingList standingList = sObjectMgr.GetStandingListBySide(side);
 
     for (HonorStandingList::iterator itr = standingList.begin(); itr != standingList.end() ; ++itr)
         if (itr->guid == guid)
-            return itr->GetInfo();
+            return *itr;
 
-    return 0;
+    return HonorStanding();
 }
 
 
-HonorStanding* ObjectMgr::GetHonorStandingByPosition(uint32 position, uint32 side)
+HonorStanding ObjectMgr::GetHonorStandingByPosition(uint32 position, uint32 side)
 {
     HonorStandingList standingList = sObjectMgr.GetStandingListBySide(side);
     uint32 pos = 1;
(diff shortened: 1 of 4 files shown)


src/game/Globals/ObjectMgr.h
```diff
@@ -421,6 +421,11 @@ class HonorStanding
         {
             return honorPoints > rhs.honorPoints;
         }
+	
+        operator bool() const
```
evil-at-wow, line 425: Typically a good practice is to make conversions like this ```explicit``` so you don't get conversions "by accident" to say so. They would still be "contextually converted to bool" so you could continue to use this in if, while, for, built-in logical operators, and so on.

src/game/Globals/ObjectMgr.cpp
```diff
@@ -3681,31 +3681,31 @@ HonorStandingList ObjectMgr::GetStandingListBySide(uint32 side)
...
     for (HonorStandingList::iterator itr = standingList.begin(); itr != standingList.end() ; ++itr)
     {
         if (pos == position)
-            return itr->GetInfo();
+            return *itr;
         pos++;
     }
 
-    return 0;
+    return HonorStanding();
```
evil-at-wow, line 3708: Given the semantics of these functions, that are not guaranteed to give a result, maybe returning an ```std::optional<HonorStanding>``` would make sense in these cases? 

Given the already existing default constructor of HonorStanding, which constructs some kind of nothing-object, his is acceptable for me as well, but I'd personally prefer an API that makes it clear that "this may not return what you were looking for".

evil-at-wow:
Good catch!

src/game/Globals/ObjectMgr.cpp
mostlikely4r: Thank you. I did not know about std::optional and it looks like I can use that in more situations in the future!

mostlikely4r:
I modified the change to use std::optional and removed the implicit bool conversion. Does this solution look better?

src/game/Tools/Formulas.h
```diff
@@ -151,13 +151,13 @@ namespace MaNGOS
             for (uint8 i = 1; i <= 13; i++)
             {
                 honor = 0.0f;
-                tempSt = sObjectMgr.GetHonorStandingByPosition(sc.BRK[i], team);
+                std::optional<HonorStanding> tempSt = sObjectMgr.GetHonorStandingByPosition(sc.BRK[i], team);
                 if (tempSt)
                 {
-                    honor += tempSt->honorPoints;
+                    honor += tempSt.value().honorPoints;
```
killerwife, line 157: pretty sure std::optional has a -> operator. Or was there a reason for not using it.

mostlikely4r: Good catch. Using the -> operator is beter here. I amended the PR.

killerwife: Mhm, for the future, squash a small pr like this into a single commit. I will do it myself and merge it, this was my only remark.

PR #501, opened 2023-05-23
fix multithread access segmentation fault

cdkr:
## 🍰 Pullrequest
fix segmentation fault caused by multithread access m_objects in CreatureGroup::Despawn

### Proof
Screenshot from 2023-05-23 16-45-05

### Issues
I found my mangosd crashes every day at 12:00 a.m. After added some log I found these lines of code called simultaneously.
```
void SpawnGroup::RemoveObject(WorldObject* wo)
{
    m_objects.erase(wo->GetDbGuid());
    // ...
}

void CreatureGroup::Despawn(uint32 timeMSToDespawn, bool onlyAlive, uint32 forcedDespawnTime)
{
    //...
    for (auto objItr : m_objects)   // segmentation fault here
    {
    }
}
```

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Maps/SpawnGroup.cpp
+++ b/src/game/Maps/SpawnGroup.cpp
@@ -487,7 +487,9 @@ void CreatureGroup::MoveHome()
 void CreatureGroup::Despawn(uint32 timeMSToDespawn, bool onlyAlive, uint32 forcedDespawnTime)
 {
     time_t when = time(nullptr) + forcedDespawnTime;
-    for (auto objItr : m_objects)
+    auto objects = m_objects;
+    
+    for (auto objItr : objects)
     {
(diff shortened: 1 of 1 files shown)


killerwife:
Not multithread access, access in same stack in same thread. However a valid remark nonetheless

killerwife:
And identical issue in GO group

evil-at-wow:
That's a nice find, and also very nice timing, because seeing that call stack I believe this to be the same issue as reported in cmangos/issues#3410. Which for some unknown reason I wasn't able to reproduce yet in my TBC core (despite also working on Linux), but this surely avoids I'll have to spend more time on that :+1:

PR #506, opened 2023-06-07
Delete QueryResult* for _LoadForgottenSkills

mostlikely4r:
## 🍰 Pullrequest
Tiny memory leak for forgotten skills on player login.

### Proof
Forgotten skills uses a QueryResult* as parameter here: https://github.com/cmangos/mangos-classic/blob/8a1666aa8880d61980cc600d40fd7ff4f6b6b30d/src/game/Entities/Player.cpp#L13898

However this QueryResult* is never deleted leading to a minor memory leak.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -13924,6 +13924,8 @@ void Player::_LoadForgottenSkills(QueryResult* result)
 
             m_forgottenSkills.insert(std::make_pair(skill, value));
         } while (result->NextRow());
+	    
+        delete result;
     }
 }
 


killerwife:
The better solution here is to initialize a unique_ptr with  the query result that does it for us. I have been converting it that way since it more futureproof foolproof solution.

evil-at-wow:
I wasn't aware of this PR as I usually do stuff in the TBC core. Recently I've done a bunch of changes that got rid of most of the manual memory management for database query results, and because these changes touched so many files I also ported my commits to the Classic and WoTLK cores. As part of that process, I also noticed this memory leak as a side effect - see my commit cmangos/mangos-classic@b634d2649183124fec2b7d6af96ad8f01b0b82c1.

Nevertheless, thank you for bringing this to our attention, and my apologies for having made your work obsolete!

PR #507, opened 2023-06-10
missing break in Spell.cpp

celguar:
## 🍰 Pullrequest
Adds missing break;

### Proof
In wotlk is present

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -2003,6 +2003,7 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, bool targ
                 }
             }
             m_targets.setDestination(nextPos.x, nextPos.y, nextPos.z);
+	    break;
         }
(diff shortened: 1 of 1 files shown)


insunaa:
Does `TARGET_LOCATION_CASTER_FRONT_LEAP` even exist in 1.12? I thought leaping was a WotLK only thing. Because if it doesn't this code path will never be hit and so the break is pointless either way

Gamemechanicwow:
> Does TARGET_LOCATION_CASTER_FRONT_LEAP even exist in 1.12? 

Yes. It's used for blink.

celguar:
Frankly this is not critical. And blink even works. I only changed it because my ike3 bots would cast blink as self target (so spell had unitTarget, so w/o break code would go lower and bots would teleport to themselves. I fixed it in bots code, but this is where I found it by debugging, so figured I'd make a PR, tbc also doesn't have break here, only wotlk. Dunno if it's intentional

PR #513, opened 2023-07-04
Auth: Port VMaNGOS implementation of TOTP for 1.12

insunaa:
## 🍰 Pullrequest

This PR prevents realmd from requiring a valid token response from the client if the client is older than 1.12.3, as it was only implemented in 2.4.3. Checking specifically for 1.12.3 and not for 2.4.3 because a TBC server *may* possibly allow other builds than just 2.4.3 and it shouldn't be possible to use these to circumvent a token that would otherwise be enforced with 2.4.3

### DIFF
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -566,10 +566,10 @@ bool AuthSocket::_HandleLogonProof()
 
     bool pinResult = true;
 
-    if (m_promptPin && lp.securityFlags & SECURITY_FLAG_PIN)
+    if (m_promptPin && (lp.securityFlags & SECURITY_FLAG_PIN))
         pinResult = false;
 
-    if (m_promptPin && lp.securityFlags & SECURITY_FLAG_PIN)
+    if (m_promptPin && (lp.securityFlags & SECURITY_FLAG_PIN))
     {
         auto pin = generateToken(_token.c_str());
 
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -31,6 +31,7 @@
 #include "AuthCodes.h"
 #include "Auth/SRP6.h"
 #include "Util/CommonDefines.h"
+#include "Util/Util.h"
 
 #include <openssl/md5.h>
 #include <ctime>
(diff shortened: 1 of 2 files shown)


killerwife:
Vanilla client has a keypad that we are using for token. Was not aware it was broken. This keypad is then not present in tbc.

insunaa:
How do the implementations differ? Because if the token field is filled with a token and a vanilla client tries to log in, it just fails with a connection error.

insunaa:
@killerwife I added the VMaNGOS code for TOTP authentication using the PIN keypad (tho I disabled the random scrambling of it. it's hard enough as it is to type 6 digits on that keypad within 30 seconds, to do it with a scrambled keypad is a huge pain)

How do I properly credit VMaNGOS?

src/realmd/AuthSocket.cpp
```diff
@@ -473,13 +474,20 @@ bool AuthSocket::_HandleLogonChallenge()
...
+                        securityFlags = SECURITY_FLAG_PIN;
+
                     pkt << uint8(securityFlags);                    // security flags (0x0...0x04)
 
                     if (securityFlags & SECURITY_FLAG_PIN)          // PIN input
                     {
-                        pkt << uint32(0);
-                        pkt << uint64(0);
-                        pkt << uint64(0);
+                        uint32 gridSeedPkt = m_gridSeed = static_cast<uint32>(0/*urand()*/);
```
insunaa, line 484: Leaving this as 0 instead of urand() (neither of which really need a static_cast, but the compiler will optimize that out anyway) means that the keypad on the client is not scrambled. Adding any other number (or urand()) here will scramble the keypad, making text entry more difficult for no real benefit. It's not a physical keypad where an attacker could look at your finger movements to determine the code you're entering, after all. If an attacker can see your screen, you have bigger problems.

Left it in in case someone sees the need to enable scrambling, tho.

insunaa:
Any progress on this?

src/realmd/AuthSocket.cpp
vcorbuleanu: It would be interesting to check which value makes the keypad have the buttons in order from top-left to bottom-right (the way old phones used to have them).

killerwife:
Other one was merged so closing this

PR #520, opened 2023-10-29
Refactor PetAI

deiteris:
## 🍰 Pullrequest

Refactors PetAI logic to be more concise and possible a bit more optimized:

* Remove PetAI::UpdateAllies() and timer and iterate over the owner's group directly.
* Remove unnecessary duplicate check for disabled actions in PetAI::UpdateAI()
* Convert and store Creature pointer as a pet in PetAI to simplify checks.
* Create Pet::HasActionsDisabled() and refactor checks for disabled pet mode
* Refactor spell pick and cast into separate PetAI::Cast() and PetAI::PickSpellWithTarget() functions.
* Refactor autocast spell picking logic and order.
* Use a single spell-target pair instead of a vector storage and random choice.

### Proof
- None

### Issues
- None

### How2Test
- None

### Regression test checklist
- [X] Pet autocasts combat spells and buffs in combat.
- [X] Pet autocasts only non-combat spells outside combat.
- [X] Pet buffs party members.
- [X] Pet automatically attacks an enemy in line of sight.
- [X] Pet returns to follow state after combat.
- [X] Pet returns to its stay position after combat.

### DIFF
--- a/src/game/AI/BaseAI/PetAI.cpp
+++ b/src/game/AI/BaseAI/PetAI.cpp
@@ -35,24 +35,19 @@ int PetAI::Permissible(const Creature* creature)
     return PERMIT_BASE_NO;
 }
 
-PetAI::PetAI(Creature* creature) : CreatureAI(creature), inCombat(false)
+PetAI::PetAI(Creature* creature) : CreatureAI(creature), m_inCombat(false)
 {
     m_followAngle = M_PI_F / 2;
     m_followDist = 1.5f;
-    m_AllySet.clear();
-    UpdateAllies();
+    m_pet = creature->IsPet() ? static_cast<Pet*>(creature) : nullptr;
 
-    if (creature->IsPet())
+    if (m_pet && m_pet->isControlled())
     {
-        Pet* pet = dynamic_cast<Pet*>(creature);
-        if (pet->isControlled())
-        {
-            if (sWorld.getConfig(CONFIG_BOOL_PET_ATTACK_FROM_BEHIND))
-                m_attackAngle = M_PI_F;
+        if (sWorld.getConfig(CONFIG_BOOL_PET_ATTACK_FROM_BEHIND))
+            m_attackAngle = M_PI_F;
 
-            if (Unit* owner = m_unit->GetOwner())
-                RequestFollow(owner);
-        }
+        if (Unit* owner = m_unit->GetOwner())
+            RequestFollow(owner);
     }
 
     switch (creature->GetUInt32Value(UNIT_CREATED_BY_SPELL))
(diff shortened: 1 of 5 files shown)


killerwife:
Did you make sure a pet breaks attack off when a target was cced after attack order was issues?

deiteris:
Aha! So that additional check for victim was actually to support this behavior. I noticed that pet stops attacking polymorphed creature, but then ignores attack order.

Though I see that pet still attacks feared target, even though this a CC that we might not want to break. Is this expected? 🤔

insunaa:
It's common for warlocks to fear-juggle enemies. If their pets would stop attacking with every fear that'd be a significant DPS loss, so I'm pretty sure the only CCs that should stop pet attacks are those that break with any damage, such as sap, sheep or gouge

killerwife:
The pet needs to ignore damage breakable cc that occurs after attack order. You can always tell pet to attack after the fact and it needs to comply.

killerwife:
Lookup the commits adding it to see the rationale using git blame.

deiteris:
> The pet needs to ignore damage breakable cc that occurs after attack order. You can always tell pet to attack after the fact and it needs to comply.

Yes, I understand and I will fix this. I'm trying to identify the culprit of this issue since I see it occasionally happening even with the current code (and the AI logic does not seem to handle this in any way, it *seems* to me as a timing issue or improper state). The attack order may still be ignored sometimes after the target was polymorphed and only target switching helps.

> It's common for warlocks to fear-juggle enemies. If their pets would stop attacking with every fear that'd be a significant DPS loss, so I'm pretty sure the only CCs that should stop pet attacks are those that break with any damage, such as sap, sheep or gouge

I don't object this kind of logic, but I'd like to understand whether this is a blizzlike behavior. Another example, pets also don't stop attacking enemies affected by frost nova, but this is a CC that breaks immediately.

> For example they don't always cast sprint on attack order, even if it's off cd, but then occasionally in melee combat they cast sprint.

I think this might be an issue with updated logic too (it most likely will always autocast it), but I'm not sure how to check. Can I somehow make pet learn spells?

deiteris:
I think I've sorted the issues out. Pets stop attacking CCed targets correctly in this PR now, and there were also a few fixes added to different behaviors:
* Pets kept switching targets when commanded to retreat, causing them to stop attacking until they're in the owner's following distance.
* Pets could target CCed targets in aggressive mode after you command attack and cancel attack with follow.

killerwife:
Ok and now one last thing to bother you with

PetAI is notoriously known to be ass to script.

Something like this:

https://github.com/cmangos/mangos-tbc/blob/master/src/game/AI/ScriptDevAI/scripts/outland/hellfire_peninsula.cpp#L2231

When you start a script, nothing that a player can do, short of abandon or unsummon should be able to interrupt such a sequence. I understand its a tbc script but I am sure you could write a similar thing in vanilla where a pet moves a bit, does some emote and wont have its script broken.

deiteris:
> When you start a script, nothing that a player can do, short of abandon or unsummon should be able to interrupt such a sequence.

Done. I've tested on TBC and scripted pet at least doesn't get stuck anymore. Though it's not ideal since this is a charm, not an actual pet, so aggressive mode doesn't work correctly for example. But this a different story that needs to be looked into.

PR #522, opened 2023-11-25
OutdoorPvP - using conf variables to enable/disable it

Hectelion:
## 🍰 Pullrequest
These changes are designed to allow you to enable or disable OutdoorPvP in the Eastern Plaguelands and Silithus via the appropriate variables that were already present in mangosd.conf

### Proof
- None

### Issues
- None

### How2Test
- Test1: Set in your mangosd.conf the OutdoorPvp.SIEnabled = 0 go Silithus and you should not see the AlwaysUpFrame* that show silithyst resource gathered by faction
- Test2: Set in your mangosd.conf the OutdoorPvp.EPEnabled = 0 go to Eastern Plaguelands and you should not see the AlwaysUpFrame* that show towers conquered by faction

### Todo / Checklist
I don't understand why have been placed 2 variables while the outdoorpvp in these zones was released without distinction
I am trying to set a sort of progression starting .
Outdoor pvp in Silithus and Eastern Plaguelands areas is one of the latest things added (1.12) and being able to disable and enable it is a small step towards introducing a progression


insunaa:
I don't think that this is how you should do it.
You're just enclosing everything in if statements that has anything to do with outdoor pvp, which is just pointless.

You can just disable the worldstates

insunaa:
Yeah, this is not mergeable. Almost everything you do here is entirely redundant because it's already handled by the regular code.
The only difference there *may* be, (I haven't tested on Classic, only WotLK) is that your code may or may not disable the worldstate overlay.
If it does that, go through your code again and isolate what part of your code is responsible for it, because everything else is already handled by the code before your patch.

Hectelion:
I have limited the changes to only the points where I think they are necessary

src/game/Entities/Player.cpp
```diff
@@ -7677,12 +7683,22 @@ void Player::SendInitWorldStates(uint32 zoneid) const
...
                 outdoorPvP->FillInitialWorldStates(data, count);
             break;
-    }
+        }
+            
+    case 1377:                                      // Silithus
+        if (sWorld.getConfig(CONFIG_BOOL_OUTDOORPVP_SI_ENABLED))
+        {
+            if (OutdoorPvP* outdoorPvP = sOutdoorPvPMgr.GetScript(zoneid))
+                outdoorPvP->FillInitialWorldStates(data, count);
```
insunaa, line 7698: This shouldn't be necessary because the config option would not load a script for the zoneid, so GetScript(zoneid) returns nullptr for every zone that's disabled by config

src/game/OutdoorPvP/OutdoorPvPEP.cpp
```diff
@@ -43,11 +43,14 @@ OutdoorPvPEP::OutdoorPvPEP() : OutdoorPvP(),
 
 void OutdoorPvPEP::FillInitialWorldStates(WorldPacket& data, uint32& count)
```
insunaa, line 44: same here, the OutdoorPVPEP class should never get an object because it's disabled through config

insunaa:
try it with this please
```cpp
OutdoorPvPMgr::OutdoorPvPMgr()
{
    m_updateTimer.SetInterval(TIMER_OPVP_MGR_UPDATE);
    for (auto* script : m_scripts)
        script = nullptr;
//    memset(&m_scripts, 0, sizeof(m_scripts));
}
```

I can't try myself because I'm currently doing something else and can't stash the current work

Hectelion:
I tried leaving the changes only in the Player.cpp in the Player::UpdateZone
in the OutdoorPvPEP in the OutdoorPvPEP::ProcessCaptureEvent and in the OutdoorPvPSI in the OutdoorPvPSI::HandleAreaTrigger and it seems to work.
Now I'll try to remove these too and make the change you suggested and I'll let you know. Thank you

insunaa:
I forgot, the change also needs this to work fully

```cpp
OutdoorPvPMgr::~OutdoorPvPMgr()
{
    for (auto& m_script : m_scripts)
        if (m_script)
            delete m_script;
}
```
it should work without that, but then crash the server on shutdown if this is missing

Hectelion:
I tried with these:
OutdoorPvPMgr::OutdoorPvPMgr()
{
    m_updateTimer.SetInterval(TIMER_OPVP_MGR_UPDATE);
    for (auto* script : m_scripts)
        script = nullptr;
    // memset(&m_scripts, 0, sizeof(m_scripts));
}
OutdoorPvPMgr::~OutdoorPvPMgr()
{
    for (auto* m_script : m_scripts)
        if (m_script)
            delete m_script;
}

but something strange happens, with the outdoorpvp variables turned on it starts with the variables turned off the server doesn't start it seems to stop when loading the maps

insunaa:
I could only test this very poorly on WotLK for reasons.
With these changes the server starts completely fine for me and teleporting to eastern plaguelands correctly doesn't have any active towers and doesn't show them on the map, but the towers do still show in the overlay at the top, and I'm not sure why, yet.

Hectelion:
Actually I'm testing with the previews changes at Player, OutdoorPvPEP, and OutdoorPvPSI, with both on and both off it works perfectly,. with one on and one off instead seems to be more unpredictable, even could carry the icons from a pvp zone that is on to another the other that is off.... but I think this is a limit case because normally they should be both on or off simultaniusly

src/game/Entities/Player.cpp
```diff
@@ -7677,12 +7683,22 @@ void Player::SendInitWorldStates(uint32 zoneid) const
 
     switch (zoneid)
     {
-        case 139:                                       // Eastern Plaguelands
-        case 1377:                                      // Silithus
+    case 139:                                       // Eastern Plaguelands
+        if (sWorld.getConfig(CONFIG_BOOL_OUTDOORPVP_EP_ENABLED))
+        {
             if (OutdoorPvP* outdoorPvP = sOutdoorPvPMgr.GetScript(zoneid))
```
insunaa, line 7689:
```cpp
if (OutdoorPvP* outdoorPvP = sOutdoorPvPMgr.GetScript(zoneid))
{
sLog.outError("%s in: %s at: %d", __PRETTY_FUNCTION__, __FILE__, __LINE__);
outdoorPvP->FillInitialWorldStates(data, count);
```
as an example

Hectelion (commented):
Removed the useless changes, 
Outdoor with a single active area would need to be corrected, but I think it's out of interest

insunaa:
it's impossible for me to reproduce the effect of the icons not appearing in WotLK, they always do, no matter what I do, so I don't know where it might be on classic

Hectelion:
I'm closing it because I have some ko tests. Need to be tested better. I hope to be able to do it again soon. Thanks for your help, you have simplified the points for me to see

PR #528, opened 2024-01-21
Mangosbot integration

davidonete:
## 🍰 Pullrequest
Mangosbot system integration to cmangos core

The system lives in a separate repository and will be fetched only if cmake flags are enabled.

Mangosbots repository: https://github.com/celguar/mangosbot-bots


src/game/Entities/Player.cpp
```diff
@@ -670,16 +676,16 @@ Player::~Player()
     for (auto& itr : m_boundInstances)
             itr.second.state->RemovePlayer(this);
 
-#ifdef BUILD_PLAYERBOT
+#if defined(BUILD_PLAYERBOT) || defined(ENABLE_MANGOSBOTS)
     if (m_playerbotAI)
     {
         delete m_playerbotAI;
```
killerwife, line 682: Use unique_ptr

PR #530, opened 2024-02-06
Clang: Fix compiler warnings

insunaa:
## 🍰 Pullrequest

See https://github.com/cmangos/mangos-wotlk/pull/493

I'd again appreciate a second pair of eyes going over this real quick :)

### DIFF
--- a/src/game/AI/BaseAI/CreatureAI.h
+++ b/src/game/AI/BaseAI/CreatureAI.h
@@ -52,7 +52,7 @@ class CreatureAI : public UnitAI
 
         void AddUnreachabilityCheck(); // use in constructor
 
-        CreatureSpellList const& GetSpellList() const;
+        CreatureSpellList const& GetSpellList() const override;
 
         void TimedFleeingEnded() override;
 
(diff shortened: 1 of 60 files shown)


evil-at-wow (approved):
This looks consistent with the previous ones you did in WotLK/TBC, and I didn't see anything unexpected with a quick glance.

The GridMap change is the same as in the TBC core and expected here as well. 

The OutdoorPVP ones were new, but again totally expected with ```SetGraveYardLinkTeam``` taking an ```uint32``` as last argument (makes one wonder who wrote ```{ 0 }``` in the first place, but it turns out the call was changed years ago to take a ```std::vector<uint32>``` when that was done and then a few days later the call was changed back to ```uint32``` with apparently no compilers complaining hard enough, which isn't really surprising given list initialization).

insunaa:
@evil-at-wow That is pretty interesting, yeah. I figured it must've been something like that, because why else use a list initializer for a single uint32 :D
Thank you again for reviewing the PR :heart:

PR #547, opened 2024-04-05
Ruins of Ahn'Quiraj memory leak fix

Vigerus:
## 🍰 Pullrequest
With each Ossirian attempt the instance map is filled with new spawned crystals, that never get removed from there (only if we instance is unloaded i guess)

### Proof
ticket on discord

### Issues
- fixes #Ruins of Ahnquiraj memory leak

### How2Test
- Start the instance presumably with bots, go to ossirian, have couple tries on him and just attach with VS and watch the m_goEntryGuidCollection of instance_ruins_of_ahnqiraj

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/ruins_of_ahnqiraj/ruins_of_ahnqiraj.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/ruins_of_ahnqiraj/ruins_of_ahnqiraj.cpp
@@ -84,6 +84,17 @@ void instance_ruins_of_ahnqiraj::OnObjectCreate(GameObject* go)
         m_goEntryGuidCollection[go->GetEntry()].push_back(go->GetObjectGuid());
 }
 
+void instance_ruins_of_ahnqiraj::OnObjectDestroy(GameObject* go)
+{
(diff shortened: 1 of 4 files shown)


insunaa:
I'm not sure your code does what you think it does.

m_goEntryGuidCollection may be growing larger because of dynguids, but it's only a collection of ObjectGuids, so even if the collection is only cleared on instance unload it shouldn't matter.
In the non-dynguid case the newly spawned GObs inherit the ObjectGuid of the GOb that was in the same place before this, unless of course they really are wild GObs. In that case however you should still not remove every single crystal from the collection upon object destruction, you should at most remove the currently destroyed object, I think.

Vigerus:
> I'm not sure your code does what you think it does.
> 
> m_goEntryGuidCollection may be growing larger because of dynguids, but it's only a collection of ObjectGuids, so even if the collection is only cleared on instance unload it shouldn't matter. In the non-dynguid case the newly spawned GObs inherit the ObjectGuid of the GOb that was in the same place before this, unless of course they really are wild GObs. In that case however you should still not remove every single crystal from the collection upon object destruction, you should at most remove the currently destroyed object, I think.

Yeah, my bad, it should be a double search, update inc.

PR #551, opened 2024-04-11
Script Rework for Quest 7631 - Dreadsteed of Xoroth (Warlock Mount Quest)

Sneakyabc:
Quest was not functioning properly.
- 3 Summoned objects should no longer bug out the first time you activate the quest, instantly resetting
- Waves of demons spawning should be closer to how it should be (more random and consistent, not big waves all at once)
- 3 Objects now properly deactivate and need to be clicked by the warlock to reactivate
- Demons should no longer float through the air awkwardly
- Adjusted timers a little bit
- Changed Unsummon dialog to new, previously missing entry

The check to ensure the dreadsteed event is in progress, and if not to despawn the 3 quest GOs is probably not the best way to do that. After trying several other things I could not get them to stop spawning in by default, which causes problems.

I mostly watched this video as my reference:
youtube.com/watch?v=a5eAYFIJj68

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/dire_maul/dire_maul.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/dire_maul/dire_maul.cpp
@@ -90,6 +90,8 @@ enum
     SUMMON_WAVES_RELAY_SCRIPT_START_ID = 1450100,
 };
 
+static const float pedestalCords[3] = { -38.5f, 811.8f, -29.5f};
+
 struct npc_warlock_mount_ritualAI : public  Scripted_NoMovementAI
 {
     npc_warlock_mount_ritualAI(Creature* pCreature) : Scripted_NoMovementAI(pCreature)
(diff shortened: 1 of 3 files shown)


src/game/AI/ScriptDevAI/scripts/kalimdor/dire_maul/dire_maul.cpp
```diff
@@ -118,20 +122,48 @@ struct npc_warlock_mount_ritualAI : public  Scripted_NoMovementAI
     void JustSummoned(Creature* pSummoned) override
     {
         m_luiSummonedMobGUIDs.push_back(pSummoned->GetObjectGuid());
+        // The spawning of the waves is handled in the DB relay scripts, the movement to the center I'm doing here because simply using a
```
killerwife, line 125: Generally all waves should be spawn groups with their defined points and each wave be a new one instead of doing this. However, I am not hellbent on someone reworking that works and doesnt clash with any overarching architecture.

src/game/AI/ScriptDevAI/scripts/kalimdor/dire_maul/dire_maul.cpp
```diff
@@ -170,6 +235,13 @@ struct DreadsteedQuestObjects : public GameObjectAI
 
     void JustSpawned() override
     {
+        if (instance_dire_maul* instance = (instance_dire_maul*)m_go->GetInstanceData()) // Without this check, the 3 objects are spawning when first loaded in. Couldn't find another way to stop them from doing this
```
killerwife, line 238: Old spawning system, if migrated to spawn groups OR spawnmask 0, this issue doesnt occur. One of the reasons I had to move away from how it used to work.

PR #563, opened 2024-07-05
Implement sAuctionHouseBot.IsInitialized

mostlikely4r:
## 🍰 Pullrequest
IsInitialized() implementation for AuctionhouseBot to allow external modules (such as playerbot) to detect if GetItemData() can be used safely. 

### Proof
When AuctionHouseBot.Chance.Sell and  AuctionHouseBot.Chance.Buy are both set to 0 in the config GetItemData() will return uninitialized values. This causes issues in for example playerbot code.

### How2Test
Disable Auctionhousebot in the config with playerbots enabled.

### Todo / Checklist
- [ ] Use IsInitialized in botcode.

### DIFF
--- a/src/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -30,7 +30,7 @@
 
 INSTANTIATE_SINGLETON_1(AuctionHouseBot);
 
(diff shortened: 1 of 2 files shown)


mostlikely4r:
Added the option to reload the config and have IsInitialized reflecting the current settings.

killerwife:
Would appreciate @insunaa or someone who might use these things to validate this. I am as usual fine with merging this but bots are not in my purview.

mostlikely4r:
No longer needed with upcoming pr.

PR #567, opened 2024-07-10
WorldSession: Remove unimplemented method declarations

Exxenoz:
## 🍰 Pullrequest
Remove unimplemented method declarations from WorldSession

### DIFF
--- a/src/game/Server/WorldSession.h
+++ b/src/game/Server/WorldSession.h
@@ -244,7 +244,6 @@ class WorldSession
         std::string const& GetAccountName() const { return m_accountName; }
         Player* GetPlayer() const { return _player; }
         char const* GetPlayerName() const;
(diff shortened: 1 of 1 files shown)


insunaa:
Most of these are implemented in WotLK (and some possibly in TBC)
They shouldn't be called in vanilla but they may make cherry-picking easier. Thoughts?

Exxenoz:
> Most of these are implemented in WotLK (and some possibly in TBC) They shouldn't be called in vanilla but they may make cherry-picking easier. Thoughts?

I've thought about that too, but in my opinion, it doesn't justify keeping dead code. However, I'm not the one picking the cherries, so 😉️

killerwife:
I frequently keep some in core to avoid merge errors, but not these. These are likely old deprecated things from porting down from tbc.

PR #568, opened 2024-07-10
Core/PacketIO: Update RaidGroupError enum

Exxenoz:
## 🍰 Pullrequest
Updates raid group error definitions for SMSG_RAID_GROUP_ONLY (1.12. and 2.4.3 only). Verified using IDA.


killerwife:
Maybe also figure out if they are meant to be sent with other values than ERR_RAID_GROUP_NONE

PR #571, opened 2024-07-14
OutdoorPVP SI/EP

Hectelion:
entering silithus/eastern plaguelands when outdoorpvp is not enabled will show no longer the AlwaysUpFrame* on top of the screen

## 🍰 Pullrequest
Avoid the AlwaysUpFrame* showing up when the OutdoorPVP is set to not enable in mangosd.conf

### Issues
doing the teleport to SI/EP without this update may often cause the AlwaysUpFrames show up even when outdoorpvp is not enable

### How2Test
- without patch: set both the OutdoorPvp.*Enabled = 0, teleport to SI or EP and see the AlwaysUpFrames on top of the screen
- with patch: set both the OutdoorPvp.*Enabled = 0, teleport to SI or EP and see the AlwaysUpFrames are not showed

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -6785,7 +6785,11 @@ void Player::UpdateZone(uint32 newZone, uint32 newArea, bool force)
         sOutdoorPvPMgr.HandlePlayerEnterZone(this, newZone);
         sWorldState.HandlePlayerEnterZone(this, newZone);
 
-        SendInitWorldStates(newZone);                       // only if really enters to new zone, not just area change, works strange...
+//        SendInitWorldStates(newZone);                       // only if really enters to new zone, not just area change, works strange...
+	if ((newZone == 139 || newZone == 1377) && !sWorld.getConfig(CONFIG_BOOL_OUTDOORPVP_EP_ENABLED))
+	{}
+	else
+		SendInitWorldStates(newZone);                       // never send when entering EP/SI end outdoorpvp is disabled	    
 
         if (sWorld.getConfig(CONFIG_BOOL_WEATHER))
         {


insunaa:
Please provide proof that the worldstate panel should be invisible in the area when PVP is disabled. Also for the love of all the gods don't do things like
```cpp
if (someCondition)
  {}
else
  someStuff;
```

Hectelion:
Immagine 2024-07-14 193447
Immagine 2024-07-14 193520
Immagine 2024-07-14 193651
Immagine 2024-07-14 193721
added some screenshots; about the code any suggestion to do it better is welcome

insunaa:
I meant proof for how it's supposed to be on ABK servers. I believe you when you say that those are the changes you made, but the question is if those changes are legitimate in the first place.

As for the code, you'll have to rearrange the condition so that you can do
```cpp
if (someCondition)
  someStuff;
```

Exxenoz:
In your PR, no world states are sent at all for the two zones when opvp is disabled. It would be better to take a more targeted approach here. There is probably a world state or opcode for showing/hiding this specific display.

insunaa:
That assumes that the world state display is actually meant to be hidden. I suspect that since the World-PVP Event in those zones is never disabled on live servers and was never meant to be disabled either, that there's no dedicated functionality to actually hiding the worldstate display.

Exxenoz:
> That assumes that the world state display is actually meant to be hidden. I suspect that since the World-PVP Event in those zones is never disabled on live servers and was never meant to be disabled either, that there's no dedicated functionality to actually hiding the worldstate display. But I don't _know_ so I'd like to see proof that this change accurately reflects the reference environment.

Seems like display activation is bound to SMSG_INIT_WORLD_STATES and its zone id. The display is even activated when an empty (without any world state data) SMSG_INIT_WORLD_STATES is sent.

PR #573, opened 2024-07-24
Modify WorldSocket::HandleAuthSession to store realmID instead of LOGIN_TYPE_MANGOSD constant in loginSource column of account_logons table.

vcorbuleanu:
This PR modifies WorldSocket::HandleAuthSession to store the realmID in the loginSource column of the account_logons table, rather than the LOGIN_TYPE_MANGOSD constant.
This is required for those who use a single realmd instance to serve multiple mangosd instances.

As it stands a server "owner" can't tell the difference between an authentication to mangosd instance X and an authentication to mangosd instance Y. They're logged with the same number (1).

This change would maintain the value 0 for AuthSocket::verifyVersionAndFinalizeAuthentication, and different realmIDs (non 0 values) for each separate mangosd instance.


insunaa:
Please don't open PRs for changes that were already merged: https://github.com/cmangos/mangos-tbc/commit/7b6f9a4ed7bfc548ef52a8bde85a7d39183a954b

PR #574, opened 2024-07-31
disable map optimizations for playerbots if sPlayerbotAIConfig.DisableBotOptimizations

Redbu11dev:
## 🍰 Pullrequest
disable map optimizations for playerbots if sPlayerbotAIConfig.DisableBotOptimizations (default off - optimizations enabled) to allow full activity at all times

because no other way to disable this behavior from config, and it hinders bot activity

### How2Test
- build with Playerbots
- set AiPlayerbot.DisableBotOptimizations = 1 in aiplayerbot.conf
- launch the server

### DIFF
--- a/src/game/Maps/Map.cpp
+++ b/src/game/Maps/Map.cpp
@@ -847,7 +847,14 @@ void Map::Update(const uint32& t_diff)
             plr->Update(t_diff);
 
 #ifdef ENABLE_PLAYERBOTS
-            plr->UpdateAI(t_diff, !shouldUpdateBot);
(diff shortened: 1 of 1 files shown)


mostlikely4r:
I am the original author of most of these bot optimizations and I approve this PR.

These are parts of the map update that are less important for bots but it is perfectly fine to give users control to disable these. 

The bot system and server behavior are not impacted by enabling this config (everything ran fine for years without these optimizations) besides lowering some performance and increasing bot reactivity.

PR #576, opened 2024-08-03
Send Spell mods on player reconnect

Novivy:
## 🍰 Pullrequest
This fix a very old bug where the client would not apply spell modifiers like talents when a player is reconnecting while having its character still connected in-game.

### Proof
- None

### Issues
- None

### How2Test
- Login on a warlock character.
- Take the improved 5/5 corruption talent which makes corruption spell instant cast.
- .save
- Open another client, connect on same character while having the character still in-game on the first client
- Try to cast corruption while moving. 
- 
Without this fix, you won't be able to cast it while moving, despite having the 5/5 talent.


Novivy:
Another related issue to this particular situation: If you are on a fly path when reconnecting, you will be drop to the ground.

On my end I fixed it by kicking the player before resetting the session state (` if (!m_session->RequestNewSocket(this))`) if is he is on a flypath as I don't have much time to find the root cause right now.

killerwife:
Please, each PR needs to be its own branch.

PR #582, opened 2024-09-18
Stormwind: Fix Windsor not despawning after The Great Masquerade

insunaa:
## 🍰 Pullrequest

This PR ensures that Reginald Windsor despawns at the end of The Great Masquerade (and also doesn't have his horse model when fake-dead)

### Proof

- [None](https://youtu.be/myj4g07wb88?si=sJi1SaErCsVidIx0&t=217)

### Issues

- fixes https://github.com/cmangos/issues/issues/3482 fully

### How2Test

- 😱😱😱😱😱


killerwife:
@Tobschinski if you could look how he is meant to die, or at least throw a sniff with this q my way, I would appreciate it. Suicide in code generally means we are missing smth. Functionally no comments about this pr, just this one question.

PR #583, opened 2024-10-20
Auth: Enable 2FA on Classic

insunaa:
## 🍰 Pullrequest

This PR adds 2FA capabilities to 1.12 clients by abusing the PIN field

### DIFF
--- a/src/realmd/AuthSocket.cpp
+++ b/src/realmd/AuthSocket.cpp
@@ -117,6 +117,7 @@ struct sAuthLogonProof_C
     uint8   crc_hash[20];
     uint8   number_of_keys;
     uint8   securityFlags;                                  // 0x00-0x04
+    PINData pinData;
 };
 /*
 typedef struct
@@ -492,13 +493,20 @@ bool AuthSocket::_HandleLogonChallenge()
                             if (!self->_token.empty() && self->_build >= 8606) // authenticator was added in 2.4.3
                                 securityFlags = SECURITY_FLAG_AUTHENTICATOR;
 
+                            if (!self->_token.empty() && self->_build <= 6141)
+                                securityFlags = SECURITY_FLAG_PIN;
+
                             *pkt << uint8(securityFlags);                    // security flags (0x0...0x04)
 
                             if (securityFlags & SECURITY_FLAG_PIN)          // PIN input
                             {
-                                *pkt << uint32(0);
-                                *pkt << uint64(0);
-                                *pkt << uint64(0);
+                                uint32 gridSeedPkt = self->m_gridSeed = static_cast<uint32>(0);
+                                EndianConvert(gridSeedPkt);
+                                self->m_serverSecuritySalt.SetRand(16 * 8); // 16 bytes random
+                                self->m_promptPin = true;
+
+                                *pkt << gridSeedPkt;
+                                pkt->append(self->m_serverSecuritySalt.AsByteArray(16).data(), 16);
                             }
 
                             if (securityFlags & SECURITY_FLAG_UNK)          // Matrix input
@@ -539,7 +547,8 @@ bool AuthSocket::_HandleLogonProof()
     DEBUG_LOG("Entering _HandleLogonProof");
     ///- Read the packet
     std::shared_ptr<sAuthLogonProof_C> lp = std::make_shared<sAuthLogonProof_C>();
-    Read((char*)lp.get(), sizeof(sAuthLogonProof_C), [self = shared_from_this(), lp](const boost::system::error_code& error, std::size_t read)
+    auto size = m_promptPin ? sizeof(sAuthLogonProof_C) : sizeof(sAuthLogonProof_C) - sizeof(PINData);
+    Read((char*)lp.get(), size, [self = shared_from_this(), lp](const boost::system::error_code& error, std::size_t read)
     {
         if (error)
         {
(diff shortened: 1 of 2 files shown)


src/realmd/AuthSocket.cpp
```diff
@@ -575,10 +584,23 @@ bool AuthSocket::_HandleLogonProof()
...
+        {
+            auto pin = self->generateToken(self->_token.c_str());
+
+            if (pin != uint32(-1))
+                pinResult = self->VerifyPinData(pin, lp->pinData);
+        }
+
         ///- Check if SRP6 results match (password is correct), else send an error
-        if (!self->srp.Proof(lp->M1, 20))
+        if (!self->srp.Proof(lp->M1, 20) && pinResult)
```
killerwife, line 601: You flipped the flow of code. All code first checks password, then pin/authenticator. Unless there is a good reason for that, rewrite it so that proof is checked first and pin after

src/realmd/AuthSocket.cpp
```diff
@@ -117,6 +117,7 @@ struct sAuthLogonProof_C
     uint8   crc_hash[20];
     uint8   number_of_keys;
     uint8   securityFlags;                                  // 0x00-0x04
+    PINData pinData;
```
killerwife, line 120: Dont you think that if you are doing memory magic there should be a code comment at least mentioning it?

src/realmd/AuthSocket.cpp
```diff
@@ -539,7 +547,8 @@ bool AuthSocket::_HandleLogonProof()
     DEBUG_LOG("Entering _HandleLogonProof");
     ///- Read the packet
     std::shared_ptr<sAuthLogonProof_C> lp = std::make_shared<sAuthLogonProof_C>();
-    Read((char*)lp.get(), sizeof(sAuthLogonProof_C), [self = shared_from_this(), lp](const boost::system::error_code& error, std::size_t read)
+    auto size = m_promptPin ? sizeof(sAuthLogonProof_C) : sizeof(sAuthLogonProof_C) - sizeof(PINData);
```
killerwife, line 550: Cyberium had a good point that these sizeof magics should be a function of the struct.

insunaa: I honestly don't know how I would accomplish that

src/realmd/AuthSocket.cpp
insunaa: yeah, fair enough, I'll add a comment

src/realmd/AuthSocket.cpp
insunaa: should be possible, I'll change it

src/realmd/AuthSocket.cpp
```diff
@@ -1036,6 +1058,78 @@ bool AuthSocket::_HandleXferAccept()
     return true;
 }
 
+// Verify PIN entry data
+bool AuthSocket::VerifyPinData(uint32 pin, const PINData& clientData)
```
killerwife, line 1062: Not saying this code is useless by any means. This is basically a FIXED PIN function. Essentially Password2. All other cores use TOTP algorithm for verification.

https://github.com/vmangos/core/blob/development/src/realmd/AuthSocket.cpp#L663

look here, they have the option to use either. Our default is totp, so you should use that as a default for the pin check. IF you wish to add a db flag to an account to also enable fixed pin, im perfectly happy with that (note column flags is used by mangosd, so its not safe for realmd to also use it), but default should be totp when a token is used with an account

insunaa: Classic/Vanilla client does not have TOTP support, only PIN. So with this I'm faking TOTP support by entering it in the PIN pad

src/realmd/AuthSocket.cpp
killerwife: size_t size = lp->getsize(bool withPin)

src/realmd/AuthSocket.cpp
killerwife: Client doesnt do TOTP. Its about the code you input. All other cores input a google authenticator code (also known as totp) from https://datatracker.ietf.org/doc/html/rfc6238 instead of a fixed password pin. Vmangos does exactly that if you read the link i posted with the vanilla pin.

insunaa: Yeah, this is what I'm also doing. I just didn't add an option for fixed PIN because that's pointless

killerwife: Guess i missed that

src/realmd/AuthSocket.cpp
```diff
@@ -613,7 +626,26 @@ bool AuthSocket::_HandleLogonProof()
                 });
                 return;
             }
+            bool pinResult = true;
+
+            if (self->m_promptPin && (lp->securityFlags & SECURITY_FLAG_PIN))
```
killerwife, line 631: Why even do this pin variable shenanigans. Just put the whole thing in an IF statement and check if token is empty there. Structure it like the if (lp->securityFlags & SECURITY_FLAG_AUTHENTICATOR || !self->_token.empty()) statement. In fact m_promptPin is redundant and only SECURITY_FLAG_PIN and token needs to be checked.

src/realmd/AuthSocket.cpp
```diff
@@ -613,7 +626,26 @@ bool AuthSocket::_HandleLogonProof()
...
+                pinResult = false;
+
+            if (self->m_promptPin && (lp->securityFlags & SECURITY_FLAG_PIN) && !self->_token.empty())
+            {
+                auto pin = self->generateToken(self->_token.c_str());
+
+                if (pin != uint32(-1))
```
killerwife, line 638: Generate token has no case where it returns -1 to my knowledge

src/realmd/AuthSocket.cpp
insunaa: there are multiple cases:
pin requested by the server and correct pin provided by the client
pin requested by the server and incorrect pin provided by the client
pin not requested by the server and pin provided by the client
so I use m_promptPin to check if server requested pin and SECURITY_FLAG_PIN to check if the client provided a PIN

killerwife: Well but we only care about if token is empty or pin was sent. I get that the other cases exist, but we do not need to handle those.

killerwife:
Ok, anything on this pr that isnt done except me testing it? As its marked as a draft.

insunaa:
@killerwife sorry no, only needs review, afaik. Viccroy and I tested the changes and for us they work, but I'm always wary of getting the correct result by using the wrong method, so I'd appreciate if you checked the code for any more stupidities I may have commited

PR #586, opened 2024-11-04
make razorgore vanilla blizzlike

Vigerus:
## 🍰 Pullrequest
Make Razorgore Vanilla Blizzlike

### Issues
https://github.com/cmangos/issues/issues/3714


insunaa:
Did you test this?

Vigerus:
> Did you test this?

yup

Vigerus:
The debuff is applied at the mc moment now. It's a bit silly this way (the debuff is shorter than the mc time) and i think SoM had a "real" mechanics idea, but we should keep it vanilla way i guess.

PR #587, opened 2024-11-09
Add Evade persistance for s.4161,7486,8273,8393,8434,9463,9941,11984,12002,16592,21190

AnonXS:


killerwife:
These also apply to tbc right?

AnonXS:
yes

PR #588, opened 2024-11-11
Makes BRD arena spectator stairs walkable

Novivy:
## 🍰 Pullrequest
Stairs on top of BRD arena don't have any path out of it and prevent mobs from reaching player in melee.

### Proof
image

This fix the pathing on the tile

### DIFF
--- a/contrib/extractor_scripts/config.json
+++ b/contrib/extractor_scripts/config.json
@@ -9,8 +9,8 @@
 	},
 	"230": {
 		"3230": {
-			"walkableHeight": 6.9,
-			"walkableClimb": 6.4,
+			"walkableHeight": 7,
+			"walkableClimb": 6,
 			"_Info": "Makes BRD arena spectator stairs walkable"
 		}
 	},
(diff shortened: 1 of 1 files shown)


celguar:
and how it looks after

contrib/extractor_scripts/config.json
```diff
@@ -7,6 +7,13 @@
 			"_Info": "walkableClimb <= 8 else .go c 1000403 goes over the edge"
 		}
 	},
+	"230": {
+		"3230": {
+			"walkableHeight": 6.9,
```
killerwife, line 12: They are integers. https://recastnav.com/structrcConfig.html check here

Novivy:
Looks like this after:

image

killerwife:
So this fix is meant to be that an npc can walk up the stairs while chasing. Could i get a video from official where i can clearly see that is the case?

killerwife:
I will again ask for a video of how it looks on official just so ppl dont think i like to watch prs languish.

al3xc1985:
why close?

cpevors:
The head repository from which this PR stemmed was deleted.

PR #589, opened 2024-11-13
Playerbot: Prevent double login while player is teleporting

mostlikely4r:
## 🍰 Pullrequest
This is a minor fix that stops players (and bots) from actually logging into a character that is already online. This would result in player duplication where two players with the same guid exist in the server.

### Proof
During a teleport players are marked as inWorld=false. Before this fix only players where inWorld=true were being detected as already logged in.

### How2Test
With latest botcode this can no longer be reproduced since the check to see if a login request should be sent has also been adjusted to include teleporting players.

### DIFF
--- a/src/game/Entities/CharacterHandler.cpp
+++ b/src/game/Entities/CharacterHandler.cpp
@@ -124,7 +124,7 @@ void PlayerbotHolder::HandlePlayerBotLoginCallback(QueryResult* dummy, SqlQueryH
     botSession->SetNoAnticheat();
 
     // has bot already been added?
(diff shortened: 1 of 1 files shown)


mostlikely4r:
Note this is duplicate to https://github.com/cmangos/mangos-tbc/pull/707 which was already merged in tbc.

Exxenoz:
It was already backported here: https://github.com/cmangos/mangos-classic/commit/36b1b192b734fdf8775244044ad1b5cb3b049a45

mostlikely4r:
Confirmed. Already merged.

PR #591, opened 2024-12-07
[SD] Cthun: Always handle phase transition no matter the combat state

AnonXS:


ileboii:
Tested in 40man bot raid, can confirm this fixes it.

PR #596, opened 2025-02-12
Upgrade soon to be deprecated runners to newer versions

Niam5:
## 🍰 Pullrequest
Ubuntu 20.04 runners will soon be shut down by GH. Some of the Discord notifiers were still using this version. Also since there is only 1 remaining x64 Apple runner, upgraded the Apple CI test to be compatible with Apple Silicon.

### Proof
(https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/)

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -7,15 +7,13 @@ env:
   BUILD_TYPE: Release
   REPO_DIR : ${{github.workspace}}
   BUILD_DIR: ${{github.workspace}}/bin/builddir
-  BOOST_VERSION: "1.83.0"
-  BOOST_PLATFORM_VERSION: "13"
 
 permissions:
   contents: read
 
 jobs:
   build:
-    runs-on: macos-13
+    runs-on: macos-14
     permissions:
       contents: read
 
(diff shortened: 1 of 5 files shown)


.github/workflows/macos.yml
```diff
@@ -7,15 +7,13 @@ env:
...
-  BOOST_VERSION: "1.83.0"
-  BOOST_PLATFORM_VERSION: "12"
 
 permissions:
   contents: read
 
 jobs:
   build:
-    runs-on: macos-12
+    runs-on: macos-14
```
evil-at-wow, line 16: I updated the macOS runners a while ago after pushing changes and noticing the macOS builds were failing (see cmangos/mangos-tbc@9de293e350f06e3ce6b3144619a78de55f8cbd53 and the merges this week to the other cores), so macOS builds should be fine again for now.

There's certainly something to say for going to macos-14 and arm64 builds, but also for using a somewhat older system (with typically older toolchain). The first tests that "newest" stuff can build the core, the second that it also still works with "older" stuff. I think we tend to go with older because newer is tested more by devs/users.

I'm going to guess the boost and MySQL changes are related to the move to arm64 macOS builds?

src/realmd/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if(UNIX)
 endif()
 
 if(APPLE)
-  set(EXECUTABLE_LINK_FLAGS "-framework Carbon ${EXECUTABLE_LINK_FLAGS}")
+  set(EXECUTABLE_LINK_FLAGS "-Wl,-no_fixup_chains -framework Carbon ${EXECUTABLE_LINK_FLAGS}")
```
evil-at-wow, line 80: I actually fixed that problem a while ago (see cmangos/mangos-tbc@111e48d17713e4f0dd9b1ac02d5e4f43b3fdd993 and the merge of that to other cores this week), so we don't need to sweep that issue under the carpet any more :wink:

evil-at-wow (commented):
Using ubuntu-latest for the Discord stuff seems fine. Especially because this PR shows this part gets forgotten anyway, unlike the actual builds.

.github/workflows/macos.yml
killerwife: Answer to this would be preferable before merging. Otherwise no comments from me

Niam5: I commented below but it seems GH didn't send it out. The reason I set it to 14 is 13 is the last Intel runner as detailed here https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources so the choice is to either change now and fixup the runner for the direction GH is headed or just kick the can down the road and when 13 goes away do it then. Figured it was easier to fix it now and follow the direction the runners are going then just postponing it

Phatcat: What? No it's not, just use macos-14-large or macos-15-large, those are intel..?

Seems like the info in your link is outdated. This is from the images repo directly:
https://github.com/actions/runner-images
Screenshot 2025-05-17 at 23 21 09

https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md <- that's not arm.

Niam5:
Because large runners are not free. Only standard runners are included for
no costs. And there are no standard Mac runners on MacOS 14+. See here
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Phatcat:
wow really? locking mac intel, which is already legacy, behind a paywall is insane. I never tried the large ones, weird how it's not better documented at the actual repo, not a single mention of paid vs free runners. tsk.
Oh well, doesn't matter cause I concur with you, just move the runner onto apple silicon, it's not like apple is going back to intel anyway.

killerwife:
If any part of this is still valid, pls rebase.

Niam5:
Rebased and I added a few more changes. The new 7-zip that got added into the Windows runner causes the Boost installer to fail should the version be upgraded. Changed Boot install to use Chocolaty.

PR #599, opened 2025-04-01
G3D: Add support for RISC-V CPU's

Niam5:
## 🍰 Pullrequest
This change allows for compilation under RISC-V instruction set CPU's. Tested functionality of server, could enter game, playerbots compiles, all extractors work.

### Proof
- None

### Issues
- When compiling on a RISC-V CPU would error out in build due to undefined processor definition causing to to fall back to x86 code.

### How2Test
- Attempt to build on RISC-V
- Error
- Apply change, error gone and compile successful

### Todo / Checklist
- [X] None

### DIFF
--- a/dep/g3dlite/System.cpp
+++ b/dep/g3dlite/System.cpp
@@ -1710,6 +1710,15 @@ void System::cpuid(CPUIDFunction func, uint32& eax, uint32& ebx, uint32& ecx, ui
     edx = 0;
 }
 
+#elif defined(G3D_LINUX) && ( defined(__riscv) || defined(__riscv32) || defined(__riscv__) || defined(_riscv))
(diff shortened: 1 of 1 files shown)


insunaa:
Newer versions of G3D don't have the `cpuid` function anymore, so I don't see a need to upstream the changes, especially because there seem to be no guidelines for contributions.

Niam5:
I have to work with what is in our repo unfortunately. I looked at
upgrading G3D but it seems they added a depedance on TBB and I don't really
want to get into those weeds.

insunaa:
@Niam5 Yeah sorry, before the edit my reply was phrased poorly. I think your contribution is good and valid, just saying there would probably be no point in trying to upstream the change to the G3D project.

PR #602, opened 2025-06-21
AHBot: Add support for dynamic or static filtering by RequiredLevel and ItemLevel

mr-jjg:
### Summary

This feature adds support for filtering items listed by AuctionHouseBot using either a **static** or **dynamic** RequiredLevel cap.

- **Static filtering** uses a fixed value from the config
- **Dynamic filtering** adjusts based on the highest-level online character(s), with optional exclusion of GM accounts

This controls both the maximum RequiredLevel and ItemLevel allowed in auction listings, helping tailor AHBot behavior to the server’s progression.

---

### Behavior

- If `AuctionHouseBot.Level.DynamicMaxRequired = 1`, RequiredLevel is capped using the result of:
  `SELECT MAX(level) FROM characters WHERE online = 1`
- If disabled (`= 0`), the static config value `AuctionHouseBot.Level.MaxRequired` is used instead
- If `AuctionHouseBot.Level.IgnoreGmAccounts = 1`, GM accounts are excluded from dynamic level calculation
- `AuctionHouseBot.Level.DynamicRefreshInterval` sets how often AHBot re-evaluates online character levels (in minutes)

ItemLevel is capped to match RequiredLevel in either case, **except** when RequiredLevel reaches **60 or higher** — at which point the ItemLevel cap is lifted entirely (`255`), allowing full item availability.

---

### Edge Case

Note: The presence of a level 60 character (e.g., an admin or test account) will lift the ItemLevel cap in dynamic mode, potentially exposing high-end gear earlier than intended on fresh realms.

---

### Configuration

**New/Updated Keys in `ahbot.conf`:**
```
AuctionHouseBot.Level.DynamicMaxRequired = 1       ; Enable dynamic level filtering
AuctionHouseBot.Level.MaxRequired = 10             ; Fallback static RequiredLevel cap
AuctionHouseBot.Level.IgnoreGmAccounts = 1         ; Exclude GM accounts from level check
AuctionHouseBot.Level.DynamicRefreshInterval = 20  ; Time (in minutes) between level recalculations
```

### DIFF
--- a/src/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -563,7 +563,7 @@ void AuctionHouseBot::CalculateItemLevelCap()
     }
     else
     {
-        m_maxItemLevel = m_maxRequiredLevel;
+        m_maxItemLevel = m_maxRequiredLevel + 5; // Typical item level is required level + 5
     }
 }
 
--- a/src/game/AuctionHouseBot/AuctionHouseBot.cpp
+++ b/src/game/AuctionHouseBot/AuctionHouseBot.cpp
@@ -17,6 +17,7 @@
  */
 
 #include "AuctionHouseBot.h"
+#include "Common.h"
 #include "Globals/ObjectMgr.h"
 #include "Log/Log.h"
 #include "Policies/Singleton.h"
@@ -164,6 +165,17 @@ void AuctionHouseBot::Update()
     AuctionHouseObject* auctionHouse = sAuctionMgr.GetAuctionsMap(houseType);
     if (m_houseAction < MAX_AUCTION_HOUSE_TYPE && urand(0, 99) < m_chanceSell)
     {
+	// Lazy-refresh dynamic level
+	if (m_useDynamicMaxLevel)
+	{
+	    uint32 now = time(nullptr);
+	    if (now - m_lastLevelUpdateTime >= m_levelRefreshInterval)
+	    {
+		UpdateDynamicMaxLevel();
+		CalculateItemLevelCap();
+		m_lastLevelUpdateTime = now;
+	    }
+	}
         // Sell items
         std::unordered_map<uint32, uint32> itemMap;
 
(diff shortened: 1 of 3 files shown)


src/game/AuctionHouseBot/AuctionHouseBot.cpp
```diff
@@ -447,6 +454,44 @@ void AuctionHouseBot::FillUintVectorFromQuery(char const* query, std::vector<uin
...
+        if (auto result = CharacterDatabase.PQuery("SELECT MAX(level) FROM characters"))
+        {
+            Field* field = result->Fetch();
+            m_maxRequiredLevel = field[0].GetUInt32();
+            sLog.outString("AHBot: Dynamic max required level set to %u", m_maxRequiredLevel);
+        }
+        else
+        {
+            sLog.outError("AHBot: Failed to retrieve max character level. Falling back to default value 60.");
+            m_maxRequiredLevel = 60;
```
mostlikely4r, line 475: Could these hard values be replaced with the config defined max levels or at least the static MAX_DEFAULT_LEVEL

mostlikely4r:
This looks like a good addition to the ah bot functionality.
People have been asking for a way to get more control or add some progression.
This is a good first step.

killerwife:
Code wise, this is fine. Since I dont do anything else beyond that with ahbot, its fine from my end, but i would welcome if ppl using ah bot actually voiced their opinions.

mr-jjg:
I’ve made the following changes:

1. Replaced magic numbers 60 and 255 with DEFAULT_MAX_LEVEL and STRONG_MAX_LEVEL from DBCEnums.h
3. Standardized AHBot log messages to match existing formatting conventions (e.g. colons only in error messages)

Let me know if there’s anything else I can refine or clarify!

kelbren:
Perhaps you can add a config option to ignore any gm level account characters when you determine the max level.
or set the account level to include in the MAX(x) query. that would allow some control over the edge case situation.
Nice work.

mr-jjg:
> Perhaps you can add a config option to ignore any gm level account characters when you determine the max level. or set the account level to include in the MAX(x) query. that would allow some control over the edge case situation. Nice work.

That is a useful edge case that I hadn't considered. It shouldn't be a problem to implement, and it's given me a chance to get a better understanding of how CMaNGOS interfaces with the databases. Which do you think would be the most useful implementation?

- Binary toggle (ignore all GM accounts): `AuctionHouseBot.Level.IgnoreGmAccounts = true`
- Threshold value (e.g., `maxGmLevelToInclude = 1` to include GM1 but exclude GM2+)
- Exact match (e.g., `GmLevelAllowed = 0` for only regular players)

Any thoughts on which you'd prefer — or is there another approach you’d suggest?

kelbren:
I would think the settings would be good to exclude any level at or above the value,  so a value of 2 would exclude 2 and 3 etc.  if set to 0 then only applies to regular players.

mr-jjg:
**Testing Notes**
I spent time stress-testing this across multiple scenarios to ensure everything worked as expected. Here's a breakdown of what I tested and confirmed:

**Setup**
- `AuctionHouseBot.Level.DynamicMaxRequired = 0` initially to confirm static worked as expected, then 1 for remainder of tests
- `AuctionHouseBot.Level.MaxRequired = 10` (for fallback testing)
- `AuctionHouseBot.Level.IgnoreGmAccounts = 1` and toggled on and off during tests
- `AuctionHouseBot.Level.DynamicRefreshInterval` tested at 1 and 40 to observe interval behavior
- Two regular test player and one GM account at varying levels (35–60), parked at Alliance, Horde and Neutral Auction Houses

**Observations**

- AHBot only used online characters when calculating the max required level — confirmed via console logs and auction contents
- When `IgnoreGmAccounts = 1`, GM characters were correctly excluded from influencing the dynamic level cap
- Refresh interval was honored:
- With a short interval, item caps updated quickly after level changes (visible in logs)
- With a long interval, no premature updates occurred (console logs remained quiet as expected)
- Verified correct fallback when no characters were online
- Item level caps matched required levels correctly, and lifted to 255 once RequiredLevel hit 60+
- Monitored both in-game AH and direct MySQL queries to confirm item levels
- Item filtering respected level rules across all AHs (Alliance, Horde, Neutral)

**Conclusion**
Everything appears to function as intended — dynamic logic, fallback, refresh timing, GM exclusion, and item cap syncing all passed cleanly. I'm not planning any new features at this time, but I’d be happy to fix any issues that arise. If you’re able to help test, I’d really appreciate the feedback!

mostlikely4r (approved):
Not much to add. It looks pretty tidy and seems to work as suggested.

PR #608, opened 2025-09-21
CI/CD: Fix MacOS Build

insunaa:
## 🍰 Pullrequest

This PR should fix the CI/CD build for MacOS for now

Runs successfully in my own fork at least.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -27,13 +27,15 @@ jobs:
         run: |
           brew install mysql-client@8.4
           brew install openssl
-          brew install boost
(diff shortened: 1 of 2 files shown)


Niam5:
It wasn't Boost that was the issue, its cotire. That whole subroutine can be removed as CMake 3.16 has assumed all the functions it did and teh project has been archived and is unsupported. Should look at removing it entirely.

Niam5:
Slight correction, it is partially Boost as past Boost 1.88 the header file they were using has been removed per https://github.com/boostorg/system/issues/132

PR #642, opened 2026-07-07
Add clarifying comment on WoW Classic Era deviation of sending SMSG_LOOT_MONEY_NOTIFY

bdebaere:
## 🍰 Pullrequest
Add clarifying comment to that not sending SMSG_LOOT_MONEY_NOTIFY is intentional

### Proof

image
On retail WoW Classic Era you see a message when looting money, even when playing solo. When I try this with my local cmangos I don't see such a message. It would be nice if the cmangos code could clarify that this is a known deviation so future people don't need to wonder if this is intentional.

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Loot/LootMgr.cpp
+++ b/src/game/Loot/LootMgr.cpp
@@ -2272,6 +2272,10 @@ void Loot::SendGold(Player* player)
     {
         player->ModifyMoney(m_gold);
 
+        WorldPacket data(SMSG_LOOT_MONEY_NOTIFY, 4);
+        data << uint32(m_gold);
+        player->GetSession()->SendPacket(data);
+
         if (m_guidTarget.IsItem())
         {
             if (Item* item = player->GetItemByGuid(m_guidTarget))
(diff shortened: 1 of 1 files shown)


cpevors:
This certainly conforms to Classic behaviour, but it's not Blizzlike to 1.12.

bdebaere:
@cpevors Added https://youtu.be/V0IlPFAf0rU?si=4rhj8mKTb1IUrHYr&t=118. That looks to be original proof?

bdebaere:
Then again that was 2010 already...

bdebaere:
@cpevors I think you are correct, I found https://youtu.be/mJaQJw1Wf2s?si=eLlHi9g4ATqqAnj0&t=87 which shows no message when looting money. It doesn't definitely prove that there wasn't a message on the exact patch 1.12.1, since it's from patch 1.9.2, but it's closer to 1.12.1 than the other video. Also found a video patch 1.10 with the same thing. Might be handy to include a comment though stating this is a known deviation from WoW Classic Era.

PR #650, opened 2026-08-22
Fix .gm fly in Classic

T-imothy:
## 🍰 Pullrequest

Restores `.gm fly` for Classic and aligns the movement handling with vMaNGOS.

Classic uses the legacy Vanilla movement flags because the 1.12 client does not have the native flying controls used by TBC and Wrath. The change also keeps GM flight authorized through movement validation and rejects stale movement packets when the mode changes.

Normal jump handling is enabled again. On the Classic client, Space is still a regular jump rather than a flying ascent control, so jumping removes fly mode as it does in vMaNGOS. Altitude is controlled by pitching up or down while moving forward.

### Proof

- Tested in-game on Classic with fly on/off, directional movement, stopping, and jump behavior.
- Built with AHBot and Playerbot enabled.

### Issues

- The Classic client cannot use Space to ascend continuously like TBC or Wrath.

### How2Test

- Enable with `.gm fly on`.
- Check forward, backward, strafe, stopping, and changing altitude by pitching while moving.
- Press Space and confirm the normal Classic jump behavior removes fly mode.
- Enable fly again, then disable it with `.gm fly off`.

### Todo / Checklist

- [x] None

### DIFF
--- a/src/game/Entities/Object.h
+++ b/src/game/Entities/Object.h
@@ -712,13 +712,13 @@ enum MovementFlags
     MOVEFLAG_WALK_MODE          = 0x00000100,               // Walking
 
     MOVEFLAG_LEVITATING         = 0x00000400,
-    MOVEFLAG_FLYING             = 0x00000800,               // Fixed Z in the Vanilla client
+    MOVEFLAG_FLYING             = 0x00000800,               // [-ZERO] is it really need and correct value
     MOVEFLAG_JUMPING            = 0x00002000,
     MOVEFLAG_FALLINGFAR         = 0x00004000,
     MOVEFLAG_SWIMMING           = 0x00200000,               // appears with fly flag also
     MOVEFLAG_SPLINE_ENABLED     = 0x00400000,               // [-ZERO] is it really need and correct value
-    MOVEFLAG_CAN_FLY            = 0x00800000,               // Moved in the Vanilla client
-    MOVEFLAG_FLYING_OLD         = 0x01000000,               // Flying in the Vanilla client
+    MOVEFLAG_CAN_FLY            = 0x00800000,               // [-ZERO] is it really need and correct value
+    MOVEFLAG_FLYING_OLD         = 0x01000000,               // [-ZERO] is it really need and correct value
 
     MOVEFLAG_ONTRANSPORT        = 0x02000000,               // Used for flying on some creatures
     MOVEFLAG_SPLINE_ELEVATION   = 0x04000000,               // used for flight paths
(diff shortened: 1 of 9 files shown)


AnonXS:
it works, but  dont use space while doing so.

grafik

`.mod speed` or one of the other modifiying speed commands needs to be able to adjust flying speed movement.

T-imothy:
I was using .m as 5-10 and this works

AnonXS:
yes `.mod aspeed` command works thx! xD

T-imothy:
Update: normal jump handling has been re-enabled instead of blocking the Space key in the movement handler. Classic 1.12 does not have the TBC/Wrath flying ascend control, so Space is still treated as a regular jump and removes fly mode. This now matches vMaNGOS behavior; changing altitude is done by pitching up or down while moving forward.

BadgerByNature:
> Proof
Tested in-game on Classic with fly on/off, directional movement, stopping, and jump behavior.

The PR says this was tested with Space, but then AnonXS said it was broken when pressing Space. What's the story there?

The wording of these comments, especially in this last commit, feels very Claude code.

src/game/Entities/Object.h
```diff
@@ -712,13 +712,13 @@ enum MovementFlags
     MOVEFLAG_WALK_MODE          = 0x00000100,               // Walking
 
     MOVEFLAG_LEVITATING         = 0x00000400,
-    MOVEFLAG_FLYING             = 0x00000800,               // [-ZERO] is it really need and correct value
+    MOVEFLAG_FLYING             = 0x00000800,               // Fixed Z in the Vanilla client
```
BadgerByNature: I was trying to figure out what these comments mean. It's actually referring to what they are called _in VMaNGOS_. CMaNGOS `MOVEFLAG_FLYING` is actually `MOVEFLAG_FIXED_Z` in VMaNGOS. The comment description is wrong - but even so, knowing what they're called in VMaNGOS doesn't seem particularly relevant, does it?

https://github.com/vmangos/core/blob/6958f2666be3feb7b5584f3a49cb36e5acf74172/src/game/Objects/MovementInfo.h#L42

T-imothy: Those enum comment changes were unnecessary and confusing. I added them to document the equivalent vMaNGOS flag meanings, but that context belongs in the fly implementation rather than the shared enum. They have been reverted now.

Jump was blocked in the first version while I was getting `.gm fly` working in Classic, because pressing Space caused broken movement. After Anon tested and reproduced that behavior, I compared it with vMaNGOS and changed this to follow the same Classic behavior instead: jump is no longer blocked, but the 1.12 client treats Space as a normal jump and cancels fly mode. The earlier attempt to imitate the CMaNGOS TBC/WotLK flight controls did not behave correctly in Classic. Should be good now.

